PR #1, opened 2015-03-14
[BUGFIX] ManagedSynonyms language determination

hhoechtl:
The ManagedSynonymsFilterFactory was moved from indexAnalyzer to
queryAnalyzer, therefore we must look there in the schema in order
to determine the language.

Resolves: #65676
Change-Id: I63e99c240f0d019dc411d7fe56204b273e1d715c

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
 		}
 	],
 	"require": {
-		"php": ">5.3.0,~5.5.0"
+		"php": ">5.3.0,~5.6.0"
 	},
 	"require-dev": {
 		"typo3-ci/typo3cms": "dev-master"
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -33,7 +33,7 @@
 		}
 	],
 	"require": {
-		"php": ">5.3.0,~5.5.0"
+		"php": ">5.3.0,~5.6.0"
```
irnnr: I don't see a reason for this change, you're not using any 5.6 specific features, are you? The other thing is to try to keep this in line with the TYPO3 LTS requirements.

irnnr: Thought about it: This is fine with me, we actually should raise the minimum PHP version to at least 5.4, maybe even 5.5. 

We shouldn't support EOL PHP versions, even if CMS might say otherwise.
http://php.net/supported-versions.php

dkd-dobberkau: Ubuntu 12.04 LTS (Precise Pangolin) has PHP 5.3.10 and has support untill 2017.
Ubuntu 14.04 LTS (Trusty Tahr) has PHP 5.5.12 and has support until 2019.
See: https://launchpad.net/ubuntu/+source/php5

I would suggest to stick our Ubuntu reference installation with 5.3 to 5.5 for the moment.

irnnr:
LGTM

PR #3, opened 2015-03-26
[FEATURE] Add flush errors button to backend module

markusguenther:
It is not possible to reset the error currently.

This pull request add an button to the backend module to flush the error field.

### DIFF
--- a/Classes/Backend/SolrModule/IndexQueueModuleController.php
+++ b/Classes/Backend/SolrModule/IndexQueueModuleController.php
@@ -67,7 +67,7 @@ class IndexQueueModuleController extends AbstractModuleController {
 	 * @return void
 	 */
 	public function initializeIndexQueueAction() {
-		$initializedIndexingConfigurations = [];
+		$initializedIndexingConfigurations = array();
 
 		$itemIndexQueue                     = GeneralUtility::makeInstance('Tx_Solr_IndexQueue_Queue');
 		$indexingConfigurationsToInitialize = GeneralUtility::_POST('tx_solr-index-queue-initialization');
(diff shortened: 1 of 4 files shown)


Classes/Backend/SolrModule/IndexQueueModuleController.php
```diff
@@ -67,7 +67,7 @@ public function indexAction() {
 	 * @return void
 	 */
 	public function initializeIndexQueueAction() {
-		$initializedIndexingConfigurations = array();
+		$initializedIndexingConfigurations = [];
```
irnnr: Please revert these unrelated changes. Also, array short syntax is only available with PHP 5.4+, but CMS 6.2 only requires a minimum version of 5.3.7.

Classes/Backend/SolrModule/IndexQueueModuleController.php
```diff
@@ -114,6 +114,32 @@ public function initializeIndexQueueAction() {
...
 
 	/**
+	 * Removes all errors in the index queue list. So that the items can be indexed again.
+	 *
+	 * @return void
+	 */
+	public function flushLogErrorsAction() {
+		/** @var DatabaseConnection $database */
+		$database = $GLOBALS['TYPO3_DB'];
+		$flushResult = $database->exec_UPDATEquery('tx_solr_indexqueue_item', 'errors NOT LIKE ""', ['errors' => '']);
```
irnnr: same her, please don't use array short syntax (although I agree it's a lot nicer)

Classes/Backend/SolrModule/IndexQueueModuleController.php
```diff
@@ -114,6 +114,32 @@ public function initializeIndexQueueAction() {
...
+
+		$message = 'All errors has been flushed';
```
irnnr: Let's go with this: "All errors *have* been reset."

irnnr:
Great change, just a few items that prevent merging it.

Classes/Backend/SolrModule/IndexQueueModuleController.php
markusguenther: To be honest i did not used locallang labels because the whole module is not using it ;). But it would be nice for sure.

markusguenther:
ok some adjustments and the localisation of the index queue module.
If this is ok for you i can localize the other controllers later in a different pull request.

PR #5, opened 2015-04-21
[BUGFIX] Do not show frequent search terms from other languages or sites in a multisite environment

jacobsenj:
While the check for the root page id or language is present for the sql query generation, it wasn't taken into account for the cache identifier generation. This fixes the problem of items from other language or sites being shown out of the cache.


irnnr:
Makes sense, good catch!

PR #6, opened 2015-05-06
[BUGFIX] Fixed menu generation for facets.

pogosheep:
* Added '/' as separator to menu entry prefix
  to prevent wrong assignments. Before 2-144/27/28 would be assigned
  as child of 1-144/2, not only 2-144/27.

### DIFF
--- a/Classes/Facet/HierarchicalFacetHelper.php
+++ b/Classes/Facet/HierarchicalFacetHelper.php
@@ -89,7 +89,7 @@ class Tx_Solr_Facet_HierarchicalFacetHelper {
 	protected function getSubMenu(array $facetOptions, $menuName, $level) {
 		$menu = array();
 
-		$subMenuEntryPrefix = $level . '-' . $menuName;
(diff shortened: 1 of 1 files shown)


irnnr:
Can you please explain more in detail what's the issue and how to reproduce it?

pogosheep:
You need a structure like the mentioned above to reproduce it. So create something like categories which are assigned to the records you want to index. The categories need to be a tree structure, so you have children.

In the case of above structure having "1-144/2" and "2-144/27", the element "2-144/27/28" will become a child of "1-144/2" which is of course wrong as it's only a child of "2-144/27". The added "/" will prevent this behaviour, as PHP will only look for the matching part.

irnnr:
Is this the same issue? https://forge.typo3.org/issues/66946

pogosheep:
Exactly yes, that's what this PR should fix.

irnnr:
thanks!

PR #7, opened 2015-05-21
CMS 7.2, bug in gc, endtime needs to be UTC

AndreasA:
- CMS 7.2 compatbility
- Bug in GarbageCollector where removed items due to start time in future were also removed from index instead of new changed time
- endtime has to be stored in UTC otherwise solr.processor.DocExpirationUpdateProcessorFactory e.g. in CEST timezone the items would be removed 2 hourse later. <br>Only done for UTC as for others it is not important as the times will always be in local time.
It might also create breaking changes if this is done for other values.


irnnr:
Can you please split this PR into 3 separate PRs? The commits are not related to each other and especially the two bug fixes would make sense to have in 3.0.1. That'd be super helpful! Otherwise this looks pretty good already!

AndreasA:
No problem. Will separate them.
Probably sometime tomorrow.
Should I create individual PRs for the bugfix for v3.0.1 and 3.1.x ?

Because I also renamed some of the old t3lib_div class names, e.g. 
\TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
which I would like to keep for 3.1.x

AndreasA:
Something else: Is there a reason why the composer.lock file is commited?
This extension cannot be used standalone and the provided composer.lock would be ignored if it is installed via composer. So adding the complete version of the one dependency and removing the composer.lock would make more sense.

AndreasA:
Split in:
https://github.com/TYPO3-Solr/ext-solr/pull/12
https://github.com/TYPO3-Solr/ext-solr/pull/11
https://github.com/TYPO3-Solr/ext-solr/pull/10

Also added corresponding PRs for v3.0.x:
https://github.com/TYPO3-Solr/ext-solr/pull/9
https://github.com/TYPO3-Solr/ext-solr/pull/8

PR #8, opened 2015-05-26
Bugfix: endtime in UTC for 3.0.x

AndreasA:
Split of Bugfix from https://github.com/TYPO3-Solr/ext-solr/pull/7 for v3.0.x

### DIFF
--- a/Classes/FieldProcessor/Service.php
+++ b/Classes/FieldProcessor/Service.php
@@ -66,20 +66,24 @@ class Tx_Solr_FieldProcessor_Service {
 				}
 
 				switch ($instruction) {
+					case 'timestampToUtcIsoDate':
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToUtcIsoDate');
+						$fieldValue = $processor->process($fieldValue);
+						break;
 					case 'timestampToIsoDate':
-						$processor  = t3lib_div::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
 						$fieldValue = $processor->process($fieldValue);
 						break;
 					case 'pathToHierarchy':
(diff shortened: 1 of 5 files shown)


AndreasA:
Did not test the bugfix for v3.0.x

irnnr:
Can you explain what the issue is / what you're trying to solve?

AndreasA:
OK.
Let's say the following the timezone of the server is Europe/Vienna which means currently GMT+2 (with DST).

Now let's say the endtime of a record is set to "2015-07-14 12:01" (server time).
That means I want the record to be removed from solr at "2015-07-14 12:01:00" (GMT+2).
Now, the endtime in solr will be set like this: "2015-07-14T12:01:00Z".

However, as solr actually expects times to be in UTC this is wrong an will result in the record to be removed from solr at "2015-07-14 14:01:00" (server time) instead of at "2015-07-14 12:01:00" (server time).
This is not an issue for "other" time values but for the endtime it is because endtime is set to be used via "DocExpirationUpdateProcessorFactory"

irnnr:
ok yeah, just saw it in #10, thanks!

Classes/FieldProcessor/Service.php
```diff
@@ -66,20 +66,24 @@ public function processDocument(Apache_Solr_Document $document, array $processin
 				}
 
 				switch ($instruction) {
+					case 'timestampToUtcIsoDate':
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToUtcIsoDate');
+						$fieldValue = $processor->process($fieldValue);
+						break;
 					case 'timestampToIsoDate':
-						$processor  = t3lib_div::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
```
irnnr: REALLY can't do this for 3.0.1 ;)

AndreasA: Damn forgot to revert that.
Thanks for noticing :)

AndreasA:
Fix wrong class used for makeInstance

irnnr:
LGTM, thanks a lot!

PR #9, opened 2015-05-26
Bugfix: garbage collector with starttime future (keep in queue) for v3.0.x

AndreasA:
Split of Bugfix from https://github.com/TYPO3-Solr/ext-solr/pull/7 for v3.0.x

### DIFF
--- a/Classes/GarbageCollector.php
+++ b/Classes/GarbageCollector.php
@@ -50,7 +50,7 @@ class Tx_Solr_GarbageCollector {
 			$this->collectGarbage($table, $uid);
 
 			if ($table == 'pages') {
-				$indexQueue = t3lib_div::makeInstance('Tx_Solr_IndexQueue_Queue');
+				$indexQueue = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_IndexQueue_Queue');
 				$indexQueue->deleteItem($table, $uid);
 			}
 		}
@@ -76,7 +76,7 @@ class Tx_Solr_GarbageCollector {
 			$this->collectGarbage($table, $uid);
 
 				// now re-index with new properties
-			$indexQueue = t3lib_div::makeInstance('Tx_Solr_IndexQueue_Queue');
+			$indexQueue = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_IndexQueue_Queue');
 			$indexQueue->updateItem($table, $uid);
 		}
 	}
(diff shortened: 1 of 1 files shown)


AndreasA:
Did not test the bugfix for v3.0.x but it should work

Classes/GarbageCollector.php
```diff
@@ -155,6 +155,20 @@ public function processDatamap_afterDatabaseOperations($status, $table, $uid, ar
 			|| ($table == 'pages' && !$this->isIndexablePageType($record))
 		) {
 			$this->collectGarbage($table, $uid);
+
+			// if the start time is in the future and not otherwise invisible,
+			// it is necessary to add the item to the queue again to be indexed once the start time is reached
+			if ($this->isStartTimeInFuture($table, $record)
+				&& !$this->isHidden($table, $record)
+				&& !$this->hasFrontendGroupsRemoved($table, $record)
+				&& ($table != 'tt_content' && $table != 'pages_language_overlay')
```
irnnr: why do you check for these tables? Also in its current form it does not need the parentheses around the two tables.

Classes/GarbageCollector.php
```diff
@@ -155,6 +155,20 @@ public function processDatamap_afterDatabaseOperations($status, $table, $uid, ar
 			|| ($table == 'pages' && !$this->isIndexablePageType($record))
 		) {
 			$this->collectGarbage($table, $uid);
+
+			// if the start time is in the future and not otherwise invisible,
+			// it is necessary to add the item to the queue again to be indexed once the start time is reached
+			if ($this->isStartTimeInFuture($table, $record)
```
irnnr: Would it make sense to pull this if out, and simply execute it after the if 151-158/172?

irnnr:
Do you have a description on how to reproduce this issue?

AndreasA:
Reproduce:
Add Page without starttime, have solr index the item.
Then add a starttime in the future: This will result in the item to be removed from solr and from the index.
Which means it will not be re-indexed once the new starttime is reached.

Classes/GarbageCollector.php
AndreasA: This will result in the item to be re-added to the index queue event if it was e.g. set to hidden.
Not sure though if that would be a problem?

Classes/GarbageCollector.php
AndreasA: Because if it is tt_content or pages_language_overlay there is an updateItem done instead of deleteItem anyway.

However, I think we could just execute the updateItem again?

Paranheses is not necessary but personally I wanted to add the "grouping" because for those two tables the updateItem is exeucted anyway instead of deleteItem.

Classes/GarbageCollector.php
AndreasA: Ah, now I understand you where the first isHidden check is done.

It might make sense to put it before?
Because e.g. if the record is also set to "hidden" we have to delete it from the queue.

However, that means it could be that the following happens:
1.) delete item
2.) remove from queue due to future starttime
3.) re-add item to queue due to future start time
4.) due to hidden remove it from queue again.

But I think that would be acceptable?
In which case it would only be the future und isindexed check and maybe the tables check.

irnnr:
The solution is actually a lot simpler :) https://github.com/TYPO3-Solr/ext-solr/commit/db85c0e1f99529f8b85fbd3676e91e73e429ba85

AndreasA:
Actually, I tried this first but it still does not work correctly that was because:
If I now index a record, then set the record hidden, it will result in the record to be deleted due to the garbage collector.

However, the RecordMonitor is re-adding the record to the index queue without indexed_time 0.

Though maybe the issue is actually that the RecordMonitor shouldn't add hidden elements to the index queue?

AndreasA:
Added PR for RecordMonitor check:
#22

irnnr:
Will have to check. I was a bit worried about side effects, too :)

AndreasA:
One thing maybe the comment should be modified to reflect that the order is important (after RecordMonitor hook), otherwise someone might switch them again later on.

irnnr:
+1, was thinking about that, too

PR #10, opened 2015-05-26
[BUGFIX] endtime has to be indexed in UTC (3.1.x)

AndreasA:
this is especially important for endtime because otherwise
DocExpirationUpdateProcessorFactory might be too early
or too late in removing a document

### DIFF
--- a/Classes/FieldProcessor/Service.php
+++ b/Classes/FieldProcessor/Service.php
@@ -66,20 +66,24 @@ class Tx_Solr_FieldProcessor_Service {
 				}
 
 				switch ($instruction) {
+					case 'timestampToUtcIsoDate':
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToUtcIsoDate');
+						$fieldValue = $processor->process($fieldValue);
+						break;
 					case 'timestampToIsoDate':
-						$processor  = t3lib_div::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
 						$fieldValue = $processor->process($fieldValue);
 						break;
 					case 'pathToHierarchy':
-						$processor  = t3lib_div::makeInstance('Tx_Solr_FieldProcessor_PathToHierarchy');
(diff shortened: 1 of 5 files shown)


Classes/FieldProcessor/Service.php
```diff
@@ -66,20 +66,24 @@ public function processDocument(Apache_Solr_Document $document, array $processin
 				}
 
 				switch ($instruction) {
+					case 'timestampToUtcIsoDate':
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToUtcIsoDate');
+						$fieldValue = $processor->process($fieldValue);
+						break;
 					case 'timestampToIsoDate':
-						$processor  = t3lib_div::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
+						$processor  = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('Tx_Solr_FieldProcessor_TimestampToIsoDate');
```
irnnr: Please avoid unrelated changes :)

Classes/Util.php
```diff
@@ -120,23 +122,30 @@ public static function timestampToIso($timestamp) {
...
+		return $dateTime ? (int)$dateTime->format('U') : 0;
+	}
+
+	/**
+	 * Converts a date from unix timestamp to ISO 8601 format in UTC timezone.
+	 *
+	 * @param	integer	$timestamp unix timestamp
+	 * @return	string	the date in ISO 8601 format
+	 */
+	public static function timestampToUtcIso($timestamp) {
```
irnnr: This method is missing in the 3.0.x PR

Classes/FieldProcessor/Service.php
AndreasA: OK, but I wanted to add the new one with the correct class name which is why I replaced the others but could have kept t3lib_div here

Classes/Util.php
AndreasA: https://github.com/TYPO3-Solr/ext-solr/pull/8/files#diff-99a1d1c23fd6b626eea8167b0dd97934R115

It is there?

irnnr: ok sorry, must have missed it. my bad

Classes/FieldProcessor/Service.php
irnnr: don't worry about these kind of changes, will take care of that

irnnr:
Thanks!

PR #11, opened 2015-05-26
[BUGFIX] gc items removed due to start time need to be kept in queue (3.1.x)

AndreasA:


irnnr:
Thanks again for your contribution! Just a hint: All your changes to replace t3lib_div are not related to the actual bugfix. Next time please keep them separate. This way the PR is easier to review, and in this case I could have applied it to master and then simply cherry pick it into 3.0.1.

irnnr:
Closing this PR for the above reasons (unrelated changes, otherwise the same as #9), continuing over there...

PR #12, opened 2015-05-26
[TASK] 7.2 compatibility (3.1.x)

AndreasA:


Classes/Backend/IndexingConfigurationSelectorField.php
```diff
@@ -102,12 +102,15 @@ public function getSelectedValues() {
...
-	 * Uses TCEforms.
+	 * Uses \TYPO3\CMS\Backend\Form\FormEngine.
 	 *
 	 *  @return string Markup for the select field
 	 */
 	public function render() {
-		$tceForm = t3lib_div::makeInstance('t3lib_TCEforms');
+		$selectFieldRenderer = $formEngine = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('TYPO3\\CMS\\Backend\\Form\\FormEngine');
+		if (class_exists('TYPO3\\CMS\Backend\\Form\\Element\\SelectElement')) {
+			$selectFieldRenderer = \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('TYPO3\\CMS\Backend\\Form\\Element\\SelectElement', $formEngine);
```
irnnr: When would this be the case? / When would SelectElement not be available?

Classes/PluginBase/PluginBase.php
```diff
@@ -160,11 +160,10 @@ protected function baseWrap($content) {
 	 * @param array $configuration configuration array as provided by the TYPO3 core
 	 */
 	protected function initialize($configuration) {
-		$this->conf = $configuration;
-
-		$this->conf = t3lib_div::array_merge_recursive_overrule(
-			$GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.'],
-			$this->conf
+		$this->conf = $GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.'];
+		\TYPO3\CMS\Core\Utility\ArrayUtility::mergeRecursiveWithOverrule(
```
irnnr: why the order change? Also for the next chunk.

EidSuggest/Suggest.php
```diff
@@ -25,43 +25,37 @@
...
-if (version_compare(TYPO3_version, '6.1-dev', '>=')) {
-	\TYPO3\CMS\Core\Core\Bootstrap::getInstance()->loadCachedTca();
-} else {
-	$TSFE->includeTCA();
-}
-$TSFE->sys_page = t3lib_div::makeInstance('t3lib_pageSelect');
-$TSFE->rootLine = $TSFE->sys_page->getRootLine($pageId, '');
-$TSFE->initTemplate();
-$TSFE->getConfigArray();
+\TYPO3\CMS\Frontend\Utility\EidUtility::initTCA();
```
irnnr: Is this compatible with 6.2?

EidSuggest/Suggest.php
```diff
@@ -25,43 +25,37 @@
 
 # TSFE initialization
 
-tslib_eidtools::connectDB();
-$pageId     = filter_var(t3lib_div::_GET('id'), FILTER_SANITIZE_NUMBER_INT);
+$pageId     = filter_var(\TYPO3\CMS\Core\Utility\GeneralUtility::_GET('id'), FILTER_SANITIZE_NUMBER_INT);
 $languageId = filter_var(
-	t3lib_div::_GET('L'),
+	\TYPO3\CMS\Core\Utility\GeneralUtility::_GET('L'),
```
irnnr: Please import the classes

Resources/Private/Layouts/Default.html
```diff
@@ -1,7 +1,7 @@
 {namespace solr=ApacheSolrForTypo3\Solr\ViewHelpers}
 
 <f:be.container
-	addCssFile="{f:uri.resource(path:'StyleSheets/ModuleAdministration.css')}"
+	includeCssFiles="{solrAdmin: '{f:uri.resource(path:\'StyleSheets/ModuleAdministration.css\')}'}"
```
irnnr: 6.2 compatibility?

irnnr:
Overall good PR!

Resources/Private/Layouts/Default.html
AndreasA: addCssFile works in 6.2 too but was deprecated.
Could also go back there...

EidSuggest/Suggest.php
AndreasA: Cannot use "use" here as it is and eID script not a class definition.
Or at least it was always avoided in core if it is only a file.

EidSuggest/Suggest.php
AndreasA: Yes initTca is 6.2 compatible:
https://git.typo3.org/Packages/TYPO3.CMS.git/blob/f695f377f48c23bec3156c6d0f58f8a078d8a651:/typo3/sysext/frontend/Classes/Utility/EidUtility.php#l78

The $TSFE is not exactly removed but was replaced with $GLOBALS['TSFE'] instead.

Classes/PluginBase/PluginBase.php
AndreasA: Because mergeRecursiveWithOverrule is now passed by reference.
Before it returned the result.
Which means if I want the same result as before, I have $this->conf be $GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.']

otherwise $GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.'] would get the result.

Classes/Backend/IndexingConfigurationSelectorField.php
AndreasA: SelectElement is new for 7.x
With this check it works also for 6.2 because there the SelectElement does not exist and it would use FormEngine to render as before.
I could also add a version check but I personally prefer feature checks. Maybe we could also check if formengine doesn't have the required render method?

AndreasA:
Added corresponding inline comments.
Let me know if I have to change anything. Especially, regarding SelectElement.

Added another change (setup.txt has the wrong class name for PageIndexer (lowercase instead of corresponding uppercase) which might make issue with composer class loading in 7.3

AndreasA:
Same for initialize class of page indexer

AndreasA:
Added change to handle the select field provider in 7.3 / 7.2 and 6.2
Maybe you can verify that it also works in older versions.

Not sure, however, if this is the best solution. Maybe it would be better to extend the SelectCheckboxElement instead and provide a mapper for the now protected method. Of course, that would only be possible if the corresponding class exists.

AndreasA:
another wrong class which results in errors in 7.3 (due to wrong case used in class name)

AndreasA:
Does this still need changes? If so, which?

irnnr:
I'll review after we have 3.0.1 out, thanks!

AndreasA:
No problem, take your time to review it.
Especially to ensure it works with also still works correctly with 6.2.
It should but I mostly tested it with later versions.

EidSuggest/Suggest.php
irnnr: That should not matter

Classes/PluginBase/PluginBase.php
irnnr: Is this new in 7.x?

Resources/Private/Layouts/Default.html
irnnr: ok, thx

irnnr:
LGTM, just needs a rebase

EidSuggest/Suggest.php
AndreasA: True you can import without class too, it is just not generally done in the core.
But I guess that's mostly because most of the times it is required in ext_* or TCA and those are merged during caching.
But in the dumpFile eID it isn't done with imports either but that could just be because of conversion from old classnames to new ones.

PR #14, opened 2015-06-07
Store Tomcat version in own value, for easy replacement

waja:
We needed to setup this with a Tomcat 7 and had to change all this values. It maybe handy to do this just in one place.

Originally opened at https://github.com/TYPO3-extensions/solr/pull/6


dkd-dobberkau:
What do you think Ingo?

irnnr:
LGTM

PR #22, opened 2015-06-24
[BUGFIX] RecordMonitor should not add items that aren't enabled

AndreasA:


irnnr:
Looks reasonable, will try it out later

irnnr:
👍🏻

PR #23, opened 2015-06-25
[TASK] moves indexing of additional news content elements to separate TypoScript template

peterkraume:
The SOLR_RELATION for the EXT:news field "content_elements" in Configuration/TypoScript/Examples/IndexQueueNews/setup.txt should be disabled by default since it depends on the activation of this feature in EXT:news configuration in EM (records.contentElementRelation) where it is disabled by default.
Related to [Forge issue #50146](https://forge.typo3.org/issues/50146).

### DIFF
--- a/Configuration/TypoScript/Examples/IndexQueueNews/setup.txt
+++ b/Configuration/TypoScript/Examples/IndexQueueNews/setup.txt
@@ -23,13 +23,6 @@ plugin.tx_solr.index.queue {
 						field = bodytext
 						noTrimWrap = || |
 					}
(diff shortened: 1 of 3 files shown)


dkd-dobberkau:
Thanks Peter. Doing so will not create any Question if People want to index news?

peterkraume:
With this patch a basic indexing of news records will work out of the box. Those who have enabled the additional content element feature need to include the second template.

dkd-dobberkau:
Sounds fine for me.

Issue #35, opened 2015-07-14
AuthorizationService Frontend helper causes database error


astehlik:
The ```Tx_Solr_IndexQueue_FrontendHelper_AuthorizationService``` returns a user array that does not contain a value for ```uid```.

This causes a problem when the ```AbstractUserAuthentication->createUserSession()``` method is called during the Frontend user initialization because the system tries to insert a NULL value in the ```ses_userid``` column:

```
Core: Session data could not be written to DB. Error: Column 'ses_userid' cannot be null
```

The solution would be to return a non empty ```uid``` value, e.g. ```0``` in the ```getUser()``` method.

PR #36, opened 2015-07-14
[BUGFIX] Return zero as UID in the Frontend authorization service

astehlik:
The getUser() method in the frontend helper
Tx_Solr_IndexQueue_FrontendHelper_AuthorizationService
returns a zero interger as UID in the user data array to prevent
database errors during the session initialization.

Resolves issue #35


irnnr:
Simple enough, thanks!

PR #37, opened 2015-07-14
[BUGFIX] Return zero as UID in the Frontend authorization service

astehlik:
The getUser() method in the frontend helper
Tx_Solr_IndexQueue_FrontendHelper_AuthorizationService
returns a zero interger as UID in the user data array to prevent
database errors during the session initialization.

Resolves issue #35 for the 3.x branch.


irnnr:
We're not doing a 3.0.2, thanks! (unless we rely have to)

peterkraume:
I strongly disagree with the decision not to accept any more pull requests for the 3.0.x branch. Although there may be no more TER releases of the 3.0.x branch, it might be handy for people (at least me) to have a working branch  with the latest fixes to rely on. Master proved to be unstable quite often in the past!

PR #43, opened 2015-07-20
Update jquery-ui.custom.css

senders:
there was an "s" at the end. If you merge css files you will get a white screen.


irnnr:
good catch, thanks. 
Probably ended up in there from a failed CMD+S

PR #45, opened 2015-07-31
ext_table.php 7.3.x compat

Mtillmann:
using the old style definition, backend crashes on "new content element" dialog


irnnr:
LGTM

PR #50, opened 2015-08-15
Move Apache_Solr_Document to namespaced version

theilm:
The double backslash in the makeInstance parameter causes an exception. While fixing this, I also moved to namespaced version of the class name.


theilm:
Sorry, just wanted to do too much at once. I will create a new pull requests that fixes only the backslash issue.

PR #51, opened 2015-08-16
Remove backslash at the beginning of makeInstance parameter

theilm:
Backslashes before the FQCN are not allowed any more in TYPO3 7.


irnnr:
Thanks!

Issue #54, opened 2015-08-25
Multiple Hooks for detectSerializedValue will override individual return value


AndreasA:
The Hook for detectSerializedValue will always override the detected value which means that even if the first returns TRUE as long as the last hook returns FALSE, it will still be false.

<pre>
if ($serializedValueDetector instanceof Tx_Solr_SerializedValueDetector) {
	$isSerialized = (boolean) $serializedValueDetector->isSerializedValue($indexingConfiguration, $solrFieldName);
}
</pre>

Not sure what the best solution is:
1.) Only execute hook if $isSerialized is FALSE as most likely the value should be serialized, if one hook determines serialization
2.) Pass $isSerialized to hook
3.) Allow NULL return and do not override value if NULL

PR #55, opened 2015-08-25
[BUGFIX] only execute detectSerializedValue if isSerialized is FALSE

AndreasA:
See https://github.com/TYPO3-Solr/ext-solr/issues/54

Not sure if this is the best solution, see possibilities in ticket but I think it is the only one without any breaking change.


AndreasA:
Any reason against merging?

irnnr:
Just didn't have time earlier to review. LGTM

PR #60, opened 2015-08-28
[BUGFIX] Implement enableRecursiveValueResolution for MM relations

markuskobligk:
Please see https://forge.typo3.org/issues/66179 for details.

I am not sure if the dupe check at the end is needed any longer here, in the maintenance release from TER it is in, in your master branch it's gone. Maybe this is done in a different function now or has been removed.

### DIFF
--- a/Classes/ContentObject/Relation.php
+++ b/Classes/ContentObject/Relation.php
@@ -264,6 +264,13 @@ class Relation {
 
 		$foreignTableLabelField = $this->resolveForeignTableLabelField($foreignTableTca);
 
+        // Remove the first option of foreignLabelField for recursion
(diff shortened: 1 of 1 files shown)


markuskobligk:
The last commit (aba7288) is a way better solution than the first one, since it works recursive with MM relations and even works if the last relation is not MM. Basically the main functiongetRelatedItems()  is invoked again with different parentContentObject and an adjusted foreignLabelField configuration.

irnnr:
Please make sure to use tabs and keep indentation clean

irnnr:
There was a usage of t3lib_div as well that I fixed

PR #62, opened 2015-09-02
fix missing content field in news

kvonspiczak:
the content field for a news entry wasn't indexed by using the static typoscript example for news, provided by the extension.


dkd-dobberkau:
Looks good to me. But someone else must test this too.

irnnr:
Can't confirm the issue. Also your patch would break indexing news records that use content elements for the bodytext.

Please provide further information on how to reproduce the issue if it still persists for you.

kvonspiczak:
I'm not using content elements for the bodytext.
I couldn't get the content field indexed in 7.3.1 and 7.4.0.
Whenever I tried it with the static typoscript provided by the extension, it wouldn't work. 
I needed to make this change to get it to work.

PR #66, opened 2015-09-16
[BUGFIX] CMS 7.5.x does not have FormEngine class

AndreasA:


irnnr:
Does this work with TYPO3 6.2?

AndreasA:
I have not removed the FormEngine class.
I first check if the class exists and if it does create it. Otherwise go on with the other things.

So it should work but I haven't tested it.

Classes/Backend/IndexingConfigurationSelectorField.php
```diff
@@ -215,9 +215,13 @@ protected function renderSelectCheckbox($items, $selectedValues) {
...
-		$selectFieldRenderer = $formEngine = GeneralUtility::makeInstance('TYPO3\\CMS\\Backend\\Form\\FormEngine');
-		if (!method_exists($selectFieldRenderer, 'getSingleField_typeSelect_checkbox')) {
-			if (class_exists('TYPO3\\CMS\Backend\\Form\\Element\\SelectElement')) {
+		$selectFieldRenderer = $formEngine = NULL;
+		if (class_exists('TYPO3\\CMS\\Backend\\Form\\FormEngine')) {
+			$selectFieldRenderer = $formEngine = GeneralUtility::makeInstance('TYPO3\\CMS\\Backend\\Form\\FormEngine');
+		}
+
+		if (!isset($selectFieldRenderer) || !method_exists($selectFieldRenderer, 'getSingleField_typeSelect_checkbox')) {
+			if (class_exists('TYPO3\\CMS\Backend\\Form\\Element\\SelectElement') && isset($formEngine)) {
```
sascha-egerer: There is an Backslash missing `TYPO3\\CMS\Backend`

AndreasA: Thx. I will change this later. Althought because of the single ' it doesn't really matter. It would still work but CGL is two.

AndreasA: Done

NeoBlack:
with a version check you don't need check each class. but general, is it important to use FormEngine here!? Maybe build the selector without FormEngine? only an idea...

AndreasA:
A version check is never a good idea if you can check for features.
Therefore class checks is better.
Not to mention you would still need the same ifs, just more version checks.

Though of course We could remove the 7.2 check now.

And yes building the selector manually would be possible too, I only improved the existing code.
Not rewrote it.

irnnr:
LGTM

PR #67, opened 2015-09-18
[BUGFIX] Change package name of TYPO3 CMS to typo3/cms

theilm:
At the moment, the package name used in composer.json is "typo3/cms-core" instead of "typo3/cms". Thus, a installation via composer is impossible. This little change fixes this.


theilm:
Sorry, again I wrote to quickly, apparently the problem is another one. I'll have a deeper look.

PR #69, opened 2015-09-19
Fix view helper class name prefix case

irnnr:
Fixes the view helper class prefix from lowercase **tx_solr** to ucwords **Tx_Solr**.

Change-Id: I17dc1ed41bfe3f3f8ce24b7a1b3895caa697f3ab


NeoBlack:
+1 / +1 works great

PR #72, opened 2015-09-21
[TASK] update autoload section

AndreasA:


AndreasA:
Actually, I am wondering do we even need the composer.lock? After all the extension will never work without cms etc. and therefore cannot be used as standlone.
Therefore, I would prefer to remove the composer.lock file

AndreasA:
The update to composer.json However should be done

irnnr:
Sounds good to me

PR #76, opened 2015-09-25
[BUGFIX] migration map loading in 7.5.x

AndreasA:
In 7.5.x the new class alias loader is used which requires Migration maps to be added differently.

I kept the old section because in 6.x the old one still applies and added the new one additionally.


irnnr:
👍

PR #80, opened 2015-09-30
[BUGFIX] Adjust backend values to TCEforms format

### LINKED ISSUE #79: TYPO3 7.5 "PHP Warning: array_flip() expects parameter 1 to be array, string given in ...." in Backend Module ->  Index Queue
When clicking on Index Queue with Solr Collections already initialized, i get the following error in the backend: 

    PHP Warning: array_flip() expects parameter 1 to be array, string given in /usr/www/users/fhpooz/typo3_src/typo3_src-7.5.0/typo3/sysext/backend/Classes/Form/Element/SelectCheckBoxElement.php line 45

The stack trace shows the following method call as being responsible for the warning: 

    ApacheSolrForTypo3\Solr\Backend\IndexingConfigurationSelectorField::renderSelectCheckbox(array, "")

I did a typo3 upgrade from 7.4 to 7.5 with already initalized queue (pages only) - indexing and rendering seems to work as expected for now

sorenmalling:
Fixes #79

### DIFF
--- a/Classes/Backend/IndexingConfigurationSelectorField.php
+++ b/Classes/Backend/IndexingConfigurationSelectorField.php
@@ -113,16 +113,9 @@ class IndexingConfigurationSelectorField {
 	 *  @return string Markup for the select field
 	 */
 	public function render() {
-			// transform selected values into the format used by TCEforms
-		$selectedValues = array();
-		foreach ($this->selectedValues as $selectedValue) {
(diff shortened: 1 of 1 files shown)


irnnr:
This those not work with TYPO3 6.2 however. We can merge this if it works for both versions.

sorenmalling:
#82 does a better conditional check than a version compare IMO

PR #82, opened 2015-10-05
[BUGFIX] correctly render selected items for 7.5 in index queue selector

AndreasA:
In 7.5 the selectorfield does not accept a comma separated list anymore but requires an array of values.


irnnr:
Instead of detecting the class, could we please use version_compare? This way it's easier to find when cleaning up legacy code. It worked pretty well when we removed 4.5 specific code...

sorenmalling:
Feel free to grab the condition from https://github.com/TYPO3-Solr/ext-solr/pull/83 if wanted

AndreasA:
No problem.
Will modify it accordingly

irnnr:
[meta] related to #79

AndreasA:
Modified check to version check.
Also modified the old checks to version checks and removed unnecssary checks like 7.2 because that one should really not be used anymore.

PR #83, opened 2015-10-06
[TASK] Make IndexingConfigurationSelectorField backward compatible

sorenmalling:


sorenmalling:
Closed in favour of #82 which does a better conditional check than a version compare IMO

PR #86, opened 2015-10-07
[TASK] Util::initializeTsfe call settingLocale

AndreasA:
I think it would be better to call settingLocale when initializing tsfe, therefore it will also not be required to be called in AbstractIndexer.


irnnr:
:+1:

AndreasA:
Just to let you know I will create a small update to this because settingLocale also has to be called when using initializeTsfe and returning it from the cache, as the locale settings are not stored in TSFE itself.

PR #87, opened 2015-10-07
[BUGFIX] settingLocale also needs to be called when retrieving TSFE from cache

AndreasA:
As settingLocale sets locale setting by using setlocale we alson eed to call it when retrieving TSFE from cache.

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -348,7 +348,9 @@ class Util {
 			$GLOBALS['TSFE']->getConfigArray();
 
 			$GLOBALS['TSFE']->settingLanguage();
-			$GLOBALS['TSFE']->settingLocale();
+			if (!$useCache) {
+				$GLOBALS['TSFE']->settingLocale();
+			}
 
 			$GLOBALS['TSFE']->newCObj();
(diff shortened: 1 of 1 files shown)


Classes/Util.php
```diff
@@ -348,7 +348,9 @@ public static function initializeTsfe($pageId, $language = 0, $useCache = TRUE)
 			$GLOBALS['TSFE']->getConfigArray();
 
 			$GLOBALS['TSFE']->settingLanguage();
-			$GLOBALS['TSFE']->settingLocale();
+			if (!$useCache) {
```
AndreasA: We don't need to execute settingLocale here if we use cache as it will be executed later anyway.

Classes/Util.php
```diff
@@ -361,6 +363,7 @@ public static function initializeTsfe($pageId, $language = 0, $useCache = TRUE)
 
 		if ($useCache) {
 			$GLOBALS['TSFE'] = $tsfeCache[$cacheId];
+			$GLOBALS['TSFE']->settingLocale();
```
AndreasA: To ensure locales are set correctly when getting tsfe from cache

AndreasA:
Btw. wouldn't it make more sense to move tsfeCache to a protected static member variable of the class?

PR #88, opened 2015-10-11
[BUGFIX] Handling of multiple mount points to the same page

astehlik:
The handling of mount points of pages is improved. A unique identifier
for the different mount points is added to prevent overriding
existing data in the index queue item table when there are multiple
mount points pointing to the same page.


Mabahe:
Nice! THX! An additional identifier was also my plan.

PR #89, opened 2015-10-11
[BUGFIX] Respect mount point in access rootline

astehlik:
Tx_Solr_Access_Rootline::getAccessRootlineByPageId() now accepts
an additional parameter for the current mount point.

In the page indexer the mount point parameter is used to build
correct access rootlines for mounted pages.


astehlik:
See forge issue https://forge.typo3.org/issues/59633

PR #90, opened 2015-10-11
[BUGFIX] Use mount point parameter in page document ID

astehlik:
The mount parameter that is used to access a page is now passed on
to the page indexer and will be added to the page document ID if
it is not empty. This allows a page to appear multiple times in the
index in its different mount point contexts.


astehlik:
See also forge issue: https://forge.typo3.org/issues/59438

PR #91, opened 2015-10-11
[FEATURE] foreign_match_fields in SOLR_RELATION content object

astehlik:
The foreign_match_fields configuration of foreign tables is now
considered in the SOLR_RELATION content object.


astehlik:
See forge issue https://forge.typo3.org/issues/61330

PR #92, opened 2015-10-11
[BUGFIX] Respect overlay settings of mount points

astehlik:
When a page is stored in the Backend and a mount point page points
to it with overlay enabled the record monitor will now also add
the overlayed page to the index queue.

Pages with disabled mount point overlay will now be indexed as
normal pages by adjusting the default additionalWhereClause config
option.


astehlik:
See forge issue https://forge.typo3.org/issues/59443

PR #100, opened 2015-10-21
[BUGFIX] PHP Fatal Error; t3lib_div should either be \t3lib_div:: or …

hnadler:
…better GeneralUtility::

* using t3lib_div in namespace context results in a fatal error (while indexing)
* should either be \t3lib_div:: or better GeneralUtility::


irnnr:
Thanks, must have been a leftover

PR #102, opened 2015-10-30
[BUGFIX] Load Viewhelper by Namespace if old class name not exist

markussom:
## Load ViewHelper by Namespace

If you use composer in TYPO3 CMS 7.x without legacy classes you get an exception because class can not found

### DIFF
--- a/Classes/Template.php
+++ b/Classes/Template.php
@@ -203,6 +203,12 @@ class Template {
 			$viewHelperIncludePath = ExtensionManagementUtility::extPath($extensionKey)
 				. $viewHelperPath . $possibleFilename;
 
+			// @todo only dirty solution
+			if(!class_exists($possibleClassName)) {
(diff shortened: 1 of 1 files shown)


irnnr:
How would/can you disable legacy class maps? Is there a setting for that?

markussom:
Yes if you use composer and disable compatibility6 layer

markussom:
on the other Hand, need tx_solr the old classnames?

irnnr:
It's still using the old class names as properly changing the extension to allow arbitrary namespaces would be too much effort considering we're moving to fluid at some point.

So when using composer, EXT:compatibility6 is disabled automatically?

markussom:
No. But it is disabled by default

irnnr:
I agree, it's hacky. If you want to put in some work on the current template engine feel free.

PR #108, opened 2015-11-04
[BUGFIX] Check if table is loaded in TCA

andreasfernandez:
Tables for indexing could be configured while depending extensions are not installed, causing SQL errors in `AbstractInitializer::buildSelectStatement()` because the table is not available in TCA then. This patch adds a check to verify whether the table being indexed is in TCA.

### DIFF
--- a/Classes/Backend/IndexingConfigurationSelectorField.php
+++ b/Classes/Backend/IndexingConfigurationSelectorField.php
@@ -169,7 +169,9 @@ class IndexingConfigurationSelectorField
                         $table = $solrConfiguration['index.']['queue.'][$name . '.']['table'];
                     }
 
-                    $indexingTableMap[$name] = $table;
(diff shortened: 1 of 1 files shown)


timoschmidt:
Hi Andreas, thx for your pull request. I've checked the current dev-master and could not reproduce this behaviour. Can you please check it and tell me how to reproduce? Then i can review and merge your pull request.

irnnr:
Yes, I can see this happening. However, I would also add logging an error if that happens that can be shown in the Solr backend module.

dkd-kaehm:
I think this one can be closed. There was big change after this PR:

Seems to be fixed in https://github.com/TYPO3-Solr/ext-solr/commit/0185911e3c2d50cac8929cfe938d74ad1f0a3ded 
. 
Or can it still happen, that TCA is not present, despite we are using new behaviour?

PR #109, opened 2015-11-04
[BUGFIX] >= 7.5 requires flushing of runtime cache for getIndpEnv

AndreasA:


AndreasA:
Btw. the same issue exists also in solrfal (v2.0.2)

timoschmidt:
Hi Andreas, thanks for your patch & pull request, i will check it soon.

Issue #104, opened 2015-11-03
numericRange facets broken with query keyword


Mabahe:
Since a2ff9e4c085cdd019cd245445407d0edc3f1e52e the rendered link in the hidden url field is wrong because $queryGetParameter is appended now instead of prepended like before in getQueryUrl().

Mabahe:
@irnnr WDYT, is it save to move $queryGetParameter to the beginning in additionalParams in getQueryLink() or could cause this another regression and we better need an option when calling getQueryLink() in getQueryUrl()?

PR #110, opened 2015-11-04
[BUGFIX] Restore behaviour of getQueryUrl in getQueryLink

Mabahe:
Since a2ff9e4c085cdd019cd245445407d0edc3f1e52e the rendered
link in the hidden url field is wrong because $queryGetParameter
is appended now instead of prepended like before in getQueryUrl().

This broke the generated url for numericRange facets with a query
keyword.

This patch restores the prior parameter order from getQueryUrl() in
getQueryLink().

Resolves: #104

### DIFF
--- a/Classes/Query/LinkBuilder.php
+++ b/Classes/Query/LinkBuilder.php
@@ -227,10 +227,10 @@ class LinkBuilder
             'useCacheHash' => false,
             'no_cache' => false,
             'parameter' => $this->linkTargetPageId,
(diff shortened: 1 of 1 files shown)


timoschmidt:
Thx for your pull request @Mabahe sorry that it tooked so long. I'll have a look on your pull request right now.

timoschmidt:
@Mabahe Looks good! Great catch! Thx for your patch and pull request

PR #113, opened 2015-11-06
[TASK] Implement hook in UsedFacetRenderer 

timoschmidt:
Added hook to overwrite the label of a used facet. Can be used like this:

```php
$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['solr']['processUsedFacetText'][] = 'MyClass';
```


irnnr:
Can we please require to implement an interface? f.e. UsedFacetOptionsRenderer

Issue #111, opened 2015-11-05
Lowercase synonyms


dkd-friedrich:
Since the lowercase filter is executed before the synonym filter, the synonyms have to to be inserted in lowercase.

The example in the backend module should be fixed, a hint added and the synonyms automatically lowercased.

dkd-friedrich:
Setting the ignoreCase option, could be an alternative

(https://cwiki.apache.org/confluence/display/solr/Managed+Resources#ManagedResources-Synonyms)

timoschmidt:
Hi Markus, thanks for reporting i'll have a look on this Issue.

I think we should:

 - Be able to keep the original spelling of the word


Currently is see two ways:

1: Store the synonyms and stopswords also in the TYPO3 database and lowercase on sync

2: Make use of your proposed ignoreCase flag to indicate to solr to ignore the casing


1. Has the advantage to have the possibility to restore the synonyms and stopwords even when the whole index was lost. Maybe a combination of both makes also sence.

timoschmidt:
During the analysis we've recognized that the schema is different an the handling of synonyms. For german lowercasing is happens before expanding the schema. For english afterwards.

We discussed and decided to implement the following strategy

1. First lowercasing in all schemata
2. Expanding synonyms afterwards
3. Store stopwords and synonyms lowercased in solr

I'll implement this a create a pull request  for this.

timoschmidt:
Merged to Master

PR #123, opened 2015-11-12
[BUGFIX] Fix for issue #111: Lowercase synonyms

timoschmidt:
* Implemented ignoreCase handling in synonyms and adapted backend module to turn it on or off.


timoschmidt:
We need another solution for this. Because ignoreCase is not changable after initialization and rewrite.

Issue #122, opened 2015-11-12
ResultTypeIcons not working anymore (from v3.0.1) as described in documentation


chvonrohr:
We implemented the ResultTypeIcons as shown in the documentation:
https://forge.typo3.org/projects/extension-solr/wiki/TemplatingResultTypeIcons

Since version 3.0.1 html is not rendered anymore, instead the html source is shown on the website.
In 3.0.0 this was working. Tested in latest TYPO3 6.2.15.

Do we have to find a new way to realize that or is it a bug?

dkd-dobberkau:
@timoschmidt can you check?

timoschmidt:
@dkd-dobberkau Hi olivier i will check this.

timoschmidt:
As a short update:

The problem seem to exist and was introduced with the following Issue:

https://forge.typo3.org/issues/39282

And pushed with this commit bf74217ef091ecda4f47c3216311ccae55f9cdab , since this is security related i would take the time think about a solution.

timoschmidt:
@chvonrohr 

I've checked this issue and for me it looks not so easy to fix. The previous mentioned commit introduced the esacping with htmlspecialchars for title and content. In general this is good because we want to avoid html in the content in XSS. But in my opinion the location where this is done is wrong. For me it makes sence to do the filtering right after the data was retrieved from solr. I am currently working on an implementation of that in an own branch:

https://github.com/timoschmidt/ext-solr/commits/bugfix/release-3.0.x/122-resultTypeIcons-not-working-anymore (branch name bugfix/release-3.0.x/122-resultTypeIcons-not-working-anymore)

Can you maybe check if this is working for your current project?

timoschmidt:
i've implemented most of the review feedback (except what i commented above). It would be nice, when somebody could review and approve this pull request.

timoschmidt:
Merged to master

PR #134, opened 2015-11-17
[BUGFIX] Move escaping from Template.php to the point right after the retrieval from solr

timoschmidt:
Implemented escaping of fields directly after retrieval from solr and added configuration for fields that should not be escaped "plugin.tx_solr.search.skipEscapeFields". And also removed htmlspecialchar calls from Template.php

Fix for issue #122

### DIFF
--- a/Classes/Search.php
+++ b/Classes/Search.php
@@ -312,15 +312,88 @@ class Search implements SingletonInterface
         return $this->getResponse()->response;
     }
 
+	/**
+	 * Returns a list of solr fields that could be retrieved without applying htmlspecialchars
+	 *
+	 * @return array
+	 */
+	protected function getSkipEscapeSolrFields()
+	{
+		$trustedFieldArray = array();
+		if(!isset($GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.']['search.']['skipEscapeFields']) ||
+			!is_string($GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.']['search.']['skipEscapeFields'])) {
+			return $trustedFieldArray;
+		}
+
+		$trustedFields = $GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.']['search.']['skipEscapeFields'];
+		$trustedFieldArray = GeneralUtility::trimExplode(",", $trustedFields);
+
+		return $trustedFieldArray;
+	}
+
     /**
-     * Gets the result's documents
-     *
-     * @return \Apache_Solr_Document[]
+     * @deprecated use getResultEscapedDocuments now when possible
+     * @return array
      */
-    public function getResultDocuments()
-    {
-        return $this->getResponseBody()->docs;
-    }
+	public function getResultDocuments()
+	{
+		return $this->getResponseBody()->docs;
+	}
+
+	/**
+	 * Returns all result documents but applies htmlspecialchars on all fields retrieved
+	 * from solr except the configured fields in
+	 *
+	 * plugin.tx_solr.search.skipEscapeFields
+     *
+     * @return array
+	 */
+	public function getResultDocumentsEscaped()
+	{
+		return $this->applyHtmlSpecialCharsOnAllFields($this->getResponseBody()->docs);
+	}
+
+	/**
+	 * This method is used to apply htmlspecialchars on all document fields that
+	 * are not configured to be secure. Secure mean that we know where the content is comming from.
+	 *
+	 * @param array $docs
+	 * @return array
+	 */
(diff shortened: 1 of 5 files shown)


timoschmidt:
We should also maybe backport this to 3.x.x after review and merge

Configuration/TypoScript/Solr/setup.txt
```diff
@@ -106,6 +106,10 @@ plugin.tx_solr {
 	}
 
 	search {
+	    // fields that are allowed to contain html and should be skipped during the escaping after retrieval from solr
+	    // by default all fields get escaped and nothing will be skipped
+	    skipEscapeFields =
```
irnnr: just make sure indentation is correct

Classes/Search.php
```diff
@@ -312,15 +312,88 @@ public function getResponseBody()
...
+	/**
+	 * This method is used to apply htmlspecialchars on all document fields that
+	 * are not configured to be secure. Secure mean that we know where the content is comming from.
+	 *
+	 * @param array $docs
+	 * @return array
+	 */
+	protected function applyHtmlSpecialCharsOnAllFields(array $docs)
+	{
+		$trustedSolrFields = $this->getSkipEscapeSolrFields();
```
irnnr: "trusted" is better wording, how about calling the methods and the configuration option "trusted" as well? ... instead of "skipped"

Classes/Search.php
```diff
@@ -312,15 +312,88 @@ public function getResponseBody()
...
+
+	/**
+	 * Returns all result documents but applies htmlspecialchars on all fields retrieved
+	 * from solr except the configured fields in
+	 *
+	 * plugin.tx_solr.search.skipEscapeFields
+     *
+     * @return array
+	 */
+	public function getResultDocumentsEscaped()
```
irnnr: Why this wrapper method?

Classes/Search.php
```diff
@@ -312,15 +312,88 @@ public function getResponseBody()
...
+
+		return $trustedFieldArray;
+	}
+
     /**
-     * Gets the result's documents
-     *
-     * @return \Apache_Solr_Document[]
+     * @deprecated use getResultEscapedDocuments now when possible
+     * @return array
```
irnnr: Please do not change phpDoc; `\Apache_Solr_Document[]` is more precise

Classes/Search.php
```diff
@@ -312,15 +312,88 @@ public function getResponseBody()
...
+		$trustedFieldArray = GeneralUtility::trimExplode(",", $trustedFields);
+
+		return $trustedFieldArray;
+	}
+
     /**
-     * Gets the result's documents
-     *
-     * @return \Apache_Solr_Document[]
+     * @deprecated use getResultEscapedDocuments now when possible
```
irnnr: s/when possible/instead

Classes/Search.php
```diff
@@ -312,15 +312,88 @@ public function getResponseBody()
...
     }
 
+	/**
+	 * Returns a list of solr fields that could be retrieved without applying htmlspecialchars
+	 *
+	 * @return array
+	 */
+	protected function getSkipEscapeSolrFields()
+	{
+		$trustedFieldArray = array();
```
irnnr: Please no type information in variables, just `$trustedFields` is enough.

irnnr:
Looks good to me pending changes

irnnr:
Feel free to back port to 3.0.x, but I don't see a need for it. There's no plan to have another 3.0.x release if it's not needed otherwise.

Classes/Search.php
timoschmidt: I agree, i also though about the two variants. I'll change this

Classes/Search.php
timoschmidt: For two reasons:

1. The behaviour is changing when we change it directrly. Therefore i've added a new method and marked the new as deprecated and update all usages.

2. Also a boolean parameter in the signature would be bad style: ( http://programmers.stackexchange.com/questions/147977/is-it-wrong-to-use-a-boolean-parameter-to-determine-behavior) or over time hard to maintain

Therefore i think this was the best choice

Classes/Search.php
timoschmidt: I'll fix this

Classes/Search.php
timoschmidt: Also a valid point i'll change it

dkd-friedrich:
Worked for me

Classes/Search.php
irnnr: Agree with both, but it's not what I was wondering about. Anyway, I added a counter part method so it makes more sense I hope `getResultDocumentsRaw()`

PR #139, opened 2015-11-19
Bugfix/111 lowercase synonyms and stopwords

timoschmidt:
In the following pull request we've:

- Applied the lowercasing before expanding the synonyms in all schemata
- Lowercase stopwords and synonyms before saving them to solr

Stopwords and Synonyms need to be recreated afterwards.


dkd-friedrich:
Looks good to me.

irnnr:
I actually would have liked to review this first.

dkd-dobberkau:
we want to keep the speed.

PR #140, opened 2015-11-24
[BUGFIX] Assign undefined variable $pageId

andreasfernandez:
Assign the pid of the currently processed record to the undefined ``$pageId`` variable.


timoschmidt:
Thx for your pull request. I'll check it now!

timoschmidt:
I think we need to refine the patch because after applying the patch i could not index a page and get a queue error for each.

timoschmidt:
Thanks for your contribution. Checked it with my system and everything looks good now.

irnnr:
This was a long known issue and left open exactly because of the side effect. But thanks for fixing it/finding a fix for the side effect! 
We'll have to keep an eye on it nevertheless as this involves how the hash is built that verifies the indexer request is legit, preventing DOS attacks through indexer requests as these are uncached.

Issue #142, opened 2015-11-25
[BUG] ApacheSolrForTypo3\Solr\Controller|\Administration controller should reset site when unexisting page is retrieved


timoschmidt:
When a user copies a page tree and the page is referenced as site for the backend module in the backend users uc field, the stored site should be resetted.

PR #145, opened 2015-11-25
Sites of backend user should be reset when not valid

timoschmidt:
- Reinitialize Site in BackendModule when invalid site was retrieved from backend users uc data.
- Improved error message when no site is initialized.

Fixes: #142

### DIFF
--- a/Classes/Controller/AdministrationController.php
+++ b/Classes/Controller/AdministrationController.php
@@ -106,6 +106,7 @@ class AdministrationController extends ActionController
         ResponseInterface $response
     ) {
         $this->moduleData = $this->moduleDataStorageService->loadModuleData();
+
         try {
             parent::processRequest($request, $response);
             $this->moduleDataStorageService->persistModuleData($this->moduleData);
@@ -129,7 +130,7 @@ class AdministrationController extends ActionController
         }
 
         $rootPageId = $this->site->getRootPageId();
-        if($rootPageId > 0 && !Util::hasPage($rootPageId)) {
+        if($rootPageId > 0 && !Util::pageExists($rootPageId)) {
             $this->resetStoredSiteToFirstAvailableSite();
         }
 
@@ -175,7 +176,7 @@ class AdministrationController extends ActionController
         $request->setControllerName($activeModuleDescription['controller'] . 'Module');
         $request->setControllerActionName('index');
 
-        if( !is_null($this->site) ) {
+        if ( !is_null($this->site) ) {
             $request->setArgument('site', $this->site);
         }
 
@@ -217,7 +218,7 @@ class AdministrationController extends ActionController
     public function setSiteAction($site)
     {
         $site = Site::getSiteByPageId((int)$site);
-        $this->storeSiteAndResetCore($site);
+        $this->setSiteAndResetCore($site);
 
         $this->forwardHome();
     }
@@ -278,7 +279,7 @@ class AdministrationController extends ActionController
     /**
      * @param $site
      */
-    protected function storeSiteAndResetCore($site)
+    protected function setSiteAndResetCore($site)
     {
         $this->moduleData->setSite($site);
         // when switching the site, reset the core
@@ -294,7 +295,7 @@ class AdministrationController extends ActionController
     protected function resetStoredSiteToFirstAvailableSite()
     {
         $site = Site::getFirstAvailableSite();
-        $this->storeSiteAndResetCore($site);
+        $this->setSiteAndResetCore($site);
     }
 }
 
--- a/Classes/Controller/AdministrationController.php
+++ b/Classes/Controller/AdministrationController.php
@@ -25,6 +25,7 @@ namespace ApacheSolrForTypo3\Solr\Controller;
  ***************************************************************/
 
 use ApacheSolrForTypo3\Solr\Site;
+use ApacheSolrForTypo3\Solr\Util;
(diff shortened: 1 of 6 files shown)


Classes/Util.php
```diff
@@ -716,5 +716,21 @@ public static function getAllowedPageTypes($pageId)
...
     }
+
+    /**
+     * Method to check if a page exists.
+     *
+     * @param integer $pid
+     *
+     * @return bool
+     */
+    public static function hasPage($pid) {
```
irnnr: Can we call this pageExists()

irnnr: also, please try to avoid abbreviations $pid -> $pageId

Resources/Private/Layouts/Default.html
```diff
@@ -7,7 +7,19 @@
...
 				<div class="left">
-					<solr:backend.menu.SiteSelectorMenu name="site" />
+
+					<solr:backend.sites>
+						<f:if condition="{hasSites}">
+							<f:then>
+								<solr:backend.menu.SiteSelectorMenu name="site" />
+							</f:then>
+							<f:else>
+								No solr connections initialized!
```
irnnr: When talking about the Solr server, please use uppercase S "Solr". When talking about the extension, it's EXT:solr with lowercase s.

Classes/Controller/AdministrationController.php
```diff
@@ -105,7 +106,6 @@ public function processRequest(
         ResponseInterface $response
     ) {
         $this->moduleData = $this->moduleDataStorageService->loadModuleData();
-
```
irnnr: I really like my whitespace :)

Classes/Controller/AdministrationController.php
```diff
@@ -166,7 +174,11 @@ protected function invokeModuleController()
         $request->setControllerExtensionName(ucfirst($activeModuleDescription['extensionKey']));
         $request->setControllerName($activeModuleDescription['controller'] . 'Module');
         $request->setControllerActionName('index');
-        $request->setArgument('site', $this->site);
+
+        if( !is_null($this->site) ) {
```
irnnr: Is this PSR-2? You can setup PhpStorm to check/correct/produce PSR-2 compliant code automatically.

Classes/Controller/AdministrationController.php
```diff
@@ -265,5 +274,27 @@ protected function forwardToModule($module, $moduleAction = '')
 
         $this->forward('index');
     }
+
+    /**
+     * @param $site
+     */
+    protected function storeSiteAndResetCore($site)
```
irnnr: would prefer to call this setSite...

Classes/Site.php
```diff
@@ -162,6 +170,19 @@ public static function getAvailableSites()
     }
 
     /**
+     * Returns the first available Site.
+     *
+     * @param boolean $stopOnInvalidSite
```
irnnr: What's the intention of this new parameter? I don't see it used....

Classes/ViewHelpers/Backend/SitesViewHelper.php
```diff
@@ -0,0 +1,69 @@
+<?php
+namespace ApacheSolrForTypo3\Solr\ViewHelpers\Backend;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2013-2015 Ingo Renner <ingo@typo3.org>
```
irnnr: Feel free to use your name

Resources/Private/Layouts/Default.html
irnnr: In general I think I have to check the empty state again - what is shown the user hits the module the first time. Would be nice to offer a little message and a button to initialize the sites.

Resources/Private/Layouts/Default.html
```diff
@@ -23,12 +35,32 @@
...
+								<p>Please check the following:</p>
+								<ul>
+									<li>Your page has a domain record.</li>
+									<li>Your site has a page, marked as root page.</li>
+									<li>Your page has included the solr typoscript configuration.</li>
+									<li>You've initialized the solr connections in the backend with the menu item "Initialize solr connections"</li>
+									<li>Afterwards please check if the reports module is showing any errors.</li>
+								</ul>
+
+								<p>When all of these steps are done check this module again.</p>
```
irnnr: :+1:

Classes/Util.php
timoschmidt: Changed both

Resources/Private/Layouts/Default.html
timoschmidt: Also changed

Classes/Controller/AdministrationController.php
timoschmidt: Also changed

Classes/Controller/AdministrationController.php
timoschmidt: Also changed

Classes/Site.php
timoschmidt: The intension is if the Exception from the constructor is bubbled up or not. Normally i would say it is good the the exception get's thrown, but when getting the collection in a loop it breaks the look, which is not allways wanted. Sometimes you just want to skip this item, which is the default now.

Classes/ViewHelpers/Backend/SitesViewHelper.php
timoschmidt: Done

Issue #155, opened 2015-11-30
SQL error when calling getDestinationMountPropertiesByRootLine


timoschmidt:
Issue in master of last Friday: when editing the page properties of the site root (UID 1) this is shown when saving the modifications:

array(4 items)
   caller => "TYPO3\CMS\Core\Database\DatabaseConnection::exec_SELECTquery" (60 chars)
   ERROR => "You have an error in your SQL syntax; check the manual that corresponds to y
      our MariaDB server version for the right syntax to use near 'OR (mount_pid=1
       AND mount_pid_ol=1)) AND doktype = 7 AND no_search = 0 AND pages' at line 1" (228 chars)
   lastBuiltQuery => "SELECT uid, uid AS mountPageDestination, mount_pid AS mountPageSource, mount
      _pid_ol AS mountPageOverlayed FROM pages WHERE ( OR (mount_pid=1 AND mount_p
      id_ol=1)) AND doktype = 7 AND no_search = 0 AND pages.deleted=0" (215 chars)
   debug_backtrace => "call_user_func#21 // {closure}# // TYPO3\CMS\Backend\Http\Application->run#2
      0 // TYPO3\CMS\Core\Core\Bootstrap->handleRequest#94 // TYPO3\CMS\Backend\Ht
      tp\RequestHandler->handleRequest#302 // TYPO3\CMS\Backend\Http\RequestHandle
      r->dispatch#94 // TYPO3\CMS\Backend\Http\RouteDispatcher->dispatch#161 // ca
      ll_user_func_array#54 // TYPO3\CMS\Backend\Controller\EditDocumentController
      ->mainAction# // TYPO3\CMS\Backend\Controller\EditDocumentController->proces
      sData#1977 // TYPO3\CMS\Core\DataHandling\DataHandler->process_datamap#582 /
      / TYPO3\CMS\Core\DataHandling\DataHandler->hook_processDatamap_afterDatabase
      Operations#1330 // ApacheSolrForTypo3\Solr\IndexQueue\RecordMonitor->process
      Datamap_afterDatabaseOperations#898 // ApacheSolrForTypo3\Solr\IndexQueue\Re
      cordMonitor->updateMountPages#281 // ApacheSolrForTypo3\Solr\IndexQueue\Reco
      rdMonitor->getDestinationMountPropertiesByRootLine#436 // TYPO3\CMS\Core\Dat
[...]

PR #160, opened 2015-12-01
[BUGFIX] Added regression test and fix for issue #155 (SQL error when calling getDestinationMountPropertiesByRootLine)

timoschmidt:
* Added fix to avoid generating broken SQL when page is saved in the backend with propery set "is_rootpage"
* Added regression test

### DIFF
--- a/Classes/IndexQueue/RecordMonitor.php
+++ b/Classes/IndexQueue/RecordMonitor.php
@@ -452,14 +452,14 @@ class RecordMonitor
     protected function getDestinationMountPropertiesByRootLine(array $rootLine)
     {
         $mountPages = array();
-        $subPageUids = array();
+        $rootLineParentPageIds = array();
 
         $currentPage = array_shift($rootLine);
         $currentPageUid = (int)$currentPage['uid'];
 
         if (!empty($rootLine)) {
             foreach ($rootLine as $pageRecord) {
-                $subPageUids[] = $pageRecord['uid'];
+                $rootLineParentPageIds[] = $pageRecord['uid'];
 
                 if ($pageRecord['is_siteroot']) {
                     break;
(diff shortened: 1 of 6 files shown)


timoschmidt:
Need refinements for 7.6.0.

timoschmidt:
There were refinements needed that came up during testing with 7.6.0 and dev-master:

array(4 items)
   caller => "TYPO3\CMS\Core\Database\DatabaseConnection::exec_INSERTquery" (60 chars)
   ERROR => "Field 'errors' doesn't have a default value" (43 chars)
   lastBuiltQuery => "INSERT INTO tx_solr_indexqueue_item (root,item_type,item_uid,changed,indexin
      g_configuration) VALUES ('1','pages','1','0','pages')" (129 chars)
   debug_backtrace => "__phpunit_run_isolated_test#389 // PHPUnit_Framework_TestCase->run#268 // PH
      PUnit_Framework_TestResult->run#702 // PHPUnit_Framework_TestCase->runBare#6
      01 // PHPUnit_Framework_TestCase->runTest#746 // ReflectionMethod->invokeArg
      s#881 // ApacheSolrForTypo3\Solr\Tests\Functional\IndexQueue\RecordMonitorTe
      st->canUpdateRootPageRecordWithoutSQLErrorFromMountPages# // ApacheSolrForTy
      po3\Solr\IndexQueue\RecordMonitor->processCmdmap_postProcess#112 // ApacheSo
      lrForTypo3\Solr\IndexQueue\Queue->updateItem#132 // ApacheSolrForTypo3\Solr\
      IndexQueue\Queue->addItem#420 // TYPO3\CMS\Core\Database\DatabaseConnection-
      >exec_INSERTquery#500 // TYPO3\CMS\Core\Database\DatabaseConnection->debug#2
      23" (686 chars)

The "errors" field is not allowed to be empty by the database definition (text), so i initialized it with an empty string when queueItem is created.

Tests/Functional/FunctionalTest.php
```diff
@@ -49,25 +50,41 @@
...
+
+    /**
+     * Returns the directory on runtime.
+     *
+     * @return string
+     */
+    protected function getRuntimeDirectory()
+    {
+        $rc = new \ReflectionClass(get_class($this));
+        return dirname($rc->getFileName());
```
irnnr: what's wrong with `__DIR__`?

irnnr:
Just a few nitpicks and questions, otherwise LGTM

Tests/Functional/FunctionalTest.php
timoschmidt: Hi Ingo, this is because you want to place fixtures by convention beside every testcase in the "Fixtures" folder __DIR__ would only be the folder beside the AbstractTestCase and i think it is better to place the fixtures right beside the TestCase. But in general i would agree to avoid using reflection whenever possible.

Classes/Util.php
```diff
@@ -330,6 +330,7 @@ public static function getConfigurationFromPageId(
                     $initializedPageSelect = true;
                 }
 
+                    /** @var $tmpl \TYPO3\CMS\Core\TypoScript\ExtendedTemplateService */
```
timoschmidt: Nice hint! I'll install that, but i would maybe think about to change this in the later versions to make the start easier for others, we can also use the ::name Variante then when we drop 6.2.x support. I'll remove it for now.

Classes/IndexQueue/RecordMonitor.php
```diff
@@ -471,13 +471,15 @@ protected function getDestinationMountPropertiesByRootLine(array $rootLine)
             return $mountPages;
         }
 
-        $pageQuery = '';
-        if (!empty($subPageUids)) {
-            $pageQuery = 'mount_pid IN(' . implode(',', $subPageUids) . ')';
+        $pageQuerys = array();
+        if (!empty($rootLineParentPageIds)) {
+            $pageQuerys[] = '(mount_pid IN(' . implode(',', $rootLineParentPageIds) . '))';
```
timoschmidt: Changed

Classes/Util.php
irnnr: if you mean `::class`, yes definitely!

Tests/Functional/FunctionalTest.php
irnnr: ok, makes sense.

Classes/IndexQueue/RecordMonitor.php
irnnr: forgot to commit?

PR #162, opened 2015-12-01
[FEATURE] Add signal after IndexQueueWorkerTask::indexItems

Mabahe:
Usefull to add postprocessing after the indexing. Some cache clearing in my case.

### DIFF
--- a/Classes/Task/IndexQueueWorkerTask.php
+++ b/Classes/Task/IndexQueueWorkerTask.php
@@ -109,6 +109,18 @@ class IndexQueueWorkerTask extends AbstractTask implements ProgressProviderInter
                 );
             }
         }
+        $this->emitAfterIndexItemsSignal($itemsToIndex);
+    }
+
+    /**
+     * Emits a signal after all items was indexed
+     *
+     * @param array $itemsToIndex
+     */
+    protected function emitAfterIndexItemsSignal($itemsToIndex)
+    {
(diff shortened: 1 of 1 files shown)


irnnr:
Good point, but it's not yet what you're asking for (I guess).
The way it's implemented in your PR right now he signal will fire every time a batch of items of size `$limit` has been indexed.

If you actually want a signal after **all** items have been indexed you'd need to have a more complicated check; Something like is `count($itemsToIndex) === 0` and have we been indexing items on the previous execution.

But in general I'd agree. I believe I was in the need for something like that myself before, but can't remember what happened to that plan.

Mabahe:
Batch of items, for sure. The "all" is "cheese". Will change the commit message.

timoschmidt:
@Mabahe Thanks for your PR. Currently we are doing the last commits for 3.1.0 and i think we can merge this after the release of 3.1.0. Just to let you know.

Mabahe:
on hold; will update the patch!

Mabahe:
Updated. It's usefull to also have the task itself available in the slot, to get e.g. the site.

timoschmidt:
Check it with custom implemented slot after your refinements. Looks good, thanks for the patch!
I'll add a test for this feature and also a signal, when the queue is empty.

irnnr:
Oh yeah, a signal for reaching "zero index queue" would be great. I think they would go really well together and would solve the initial description of having indexed everything.

PR #178, opened 2015-12-09
Fixed deprecations of getPageRenderer and increased min TYPO3 Version to 7.6.0 (Fixes #129)

### LINKED ISSUE #129: Check for deprecated methods
Check whether we're using any deprecated methods and replace them with successors.

This also includes things that have been moved to EXT:compatibility6. We should not rely on their existence.

timoschmidt:
Thise PR contains:
- Fix usage of deprecated method calls that will be removed by TYPO3 8
- Drop 6.2.15 from travis ci since we will not support it in 3.2.0 and also some of the new methods have been introduced in 7.x
- Drop BackendIconViewHelper and use IconViewHelper from TYPO3 Core, since our ViewHelper used the SpriteManager which is deperacted
- Register Backend Icons using key convention "extensions-solr-module-<modulename>" to render it => Needs change in ext_tika after the release since IconRegistration using the SpriteManager is deprecated.


.travis.yml
```diff
@@ -1,28 +1,16 @@
 language: php
 
 php:
-  - 5.3
-  - 5.4
   - 5.5
   - 5.6
```
irnnr: can we add 7.0? TYPO3 7.6 lists it as supported, too http://typo3.org/download/

irnnr: That can be another issue/PR to make things work on PHP 7

Classes/Plugin/PluginBase.php
```diff
@@ -238,12 +246,12 @@ public function pi_loadLL()
...
             // Read the strings in the required charset (since TYPO3 4.2)
-            $this->LOCAL_LANG = GeneralUtility::readLLfile($basePath,
-                $this->LLkey, $GLOBALS['TSFE']->renderCharset);
+            $this->LOCAL_LANG = $this->languageFactory->getParsedData($basePath,
+                $this->LLkey, $GLOBALS['TSFE']->renderCharset, 3);
             $alternativeLanguageKeys = GeneralUtility::trimExplode(',',
                 $this->altLLkey, true);
             foreach ($alternativeLanguageKeys as $languageKey) {
-                $tempLL = GeneralUtility::readLLfile($basePath, $languageKey);
+                $tempLL = $this->languageFactory->getParsedData($basePath, $languageKey, $GLOBALS['TSFE']->renderCharset, 3);
```
irnnr: do we still need renderCharset? Everything should be utf8.

Tests/Functional/Fixtures/test_template.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+	<meta charset="UTF-8">
+	<title>Title</title>
+</head>
+<body>
+
+	<!-- ###SOLR_TEST### begin -->
+	<div class="tx-solr">###CROP:hallowelt|5|.###</div>
```
irnnr: english ... ;)

Tests/Functional/TemplateTest.php
```diff
@@ -0,0 +1,85 @@
...
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+/**
+ * TestCase to check if the template parsing works as expected
+ *
+ * @author Timo Schmidt
+ * @package TYPO3
+ * @subpackage solr
+ */
+class TemplateTest extends FunctionalTest
```
irnnr: I wouldn't waste any time on testing the tempting system knowing that Fluid is next...

irnnr:
ok, pending feedback re renderCharset

.travis.yml
timoschmidt: During the implementation i allready tried it and i think it makes sence, but there are issues left. To not blow up the pull request, i think adding a seperate issue, as you proposed is good.

Classes/Plugin/PluginBase.php
timoschmidt: I would keep it, because in the core it is used like this, even when everybody should allready use UTF-8.

Tests/Functional/Fixtures/test_template.html
timoschmidt: I'll change that!

Tests/Functional/TemplateTest.php
timoschmidt: Yes. I think with this it's enough, just to see if the templating works with the replacement of the deprecated method calls.

Classes/Plugin/PluginBase.php
irnnr: ok

PR #183, opened 2015-12-14
Improve functional test suite

timoschmidt:
- Index into real solr server (installed in bootstrap with install script)
   - We can see automatically if the indexing is working
   - If the installerscript is working and compatible with the code
- Check if indexing requests are triggered correctly
   - We can see if indexing requests in the frontend get triggered as expected


.travis.yml
```diff
@@ -4,6 +4,8 @@ php:
   - 5.5
   - 5.6
 
+sudo: true
```
irnnr: Please investigate a way to avoid using sudo. Using sudo results in the Travis tests being executed on a different/slower platform.

timoschmidt: The solr install script requires sudo, we can:
- adapt adapt the install script
- install solr in a different way

but like this, we test the install script as every user would use it, and skipping the test is not an option.

irnnr: I know :)
Not saying it is an immediate issue. Just keep it on your mind. It's more important to cache the Tomcat/Solr installation on Travis / check for it's existence. 20 minutes to run our few tests is simply unacceptable.

PR #187, opened 2015-12-16
[FEATURE] Add link header (prev & next) with pagebrowser

kalypso63:
Add Typoscript option to display facet with a minimum count options. I think it would be helpful if you want to hide facet with only one available option. 
`plugin.tx_solr.search.faceting.facets.<facetName>.showWithMinimumOption`


irnnr:
Did you check this option?
https://forge.typo3.org/projects/extension-solr/wiki/Tx_solrsearch#facetingminimumCount

kalypso63:
It's not really what i want. This option `faceting.minimumCount` display facet option with minimum count results. My PR offer a way to display facet if number of available options is greater or equal to typoscript option  `plugin.tx_solr.search.faceting.facets.<facetName>.showWithMinimumOption`. It would be helpful if i want to hide facet with only one option.

timoschmidt:
@irnnr I think we can merge this. What do you think?

kalypso63:
+1. I think that might be helpful

sascha-egerer:
IMHO that should be done in the view. The implementation is really misleading. Calling a `isEmpty` which gives a `true` value if the facet is not empty but has not more options than defined in the typoscript value. Dont like that...

irnnr:
I still do not understand the issue you're trying to solve. It'd be great if you could provide an example. What is the current behavior, what is your expected behavior.

Besides that, the PR as it is right now will not be merged as it changes the behavior of isEmpty() in a incorrect way.

kalypso63:
My PR allows to display facet with a minimum available options after a search. I do not need to display a facet with 1 only 1 available option.

irnnr:
Can you please provide an example. **What is the current behavior, what is your expected behavior.**

irnnr:
Specifically, I still don't understand how what you want to achieve is different from the existing option https://forge.typo3.org/projects/extension-solr/wiki/Tx_solrsearch#facetingminimumCount

kalypso63:
It's not really the same behavior. Your option (faceting.minimumCount) indicates the minimum counts for facet fields should be included in the response but mine allows to hide a facet if the response indicates it have only 1 option available per example (so  my typoscript option should be plugin.tx_solr.search.faceting.facets.<myfacet>.showWithMinimumAvailableOptions = 2).

The current behavior display all facets even if each facet returns 1 available option.

timoschmidt:
@kalypso63 looks good to me beside two things:

1. You are only checking for an int in the comparison, what happend when it is explicitly turned off be configured "0"?
2. Maybe you can check the intension in the TypoScript

irnnr:
@kalypso63 Please never never never reuse a PR by changing its title. Now all the comments are totally unrelated. Please open a new PR for your pagination change

Issue #53, opened 2015-08-17
pages:extendToSubpages not recognized


seirerman:
When I deactivate a page and additionally set the flag extendToSubpages to 1 (quite handy if I want to hide a big part of a website), its subpages will be deactivated, too.

But Solr still lists them in the search results, which leads to 404 errors.
Since the records of the subpages aren't changed, Solr doesn't notice this, and the scheduler task remains at 100%.

Clearing the index and reindexing all pages at least removes the hidden pages from the search results, but also leads to indexing errors for all those subpages:

```
1319116885: exception 'RuntimeException' with message 'Failed to execute Page Indexer Request. See log for details. Request ID: 55c1b8c640395' in /var/www/html/t3kons/typo3conf/ext/solr/Classes/IndexQueue/PageIndexerRequest.php:163
Stack trace:
#0 /var/www/html/t3kons/typo3conf/ext/solr/Classes/IndexQueue/PageIndexer.php(409): Tx_Solr_IndexQueue_PageIndexerRequest->send('https://www.kon...')
#1 /var/www/html/t3kons/typo3conf/ext/solr/Classes/IndexQueue/PageIndexer.php(55): Tx_Solr_IndexQueue_PageIndexer->getAccessGroupsFromContent(Object(Tx_Solr_IndexQueue_Item), 0)
#2 /var/www/html/t3kons/typo3conf/ext/solr/Scheduler/IndexQueueWorkerTask.php(115): Tx_Solr_IndexQueue_PageIndexer->index(Object(Tx_Solr_IndexQueue_Item))
#3 /var/www/html/t3kons/typo3conf/ext/solr/Scheduler/IndexQueueWorkerTask.php(78): Tx_Solr_Scheduler_IndexQueueWorkerTask->indexItem(Object(Tx_Solr_IndexQueue_Item))
#4 /var/www/html/t3kons/typo3conf/ext/solr/Scheduler/IndexQueueWorkerTask.php(57): Tx_Solr_Scheduler_IndexQueueWorkerTask->indexItems()
#5 /var/www/typo3/typo3-6.2.14/typo3/sysext/scheduler/Classes/Scheduler.php(148): Tx_Solr_Scheduler_IndexQueueWorkerTask->execute()
[...]

PR #192, opened 2015-12-21
[Bugfix] Added functionallity to resolve affected pages when "extendToSubpages" is set and "hidden" flag is changed

timoschmidt:
Added functionality apply recursive changes in RecordMonitor and GarbageCollector when a combination of values in the Record and ChangeSet match to a configured configuration.

* Added tests for RecordMonitor and GarbageCollector
* Implemented recursive indexing and deletion, when "extendToSubpage" and "hidden" occures in the relevant combination
* Refactored RecordMonitor and GarbageCollector to have a shared BaseClass "AbstractDataHandlerListener"

Fixes: #53

### DIFF
--- a/Classes/AbstractDataHandlerListener.php
+++ b/Classes/AbstractDataHandlerListener.php
@@ -44,7 +44,7 @@ abstract class AbstractDataHandlerListener
     /**
      * @return array
      */
-    private function getAllRelevantFieldsForCurrentState()
+    protected function getAllRelevantFieldsForCurrentState()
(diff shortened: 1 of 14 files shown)


kraftb:
Something similar happens when moving a branch with subpages (branch). The "tstamp" of the moved pages get updated by the DataHandler but not the "tstamp" fields of the subpages. The RecordMonitor will also just update the page being moved but not all pages in the branch.

I will open an issue.

PR #207, opened 2016-01-20
Bugfix/205 minor config issue in solr 4.10.4

timoschmidt:
- Fixed wrong path in solrconfig.xml
- Set TYPO3 Version to 7.6.2 in .travis.yml


irnnr:
Hey Timo, if you change solrconfig or the schema you also have to update the respective name property and update the status check so that the status report can notify users of an updated configuration.

timoschmidt:
@irnnr Thanks for pointing this out, i'll do this later

irnnr:
:+1:

Issue #163, opened 2015-12-01
Streamline typoscript configuration handling


Mabahe:
I still have the problem, that there is a lot of typoscript configuration that is used by direct access of 
`$GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.']`. Especially in context of the content objects it is very annoying, because it is  mostly not possible to handle a different configuration for a special solr based content element, defined by
```
10 = COA
10 {

	10 < plugin.tx_solr_PiResults_Results
	10 {
		search {
			ignored, like e.g. faceting configurations
		}
	}
}
```

irnnr:
I understand the problem, but what would you suggest to fix it?

Mabahe:
Too fast ;-) Here is the patch. Waiting with a pull request for your feedback. E.g. if we should call the object TyposciptConfiguration instead of Configuration.

irnnr:
ok, I left some comments :)

irnnr:
I think I actually tried to start something like that in EXT:tiki, but eventually did not follow through.

Mabahe:
@timoschmidt @irnnr so here is the squashed and optimized commit. I renamed and reorganized some parts and changed the usage of the moved Util functions, which are deprecated now. It is not yet tested in the project branch I implemented this for. I will create a PR after a successfull test.

timoschmidt:
@Mabahe thanks for you effort, looks good for me. Looking forward for your pull request

PR #210, opened 2016-01-22
[TASK] Introduce main configuration object

Mabahe:
This patch streamlines the typoscript configuration handling.
At various places the typoscript configuration path plugin.tx_solr. is
used, also in context of a content object. So typoscript settings that
should affect a single content element path like plugin.tx_solr_PiResults_Results.
was ignored. Also the flexform settings are merged in the main
configuration object now.

For now it does not support nested solr plugins.

Resolves: #163


Mabahe:
Checked this commit in my project; worked fine and fixes a bug from the first version, that I meanwhile found in suggest eid. So lets run travis :-)

PR #211, opened 2016-01-27
Raised solrconfig version as followup of ticket #204

timoschmidt:
Followup of ticket #204

### DIFF
--- a/Classes/Report/SolrConfigStatus.php
+++ b/Classes/Report/SolrConfigStatus.php
@@ -50,7 +50,7 @@ class SolrConfigStatus implements StatusProviderInterface
      *
      * @var string
      */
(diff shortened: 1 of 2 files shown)


irnnr:
Since this is master/3.2.0 also increase the version in the name

irnnr:
Go ahead and merge after tweaking the version

timoschmidt:
@irnnr 
I've changed the version to 4.0.0 since this was the next version in the last discussion. What about other schema changes? Do we change the version, when we release a new EXT:solr version of only when a schema change is really done?

irnnr:
Until now I only changed the name attribute whenever something actually changed. Looks good to me.

PR #213, opened 2016-01-28
[TASK] Introduce main configuration object

timoschmidt:
This patch streamlines the typoscript configuration handling.
At various places the typoscript configuration path plugin.tx_solr. is
used, also in context of a content object. So typoscript settings that
should affect a single content element path like plugin.tx_solr_PiResults_Results.
was ignored. Also the flexform settings are merged in the main
configuration object now.

The TypoScriptConfiguration object support array access for backwards compatibility.
In the long run, we should use speaking methods or "getObjectByPath" or "getValueByPath"

For now it does not support nested solr plugins.

Resolves: #163

This pull request is based on @Mabahe pull request with the addition that, the configuration object can be used with array access operators for backwards compatibility and adds some unit and integrations tests.

### DIFF
--- a/Classes/AdditionalFieldsIndexer.php
+++ b/Classes/AdditionalFieldsIndexer.php
@@ -41,6 +41,16 @@ class AdditionalFieldsIndexer implements SubstitutePageIndexer
 {
 
     /**
+     * @var array
+     */
+    protected $configuration;
+
+    public function __construct()
+    {
+        $this->configuration = Util::getSolrConfiguration();
+    }
+
+    /**
      * Returns a substitute document for the currently being indexed page.
      *
      * Uses the original document and adds fields as defined in
@@ -89,7 +99,7 @@ class AdditionalFieldsIndexer implements SubstitutePageIndexer
     protected function getAdditionalFieldNames()
     {
         $additionalFieldNames = array();
-        $additionalFields = $GLOBALS['TSFE']->tmpl->setup['plugin.']['tx_solr.']['index.']['additionalFields.'];
+        $additionalFields = $this->configuration['index.']['additionalFields.'];
 
         if (is_array($additionalFields)) {
             foreach ($additionalFields as $fieldName => $fieldValue) {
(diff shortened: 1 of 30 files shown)


dkd-friedrich:
Worked for me, but I would like to discuss an expansion. I think the entire access to the TypoScript configuration should be handled by the new configuration manager. Currently there are still places where the configuration is accessed directly:

1. \ApacheSolrForTypo3\Solr\Util->getConfigurationFromInitializedTSFE()
2. \ApacheSolrForTypo3\Solr\Typo3PageContentExtractor->excludeContentByClass()
3. \ApacheSolrForTypo3\Solr\SuggestQuery->__construct()
4. \ApacheSolrForTypo3\Solr\IndexQueue\AbstractIndexer->resolveFieldValue()
5. \ApacheSolrForTypo3\Solr\Facet\DateRangeFacetRenderer->loadJavaScriptFiles

To use the configuration manager in \ApacheSolrForTypo3\Solr\Util->getConfigurationFromInitializedTSFE() and \ApacheSolrForTypo3\Solr\IndexQueue\AbstractIndexer->resolveFieldValue() requires the possibility to get the full TypoScript setup, but I think this could be added easily.

Additionally a check for initialized TSFE could be added, currently we simply expect that it's already initialized.

timoschmidt:
@dkd-friedrich Thanks for your Feedback. I think you are right. There are some things that we should do in addition:

1. Find places like this an refactor them
2. Instead of accessing the arrays directly implement own speaking getters to read from the configuration.

But i would propose to merge this first and afterwards identify these parts and create smaller separate tasks but with the general goal to avoid the access to the configuration beside the ConfigurationManager/TypoScriptConfiguration object.

dkd-friedrich:
@timoschmidt Ok, let's start with this initial implementation and adapt the remaining points in a second step

timoschmidt:
Thx markus i've removed the var_dump() :)

:+1:

Issue #198, opened 2015-12-30
[PHP7] Incompatible override of Apache_Solr_Service::extract in ApacheSolrForTypo3\Solr\SolrService


irnnr:
`Declaration of ApacheSolrForTypo3\Solr\SolrService::extract(ApacheSolrForTypo3\Solr\ExtractingQuery $query) should be compatible with Apache_Solr_Service::extract($file, $params = Array, $document = NULL, $mimetype = 'applicatio...')`

https://travis-ci.org/TYPO3-Solr/ext-tika/jobs/99522872#L258

Potentially move that method to EXT:tika and have it call the Apache_Solr_Service implementation as intended.

PR #217, opened 2016-02-01
[BUGFIX]  [PHP7] Incompatible override of Apache_Solr_Service::extract in ApacheSolrForTypo3\Solr\SolrService 

timoschmidt:
Renamed method extract to extractByQuery because the signature of the

overwritten method is not compatible. Added integration test for pdf extraction as well.

Fixes: #198

Needs change in EXT:tika


timoschmidt:
After merging we need to use ApacheSolrForTypo3\Solr\SolrService::extractByQuery in EXT:tika

irnnr:
Alternatively I was thinking we could move this into EXT:tika altogether. This way we would not have a breaking change in dependencies on the tika side. OTOH EXT:tika could have checks of course whether the method exists or what EXT:solr version is being used.

What do you think?

timoschmidt:
I would propose to keep it in EXT:solr because the extract functionallity is maybe usefull as api for other extensions to. When we would move it to EXT:tika it would be an API Change nevertheless.

irnnr:
ok, sounds good

Issue #193, opened 2015-12-22
Support hl.useFastVectorHighlighter in Query


timoschmidt:
Since the performance of highlighting could get very poor (e.g. when you are using grouping and highlighting together) and some fields have the required schema settings, we should support this option in the highlighting configuration.

irnnr:
Please don't make this configurable, simply replace the existing highlighter.

dkd-friedrich:
IMHO it could be advantageous to support both types, since the FastVector Highlighter requires a fragment size of 18, doesn't accept 0 and so always restricts the field length. 

Additionally it changes the behavior to switch to the FastVector Highlighter by default, the commonly used setting of "fragmentSize = 0" will result in errors and requires configuration adaptions on extension update.

timoschmidt:
I would agree to markus because the FastVectorHighlighter only works for fields with a specific configuration (termVectors=on, termPositions=on and termOffsets=on) and this would mean highlighting is not possible for other fields.

irnnr:
I think that's fine. Right now that applies only to the content field, it might make sense to also configure the title field this way. 
Either way, I'd like to keep the extension simple for the enduser; convention over configuration. The extension does work one specific way and is supposed to fit >90% of use cases out of the box. If someone needs something more special there's always a way to customize through code by registering different classes, using hooks or using xclasses.

Making highlighting configurable is too low-level for the 90%.

timoschmidt:
So what is your proposal for the default then?

- Set termPositions="true" termOffsets="true" also on title
- Use fastVectorHighlighter when fragSize ist larger then 18 otherwise 
- fallback to normal highlighter when fragsize is smaller then 18 or highlighting field except title or content is passed

I think the advantage would be that fastVectorHighlighter will be used out of the box, but the downsides are, that the code is directly coupled to the schema and the usage is limited to a few fields.
 
I think as it is right now, it makes sence for most of the systems, because it is disabled by default and for most cases the performanceof highlighting is ok. Only when not, you need to enable it and in this case you should know what you do. 

In general it makes sence to avoid the complexity in the configuration but for me the usage of fastVetorHighlighter has to many edge cases to use it by default.

irnnr:
Highlighting on `title` never worked so far, so if we could get that fixed that'd be nice.
Solr's `fragsize` default is 100, so that should not be a problem.
[...]

PR #220, opened 2016-02-02
Added configuration "plugin.tx_solr.results.resultsHighlighting.useFastVectorHighlighter" to use a faster highlighter when needed.

timoschmidt:
Can be used to enable "hl.useFastVectorHighlighter" to improve the highlighting performance.

Fixes: #193

### DIFF
--- a/Classes/Query.php
+++ b/Classes/Query.php
@@ -90,12 +90,16 @@ class Query
     private $rawQueryString = false;
 
     /**
-     * Constructor
-     *
+     * @param string $keywords
+     * @param null $solrConfiguration
      */
-    public function __construct($keywords)
+    public function __construct($keywords, $solrConfiguration = null)
     {
-        $this->solrConfiguration = Util::getSolrConfiguration();
+        if ($solrConfiguration == null) {
+            $this->solrConfiguration = Util::getSolrConfiguration();
+        } else {
+            $this->solrConfiguration = $solrConfiguration;
+        }
 
         $this->fieldList = array('*', 'score');
         $this->setKeywords($keywords);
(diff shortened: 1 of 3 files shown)


dkd-friedrich:
I could successfully activate and test the FastVectorHighlighter, but I encountered an error with low fragment size setting

```
java.lang.IllegalArgumentException: fragCharSize(0) is too small. It must be 18 or higher.
	at org.apache.lucene.search.vectorhighlight.BaseFragListBuilder.createFieldFragList(BaseFragListBuilder.java:51)
	at org.apache.lucene.search.vectorhighlight.SimpleFragListBuilder.createFieldFragList(SimpleFragListBuilder.java:39)
	at org.apache.lucene.search.vectorhighlight.FastVectorHighlighter.getFieldFragList(FastVectorHighlighter.java:232)
	at org.apache.lucene.search.vectorhighlight.FastVectorHighlighter.getBestFragments(FastVectorHighlighter.java:185
```

Though I couldn't find a notice about this requirement in the wiki, I think we could add a notice in the TypoScript configuration

timoschmidt:
I've added a documentation in the TypoScript that useFastVectorHighlighter is only usable with a fragementSize larger then 18 and throw an Exception now, when this is not the case.

Does everything beside that look good?

dkd-friedrich:
> Does everything beside that look good?

Yes, with fragment size >= 18 highlighting worked as expected

irnnr:
Please don't make this configurable, simply replace the existing highlighter.

timoschmidt:
Hmm, but it would not work with all fields in the schema because it requires termVectors=on, termPositions=on and termOffsets=on so highlighting would be impossible for the other fields. Or do you think it is only required for content?

PR #223, opened 2016-02-03
[BUGFIX] Charset not respeceted in Typo3PageContentExtractor::excludeContentByClass

### LINKED ISSUE #218: Charset not respeceted in Typo3PageContentExtractor::excludeContentByClass
The charset is not respected and therefore the php default value is used.

timoschmidt:
Make sure that parsed content is supposed to be UTF-8.

Fixes #218


dkd-friedrich:
LGTM, my tests showed no charset errors with this adaptions.

We should keep in mind, that we're currently only supporting utf-8.

Issue #127, opened 2015-11-13
Use bootstrap modals in backend module danger zones


irnnr:
Check actions in backend module danger zones and use `TYPO3.Modal.confirm()` where it makes sense.

PR #226, opened 2016-02-03
[TASK] Use bootstrap modals for danger zone items and remove usage of ActionButtonViewHelper

timoschmidt:
Uses bootrap modals from TYPO3 core for danger zone buttons and remove the ActionButtonViewHelper since this is possible with native fluid view helpers.

Fixes: #127


irnnr:
Hmm, this is not the same unfortunately. AFAIK the link view helper renders an anchor tag, but EXT:solr's action button renders a form with a button that submits the form. This was on purpose as the form submits via POST, the link is a GET. These are destructive actions and should not use GET as that allows to reload, while a POST causes the browser to prompt when reloading. In theory it should even be a DELETE request although granted the whole thing is not fully REST compliant anyway.

timoschmidt:
I see the point, but in the TYPO3 backend nearly all of these "destructive actions" are implemented with links and on the other hand the server side implementation should avoid this by doing a redirect instead of relying on the browsers behaviour. Nevertheless it can be implemented by using the form viewhelper of fluid. I'll check that and check how modals can be used with these forms.

timoschmidt:
I've changed the implementation to use forms with a post request and implemented an own modal handler to use this in all backend dialogues.

irnnr:
cool, thanks!

Issue #48, opened 2015-08-07
Indexing of new records will crash if the name of the  Indexing Queue Configuration is different from tablename


svenleemaier:
When you have in your SOLR configuration a index queue item for a table:
```
plugin.tx_solr.index.queue {
    news {
        table = tt_news
        fields {
            ...
        }
    }
}
```
the indexing of a new record for this table will insert the wrong indexing configuration name (indexing_configuration) in the tx_solr_indexqueue_item table. It will insert "tt_news" instead of "news" and with this indexing_configuration the Indexer gets NULL instead of an array of the itemIndexingConfiguration because there is no configuration ```plugin.tx_solr.index.queue.tt_news.fields```.

The problem is in the Hook "processDatamap_afterDatabaseOperations" of ext-solr/Classes/IndexQueue/RecordMonitor.php line 232:
```PHP
if ($this->isEnabledRecord($recordTable, $record)) {
	$configurationName = NULL;
	if ($recordTable !== 'pages') {
/* #232 */		$configurationName = $this->getIndexingConfigurationName($table, $uid);
	}
	$this->indexQueue->updateItem($recordTable, $recordUid, $configurationName);
}
```
To get the correct indexing_configuration (```$configurationName```) you have to use ```$recordTable``` and ```$recordUid```:
```PHP
if ($this->isEnabledRecord($recordTable, $record)) {
	$configurationName = NULL;
	if ($recordTable !== 'pages') {
/* #232 Changed: */		$configurationName = $this->getIndexingConfigurationName($recordTable, $recordUid);
	}
	$this->indexQueue->updateItem($recordTable, $recordUid, $configurationName);
}
```

Do I see that right?

dkd-dobberkau:
If I read the manual right here: https://forge.typo3.org/projects/extension-solr/wiki/Tx_solrindex

"plugin.tx_solr.index.queue
Defines a set of table indexing configurations. By convention the name of the indexing configuration also represents the table name. You can name the indexing configuration differently though by explicitly defining the table as a parameter within the indexing configuration. That's useful when indexing records from one table with different configuration - different single view pages / URLs for example."

So is this really a Bug?

dkd-dobberkau:
Additionally there is a difference between tt_news and ext:news or?

dkd-dobberkau:
And have you seen that we have static templates for those two news extensions?

[...]

PR #229, opened 2016-02-05
Backport fix for #48 to 3.0.x

timoschmidt:
Backports the fix from #48 (Indexing of records crash when indexing configuraiton and table name is different when new record created) to the 3.0.x branch.


timoschmidt:
Reviewed together with nadja

:+1:

Issue #228, opened 2016-02-04
Avoid array usage of configuration and implement own getters for configuration options where it is possible


timoschmidt:
As a followup of #213 we should replace the old array usage where it is possible and implement and own getter whereever it makes sence. This make the extension easier to test because the code is more decoupled from the configuration and configuration structure.

timoschmidt:
The Frontendpart was done in https://github.com/TYPO3-Solr/ext-solr/pull/230. We need to check the indexing part angain and check what should be refactored there, but since the new implementation is backwards compatible we can come back to this topic later

PR #230, opened 2016-02-05
Use configuration object for most of the search. and faceting. settings

timoschmidt:
This PR refactors most of the parts where "plugin.tx_solr.(search/logging/viewHelpers/templateFiles)" was used. I think we should refactor this in smaller chucks to keep the change set in a reasonable size.

Also the integration testsuite was enhanced to cover about 40% of the extension code.

Relevant Issue: #228


dkd-friedrich:
I ran some tests and everything worked like expected, so LGTM.

PR #246, opened 2016-02-29
Run dev-master tests only with PHP7 and raised TYPO3 7.6 version to 7.6.4

timoschmidt:
Run the Test against TYPO3 dev-master with PHP 7.0 only


irnnr:
:+1:

PR #261, opened 2016-03-09
[BUGFIX] Delete all page specific documents if a page is removed

sascha-egerer:
If a TYPO3 page is hidden or deleted it will be removed from the
index. But all records stored on that page must also be removed
from the index so that they are not searchable anymore.


Classes/GarbageCollector.php
```diff
@@ -215,6 +219,33 @@ protected function deleteIndexDocuments($table, $uid)
     }
 
     /**
+     * Delete all index documents that are related to a TYPO3 "page"
+     *
+     * @param $uid
```
irnnr: please add type and description for the parameter

irnnr:
This is only part of the fix. You will also have to change RecordMonitor to remove the records from the Index Queue and also re-add them to the Index Queue when the page gets enabled again, taking into account what Indexing Configurations exist.

Classes/GarbageCollector.php
```diff
@@ -215,6 +219,33 @@ protected function deleteIndexDocuments($table, $uid)
...
+     */
+    protected function deleteRecordsOfPage($uid)
```
irnnr: although the method name mentions records, you're also deleting Solr documents. That should be separated or named differently. 

Maybe you can also move this into `collectPageGarbage()`

There are methods here already that do what this method attempts to do; have a look at `deleteIndexDocuments()` and

Classes/GarbageCollector.php
```diff
@@ -215,6 +219,33 @@ protected function deleteIndexDocuments($table, $uid)
...
+        $connectionManager = GeneralUtility::makeInstance('ApacheSolrForTypo3\\Solr\\ConnectionManager');
+        $indexQueue = GeneralUtility::makeInstance('ApacheSolrForTypo3\\Solr\\IndexQueue\\Queue');
+        $site = Site::getSiteByPageId($uid);
+
+        // a site can have multiple connections (cores / languages)
+        $solrConnections = $connectionManager->getConnectionsBySite($site);
+
+        /** @var  $solr */
+        foreach ($solrConnections as $solr) {
+            $response = $solr->search('pid:' . intval($uid), 0, 10000);
```
irnnr: You must/should not use Solr to find the documents that need to be deleted. The source of truth is the Index Queue and/or the TYPO3 DB in general. So you'll have to find the items there and then delete from Solr using that information.

You then just call `deleteIndexDocuments()` and don't iterate over the Solr connections here.

irnnr:
Looking at this issue it is far more complex than the fix in this PR.

This comes down to the long standing issue of the core not bubbling/propagating page change events through the tree. Basically when a page gets changed that event should be carried down the tree using TCA main.

Right now you're only dealing with records directly contained in the changed page, but you'd have to recurse down potential sub-pages, too.

Classes/GarbageCollector.php
sascha-egerer: `deleteIndexDocuments` is called below as it does not what i need.

Classes/GarbageCollector.php
sascha-egerer: Yeah the problem is that the page that holds the entries must not be in the index queue (page of type folder) and as already mentioned there is no easy way to find out which records belong to a specific page.

So i used this approuch to clean the index form records that should not be there anymore.

I don't see good way to get all related recoreds for now... What would you suggest?

sascha-egerer:
> This is only part of the fix. You will also have to change RecordMonitor to remove the records from the Index Queue and also re-add them to the Index Queue when the page gets enabled again, taking into account what Indexing Configurations exist for the site containing the page.

I know that and already talked about that in slack. I would keep this in 2 seperate patches.

> Looking at this issue it is far more complex than the fix in this PR.
>
> This comes down to the long standing issue of the core not bubbling/propagating page change events through the tree. Basically when a page gets changed that event should be carried down the tree using TCA main.
> 
> Right now you're only dealing with records directly contained in the changed page, but you'd have to recurse down potential sub-pages, too.

That should be already done as this command is some kind of recursive as `deleteIndexDocuments` calls `deleteRecordsOfPage` if the type is `pages` and `deleteRecordsOfPage` calls `deleteIndexDocuments` to delete the document finally.

I think this would be a quick win to solve the problem for now. I would like to have the `correct` solution but i don't think we can get it in the near future... So i would vote for a quick win...

Classes/GarbageCollector.php
irnnr: Ah true, didn't think about folder pages. However, if you have a page you should be able to get its site configuration and the indexing configuration to find what types to look for. I was first thinking about simply looking at the Index Queue, but that doesn't have a page ID. Adding it wouldn't make things easier either since then you'd have to update it whenever a record is moved.

Thinking about it further it might actually be possible; since we get notified of any changes to a record we could update the Index Queue's entry's pageID whenever we update the entry. 

Then you'd have an easy way to find records on a page and the documents that belong to it. 

Also, how about using the reference index, could that work?

timoschmidt:
@irnnr 
@sascha-egerer 

What about implementing the stuff from the first patch and add an integration or unit tests at the same time to make sure, the expected behaviour is documented?

@sascha-egerer 

Thx for your efford by now? Do you plan to implement the second part also?

Issue #283, opened 2016-03-22
[TASK] remove q.op from solrconfig.xml since it does not make sence to use together with default mm value


timoschmidt:
In edismax it does not make sence to use q.op = OR together with an default mm value. Therefore we should remove it.

PR #286, opened 2016-03-30
[TASK]  Remove q.op

dkd-friedrich:
Since the mm parameter is set as default value for the search request handler q.op has no effect and should be removed

Fixes: #283


timoschmidt:
Looks good. Thx!

Issue #284, opened 2016-03-22
[TASK] Add a dynamic field type with same configuration as spellField


timoschmidt:
When you want use an own spell field you need to have the same data type present. Therefore it makes sence to introduce an own dynamic field with the same configuration as the spell field

* Add own dynamic type
* Update schema version
* Update and check reports that they are green

PR #287, opened 2016-03-30
[TASK] Add a dynamic field type with same configuration as spellField

dkd-friedrich:
Added dynamic fields of type textSpell

Fixes: #284


timoschmidt:
Looks good Thx!

PR #291, opened 2016-03-30
Added datetime field to index and bodytext to be parsed for file links

seirerman:


timoschmidt:
Looks reasonable, thx! :+1:

PR #308, opened 2016-04-04
[BUGFIX] Re-Index after deleting a content element

### LINKED ISSUE #231: Re-Index after deleting a content element
Hi, I thinks there may be a small bug in reindexing after deleting a content element.

By deleting a content element the page will directly removed from solr index.
This seems to be fine :), but there will be no update in the index-queue, so the page will not be re-indexed.

By hiding a content-element everything works like expected:
- direct remove from index
- updating the index-queue
- re-index page via scheduler

I'm using TYPO3 v.7.6.2 with flux, fluidpages and fluidcontent.
Not sure if this happens outside a "flux-environemt". It would be great if somebody else check this.

Kind regards
Heiko

markuskobligk:
fixes #231

### DIFF
--- a/Classes/IndexQueue/Queue.php
+++ b/Classes/IndexQueue/Queue.php
@@ -314,18 +314,23 @@ class Queue
      *      different value for non-database-record types.
      * @param string $indexingConfiguration The item's indexing configuration to use.
      *      Optional, overwrites existing / determined configuration.
+     * @param boolean $reindexPageAfterContentDeletion TRUE if the item needs to be updated due to deletion of content elements, FALSE otherwise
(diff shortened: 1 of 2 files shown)


timoschmidt:
Hi Markus, thx for your pull request. What do you think of changing the resposibilities? To instead $reindexPageAfterContentDeletion have something like $forcedChangeTime then it is not page related only and when now forced change time was passed, the value from $this->getItemChangedTime($itemType, $itemUid) could be used.

markuskobligk:
Hi Timo, yes this is a better solution, I have updated the branch... please check if this is what you had in mind.

timoschmidt:
Thx for you help! I've mered your patch into dev-master

PR #314, opened 2016-04-04
[TASK] Improve performance of record indexer

dmitryd:
Cache for language information is not kept across instances of the indexer. Each new record
requires a new instance of the indexer but language information does not change between creation of
indexer instances. Language information is obtained by instantiating TSFE and parsing TS templates.
This can take 0.5-2 seconds depending on the amount of TypoScript. It makes sense to cache this
information statically. It helps to shorten indexing times dramatically and lower CPU usage on
the server.

### DIFF
--- a/Classes/IndexQueue/Indexer.php
+++ b/Classes/IndexQueue/Indexer.php
@@ -82,8 +82,7 @@ class Indexer extends AbstractIndexer
      *
      * @var array
      */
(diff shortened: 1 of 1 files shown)


dmitryd:
Helps with about 20,000 news and three languages ;)

timoschmidt:
Looks good. Thx for your contribution!

PR #319, opened 2016-04-05
[BUGFIX] Implement proper content fallback handling for records

dmitryd:
Content fallback is a TYPO3 feature that allows to show content in languages other than current and default if the content in the current language is not available.

Reference for the content fallback feature is available [here](https://docs.typo3.org/typo3cms/TyposcriptReference/Setup/Config/Index.html#sys-language-mode).

Here is how content fallback works in TYPO3. Suppose there a website in Switzerland that has the following language set up:

<table>
  <tr>
    <th>sys_language_uid</th>
    <th>value</th>
  </tr>
  <tr>
    <td>0</td>
    <td>de_CH</td>
  </tr>
  <tr>
    <td>1</td>
    <td>fr_CH</td>
  </tr>
  <tr>
    <td>2</td>
    <td>it_CH</td>
  </tr>
  <tr>
    <td>3</td>
    <td>fr_FR</td>
  </tr>
  <tr>
    <td>4</td>
    <td>it_IT</td>
  </tr>
  <tr>
    <td>5</td>
    <td>en_US</td>
  </tr>
</table>

News for swiss french and french french will be identical except for some cases. Same for two italian versions. Threfore editors will only create `fr_CH` and `it_CH` versions of news items.

Now if they try to switch language to french french or italain italian, they will not see news in french or italian but in german. TYPO3 only checks for exact language match and if it does not exist, it uses default language record.

To avoid that, the integrator does the following config in typoscript:

```
[globalVar = GP:L = 3]
# French in France
config.sys_language_mode = content_fallback; 1, 5, 0
[globalVar = GP:L = 4]
# Italian in Italy
config.sys_language_mode = content_fallback; 2, 5, 0
[config]
```

These options tell TYPO3 the following:
- In case if we are on "french french", try to fetch `fr_FR` first (requested language). If it does not exist, try `fr_CH` (uid=1), it if does not exist, try `en_US` (uid=5; may be, people in France would prefer English to German?) and if that does not exist, go for German (uid=0).
- In case if we are on "italian italian", try to fetch `it_IT` first (requested language). If it does not exist, try `it_CH` (uid=1), it if does not exist, try `en_US` (uid=5) and if that does not exist, go for German (uid=0).

Now this works fine with news extension.

The editor is happy but suddenly users compain that they cannot find news items for french french and italian italian in search. The editors tries himself and he cannot find them either. If he just uses search facet, he finds out that his `content_fallback` settings do not work in solr and solr just shows a german version of news...

This pull request solves the issue in solr.

### DIFF
--- a/Classes/IndexQueue/Indexer.php
+++ b/Classes/IndexQueue/Indexer.php
@@ -85,6 +85,13 @@ class Indexer extends AbstractIndexer
     protected static $sysLanguageOverlay = array();
 
     /**
+     * Cache of the sys_language_content information
+     *
+     * @var array
+    */
+    protected static $sysLanguageContent = array();
+
+    /**
      * Constructor
      *
      * @param array Array of indexer options
@@ -193,6 +200,7 @@ class Indexer extends AbstractIndexer
         $overlayIdentifier = $rootPageUid . '|' . $language;
         if (!isset(self::$sysLanguageOverlay[$overlayIdentifier])) {
             Util::initializeTsfe($rootPageUid, $language);
+            self::$sysLanguageContent[$overlayIdentifier] = $GLOBALS['TSFE']->sys_language_content;
             self::$sysLanguageOverlay[$overlayIdentifier] = $GLOBALS['TSFE']->sys_language_contentOL;
         }
 
@@ -202,12 +210,23 @@ class Indexer extends AbstractIndexer
             $page = GeneralUtility::makeInstance('TYPO3\\CMS\\Frontend\\Page\\PageRepository');
             $page->init(false);
 
-            $itemRecord = $page->getRecordOverlay(
+            $localizedItemRecord = $page->getRecordOverlay(
                 $item->getType(),
                 $itemRecord,
                 $language,
                 self::$sysLanguageOverlay[$overlayIdentifier]
             );
+            if (!isset($localizedItemRecord['_LOCALIZED_UID'])) {
+                $localizedItemRecord = $page->getRecordOverlay(
+                    $item->getType(),
+                    $itemRecord,
+                    self::$sysLanguageContent[$overlayIdentifier],
+                    self::$sysLanguageOverlay[$overlayIdentifier]
+                );
+            }
+            if ($localizedItemRecord) {
+                $itemRecord = $localizedItemRecord;
+            }
         }
 
         if (!$itemRecord) {


dkd-dobberkau:
Thanks for the description of this use case.

dmitryd:
@dkd-dobberkau Sure :wink:

timoschmidt:
@dmitryd 

Hi dymitry, first of all thanks for the detailed description and the provided pull request. I think in general it makes sence to have this feature. And in the use case you desribe it totally makes sence. But it could happen that the fallback is configured to show a different language what could e.g. add resulsts from wrong languages to the spellchecking, suggest etc.

I would propose that we add a configuration to the indexing configuration that allows us to enable the fallback resolution for the record type.

What do you think? Can you add this to you pull request?

dmitryd:
> it could happen that the fallback is configured to show a different language what could e.g. add resulsts from wrong languages to the spellchecking, suggest etc.

It happens without fallback too: if there no news in French, German news will be displayed in search results.

I think the option will only complicate the whole thing. Since fallback is official TYPO3 feature and it works with content, users will expect to see search results in the same language as they are displayed on the page. Right now we can show fr_CH when fr_FR is missing on the page but solr will put de_CH into the french solr core and show German, which is really unexpected.

Right now solr **already** puts wrong data to suggest and spelling when translation does not exist. Content fallback will let integrators fix it by providing fallback to a proper language.

timoschmidt:
I think regarding a fallback in the same language it makes sence (e.g. fr_CH, fr_FR) but for configured fallbacks to other languages not. It also changes the current behaviour and therefore i would say others need to have the possiblity to keep the same behaviour.

dmitryd:
> I think regarding a fallback in the same language it makes sence (e.g. fr_CH, fr_FR) but for configured fallbacks to other languages not.


Shouldn't this be decided by integrators if it is appropriate for their goals or not? 😄 Main problem is that solr now behaves differently from TYPO3. Imagine that you do not index records but index news detail page via http. In such case you would have different search results because TYPO3 core works differently with languages than solr. The goal of this PR is to make solr behavior the same as for the TYPO3 core.

If you think the option is necessary, it can be added, of course, but it is unusual to have an option to support a TYPO3 feature in solr that works in the core since TYPO3 4.x.

To me this missing support looks like a bug or a missing feature that it natural.

timoschmidt:
@dmitryd 

> Shouldn't this be decided by integrators if it is appropriate for their goals or not? :smile: Main problem is that solr now behaves differently from TYPO3. Imagine that you do not index records but index news detail page via http. In such case you would have different search results because TYPO3 core works differently with languages than solr. The goal of this PR is to make solr behavior the same as for the TYPO3 core.
> 
> If you think the option is necessary, it can be added, of course, but it is unusual to have an option to support a TYPO3 feature in solr that works in the core since TYPO3 4.x.

I agree with you, thats true, but i changes the behaviour and it could be suprising for somebody how updates and he has no chance to have the old behaviour. Therefore i would propose to have the possiblity to turn it on or off.

dkd-dobberkau:
@dmitryd I am not sure if the extension should award people for doing things not ment to be done. there is one way to index news and indexing news via the page indexer is plain wrong and makes no sense. I would even doubt that there is "a single way and correct" way people user the language concepts of TYPO3. As Language Handling will change one day we shouldnt invest to much time in edgecasing.

dmitryd:
@dkd-dobberkau 

> I am not sure if the extension should award people for doing things not ment to be done. there is one way to index news and indexing news via the page indexer is plain wrong and makes no sense.

I do not propose to do that. I only made this example to show that EXT:solr does not support an official TYPO3 feature that TYPO3 supports for >5 years.

>  I would even doubt that there is "a single way and correct" way people user the language concepts of TYPO3. 

Well, this is not about correct or not, this is about synchronizing solr behavior with TYPO3 core, that's all.

> As Language Handling will change one day we shouldnt invest to much time in edgecasing.

This is not edge casing, this is a missing functionality that breaks multilanguage sites. For now we have to fix it locally. I just thought it could be useful to make the fix public.

I can also release the fix as a separate extension that Xclasses solr.

dkd-dobberkau:
@dmitryd there is no need for that. i trust you all to find a way that will work. maybe i am just to simplistic here.

dkd-friedrich:
@dmitryd Thanks for your detailed description and your pull request. I agree, EXT:solr should behave like TYPO3 and as described, but we still see some problems.

EXT:solr is currently initializing TSFE with the id of root page, therefore the translations of the root page or if the root page is a shortcut, the shortcuts target page is the crucial factor. This e.g. requires that the root page / shortcut target is not translated to "french french" (fr_FR, ID 3), since TSFE will determine fr_FR (ID 3) as sys_language_content and not fr_CH (ID 1) or en_US (ID 5), while trying to index news for "french french" (fr_FR, ID 3).

Since the news detail pages must be configured with the required translations to achieve the preferred fallbacks, in our opinion the news detail pages could be used to index the news records and so consider the right fallbacks. An additional setting that provides the page to use for record indexing should solve this problem.

Issue #307, opened 2016-04-04
Handing of quotes in the querystring


timoschmidt:
Currently it is only possible to use quotes in the beginning of the query string

e.g.

"foo"

It should be possible to use quotes also for parts

"hello world" or planet

At the same time it is important to still escape parts of the query that are risky in terms of security

PR #320, opened 2016-04-05
[TASK] Support phrases inside a query string

timoschmidt:
Before only the usage of "hello world" as a phrase was supported. The usage of more complex phrase queries e.g. `"hello world" or "hello planet"`was not supported.

This pull request:

1. adds test cases for the escape method of the query for the previous use cases
2. implements the context depending (inside or outside a phrase) escaping for multiple phrases

Fixes: #307

### DIFF
--- a/Classes/Query.php
+++ b/Classes/Query.php
@@ -264,26 +264,63 @@ class Query
      */
     public function escape($string)
     {
-        if (!is_numeric($string)) {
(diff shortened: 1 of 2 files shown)


dkd-friedrich:
Worked for me, these changes allowed me to use several phrases and combine them with other operators and search strings

Issue #236, opened 2016-02-17
[BUGFIX] Search with whitspaces only should be treated as empty search


timoschmidt:
When you search for spaces only " " this should be threated as an empty search.

PR #335, opened 2016-04-11
[BUGFIX] Search with whitespace only should be threated as empty search

lorenzulrich:
Searches with whitespaces only are handled now as a search with an empty
query string "" unless an empty search query is explicitly allowed.

Fixes: #236


timoschmidt:
Looks good. Thx for backporting!

Issue #342, opened 2016-04-14
Backend module does not handle empty state


irnnr:
When starting out with a new TYPO3 installation, after installing EXT:solr entering the backend module leads to an exception.

Environment:

* TYPO3 7.6.5
* Introduction Package
* set up domain record
* installed EXT:solr 4.0.0-dev/master

Then entered backend module and got exception. 

I believe we try to store a selected site to the module data, but there is non yet (null) thus leading to the exception. Instead we should show an empty backend module like it used to (I believe this is a regression). It should also show some initial message about what to do.

[mysql.sql.zip](https://github.com/TYPO3-Solr/ext-solr/files/218582/mysql.sql.zip)
[TYPO3 Exception.html.zip](https://github.com/TYPO3-Solr/ext-solr/files/218586/TYPO3.Exception.html.zip)

The extension TS record in the dump was created after experiencing the exception but is not hooked up yet. TYPO3 admin account: admin/password.

PR #343, opened 2016-04-14
[BUGFIX] Backend module does not handle empty state

timoschmidt:
Added action, when no site is available.

Fixes: #342

### DIFF
--- a/Classes/Controller/AdministrationController.php
+++ b/Classes/Controller/AdministrationController.php
@@ -121,15 +121,15 @@ class AdministrationController extends ActionController
      */
     protected function initializeAction()
     {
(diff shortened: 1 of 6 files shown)


dkd-friedrich:
LGTM

Now I get an info message "There is no site configured for your solr extension.", instead of the unwanted exception:

`PHP Catchable Fatal Error: Argument 1 passed to ApacheSolrForTypo3\Solr\Domain\Model\ModuleData::setSite() must be an instance of ApacheSolrForTypo3\Solr\Site, null given`

irnnr:
👍🏻 Thanks!

PR #347, opened 2016-04-15
[TASK] Use FastVectorHighlighter by default without configuration

timoschmidt:
The FastVectorHighlighter will now be used by default when a fragmentSize larger then 17 is configured.

Fixes: #193

### DIFF
--- a/Classes/Query.php
+++ b/Classes/Query.php
@@ -1193,16 +1193,13 @@ class Query
                 $this->queryParameters['hl.fl'] = $highlightingFields;
             }
 
-            $useFastVectorHighlighter = $this->solrConfiguration->getSearchResultsHighlightingUseFastVectorHighlighter();
-            if ($useFastVectorHighlighter) {
-                if ($fragmentSize <= 18) {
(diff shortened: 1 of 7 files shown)


Classes/Query.php
```diff
@@ -1193,16 +1193,13 @@ public function setHighlighting($highlighting = true, $fragmentSize = 200)
...
-            if ($useFastVectorHighlighter) {
-                if ($fragmentSize <= 18) {
-                    throw new \InvalidArgumentException("The setting useFastVectorHighlighter can only be used with a fragementSize larger then 18");
-                }
-                $this->queryParameters['hl.useFastVectorHighlighter'] = 'true';
-            }
-
+            // the fast vector highlighter can only be used, when the fragmentSize is
+            // higher then 17 otherwise solr throws an exception
+            $useFastVectorHighlighter = $fragmentSize >= 18;
```
irnnr: Can you wrap this in braces please? `$useFastVectorHighlighter = ($fragmentSize >= 18);` Makes it more clear we're evaluating to a boolean.

irnnr:
LGTM

Issue #356, opened 2016-04-19
[BUGFIX] Frontend should not show PHP error in Frontend when connection is not initialized


timoschmidt:
Frontend should not show PHP error in Frontend when connection is not initialized

PR #357, opened 2016-04-19
[BUGFIX] Frontend should not show PHP error in Frontend when connecion is not initialized

timoschmidt:
Frontend should not show PHP error in Frontend when connection is not initialized.

Fixes: #356


dkd-friedrich:
:+1:

Issue #363, opened 2016-04-21
Exception when I click "Queue Selected Content For Indexing" 


geraldspreer:
Typo3 7.6.5
ext-solr 4.0.0

Hi there,

I get this Exception when I click "Queue Selected Content For Indexing" in the Search backend module:

```
Uncaught TYPO3 Exception
#1: PHP Runtime Deprecation Notice: Non-static method TYPO3\CMS\Core\Messaging\FlashMessageQueue::addMessage() should not be called statically, assuming $this from incompatible context in /docker/Source/web/typo3conf/ext/solr/Classes/IndexQueue/Initializer/Page.php line 215 (More information)
```

I will also submit a pull request that fixes this issue. 

Thanks,

Gerald

timoschmidt:
We should also backport this to the 4.0.x branch.

timoschmidt:
@geraldspreer Thx for reporting. I think this error does not occure in a TYPO3 standard installation, but anyhow we should fix it, since it is a deperaction. Can you tell me how, did you configure your 

```php
$TYPO3_CONF_VARS['SYS']['errorHandlerErrors'] & $TYPO3_CONF_VARS['SYS']['exceptionalErrors'] 
```

and error_reporting settings from the php.ini? Because in my case TYPO3 was not running at all when setting it to E_ALL and it would be nice to add this to our travis build.

PR #365, opened 2016-04-21
Bugfix Issue #363 Fixing FlashMessageQueue::addMessage

geraldspreer:
Hi there,
after I installed 4.0.0 of ext-solr, I ran into an exception message while I was trying to "Queue Selected 
Items for Indexing" in the Search backend module. 

https://github.com/TYPO3-Solr/ext-solr/issues/363

It would be great if you would accept this change. If not like this, then maybe in a similar fashion.

Thanks,

Gerald

### DIFF
--- a/Classes/IndexQueue/Initializer/Page.php
+++ b/Classes/IndexQueue/Initializer/Page.php
@@ -51,6 +51,7 @@ class Page extends AbstractInitializer
      */
     public function __construct()
     {
+        $this->flashMessageQueue = new FlashMessageQueue("solr_queue");
         $this->type = 'pages';
         $this->indexingConfigurationName = 'pages';
     }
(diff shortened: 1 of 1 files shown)


timoschmidt:
@geraldspreer Thx for reporting. I think this error does not occure in a TYPO3 standard installation, but anyhow we should fix it, since it is a deperaction. Can you tell me how, did you configure your 

```php
$TYPO3_CONF_VARS['SYS']['errorHandlerErrors'] & $TYPO3_CONF_VARS['SYS']['exceptionalErrors'] 
```

and error_reporting settings from the php.ini? Because in my case TYPO3 was not running at all when setting it to E_ALL and it would be nice to add this to our travis build.

dkd-friedrich:
@timoschmidt I could reproduce this by setting

```
[SYS][errorHandlerErrors] = 32759 
[SYS][exceptionalErrors] = 28672
```

After creating an invalid mount point, e.g. without source page, you'll get the decribed exception when initializing the page index queue.

Issue #371, opened 2016-04-25
PHP fatal in index queue if item conf is empty


Mabahe:
Already had this in different projects because of different reasons.

The Indexer want to update an item, but the indexing configuration is empty. This causes a (silent) fatal in AbstractIndexer::addDocumentFieldsFromTyposcript because $indexingConfiguration have to be an array.

Throw an exception in Indexer::getItemTypeConfiguration instead to trigger a failed indexing.

timoschmidt:
Fixed with #377 please reopen when needed.

PR #372, opened 2016-04-25
[BUGFIX] Avoid fatal when indexing configuration is empty

Mabahe:
Throw an exception instead to trigger a failed indexing.

 Resolves: #371

### DIFF
--- a/Classes/IndexQueue/Indexer.php
+++ b/Classes/IndexQueue/Indexer.php
@@ -269,6 +269,10 @@ class Indexer extends AbstractIndexer
         $solrConfiguration = Util::getSolrConfigurationFromPageId($item->getRootPageUid(),
             true, $language);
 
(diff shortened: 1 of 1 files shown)


timoschmidt:
I think this is a leftover from the configuration object changes and could be done easier.
I've allready replaced this part in a branch where the deprecated parts should get removed.

https://github.com/timoschmidt/ext-solr/commit/dec6f9467de042ec3a3d4b6b71211391286a8c0a

timoschmidt:
Fixed with #377 please reopen when still needed

Mabahe:
Fix in #377 looks fine, THX. But this is sadly no leftover. This problem exists since 2.0 :-)

Will reopen it for PR for 3.1 and 4.0.

timoschmidt:
No i mean, using the method instead the array keys is a leftover. I've ported your change in my commit to master. We can backport it to 3.1 and 4.0 as well, or does anything else not work with the change in master for you?

PR #376, opened 2016-04-25
[BUGFIX] Fix flash message queue usage and display of flash messages

dkd-friedrich:
If invalid mount points were found while initializing the page index
queue, flash messages are added to a flash message queue, but
addMessage() mustn't be called statically and backend module has to
consider the queue initializer flash massages.

Fixes: #363


timoschmidt:
Thx for the PR and adding a test case.

Issue #426, opened 2016-05-30
[FEATURE] Add additionalWhereClause to SOLR_RELATION


timoschmidt:
Allow to configure additionalWhereClause in solr relation

This allows to configure different fields for same relation with additional where clauses.
E.g. to select different categories by there parent.
Example configuration:
```
                        news_stringM = SOLR_RELATION
                        news_stringM {
                            localField = categories
                            foreignLabelField = title
                            multiValue = 1
                            additionalWhere = sys_category.parent = 1
                        }

                        regions_stringM < .news_stringM
                        regions_stringM {
                            additionalWhere = sys_category.parent = 2
                        }

                        product_stringM < .news_stringM
                        product_stringM {
                            additionalWhere = sys_category.parent = 6
                        }
```
Will result in:
![image](https://cloud.githubusercontent.com/assets/354250/15612646/f842b79a-242f-11e6-9a25-bf349a50b959.png)

PR #427, opened 2016-05-30
[TASK] Add additionalWhereClause to SOLR_RELATION

timoschmidt:
* Only check once for duplicate values
* As the code could be replaced with same approach as done in another
  function, use core API instead of custom implementation and reduce
  code
* Also refactor sql query to own method
* This allows to configure different fields for same relation with
  additional where clauses.
  E.g. to select different categories by there parent.

Fixes: #426


DanielSiepmann:
Thanks @timoschmidt, the manual testing still works for me.
Looks fine.

Thanks for fixing the issue and providing the tests.

Issue #433, opened 2016-06-09
[BUG] hl.simple.pre and hl.tag.pre should be set when highlighting is generated


timoschmidt:
When you configure the highlighting for content and title, the configured highlighting wrap is not applied for title.

The reason is, that only hl.tag.pre is passed, but title will be highlighted by the normal highlighter that requires hl.simple.pre.

Todo:

* Add acceptance test
* Fix the bug

PR #434, opened 2016-06-09
[BUGFIX] Highlighting wrap is not working when title and content field is used for highlighting

timoschmidt:
* Fixed bug
* Added integration test

Fixes: #433


irnnr:
nice 👍

PR #456, opened 2016-06-29
[TASK] typoscript setting for getDomain

pniederlag:
Support an alternative to domain records to determine the site (domain) for indexing.

### DIFF
--- a/Classes/Site.php
+++ b/Classes/Site.php
@@ -345,10 +345,16 @@ class Site
      */
     public function getDomain()
     {
-        $pageSelect = GeneralUtility::makeInstance('TYPO3\\CMS\\Frontend\\Page\\PageRepository');
-        $rootLine = $pageSelect->getRootLine($this->rootPage['uid']);
-
-        return BackendUtility::firstDomainRecord($rootLine);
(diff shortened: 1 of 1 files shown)


timoschmidt:
@pniederlag are you sure this will work? 

Array access has been dropped for TypoScriptConfiguration, and "getValueByPathOrDefaultValue" should be used instead. That also maybe the reasone why some integration tests fail:

 ApacheSolrForTypo3\Solr\Tests\Integration\Domain\Search\ResultSet\SearchResultSetServiceTest::canGetDocumentById

PHPUnit_Framework_Exception: PHP Fatal error:  Cannot use object of type ApacheSolrForTypo3\Solr\System\Configuration\TypoScriptConfiguration as array

irnnr:
I can see the need for that. However, this change should only be merged if/when we reconsider handling of site configuration via root level and how we store the site - registry vs. records. I think this is related.

At the moment there is no section `siteConfig`, "config" within a configuration is not needed. If at all, it should just be "site". This PR also misses changes for the documentation.

irnnr:
We should discuss this during the code sprint in August.

Issue #491, opened 2016-07-12
[BUG] absRefPrefix = auto is not working when indexing in cli context (e.g. by news)


timohund:
In EXT:solr 4.0.2 I encounter the dreaded /var/www/html/.... prefix but now in relative URLs for EXT:news records. Configuration of the url field is very basic:
```bash                    
url = TEXT
url {
   typolink.parameter = {$plugin.tx_news.settings.defaultDetailPid}
   typolink.additionalParams = &tx_news_pi1[news]={field:uid}&tx_news_pi1[controller]=News&tx_news_pi1[action]=detail
   typolink.additionalParams.insertData = 1
   typolink.useCacheHash = 1
   typolink.returnLast = url
}
````

timohund:
To reproduce:

* Configure news indexing
* Set absRefPrefix to auto
* Run the indexing with the scheduler in **cli mode**

Result: wrong link is generated

PR #492, opened 2016-07-13
[TASK] absRefPrefix auto not working in cli mode

timohund:
The reason why absRefPrefix is not working in the cli context is, because the webroot is not available in the cli context.

To resolve this. We asume that the webroot is PATH_typo3 (what is most likely the case). If you need to use a different webroot, you can pass it in the indexing scheduler task. When you want to configure a relativ path here, you can use the marker ###PATH_site### and ###PATH_typo3### in the path and they will be replaced with the value of the corresponding constants.

Fixes: #491


Resources/Private/Language/locallang.xlf
```diff
@@ -57,6 +57,9 @@
 			<trans-unit id="indexqueueworker_field_documentsToIndexLimit" xml:space="preserve">
 				<source>Number of documents to index</source>
 			</trans-unit>
+			<trans-unit id="indexqueueworker_field_forcedWebRoot" xml:space="preserve">
+				<source>Forced webroot (only needed when webroot in not PATH_site)</source>
```
irnnr: typo: in -> is

timohund: Will fix it, thx!

PR #609, opened 2016-08-17
[FEATURE] Add hook to manupilate data after they are fetch from database

nlx-sascha:
In some cases you want to index external records for example
products from a ecommerce platform. In this case the records
are not in the TYPO3 database and we need only the id to index
them with our own indexer.


timohund:
@nlx-sascha hi, is it also possible to add a hook there?

nlx-sascha:
I will give it a try

timohund:
@nix-sascha Can you maybe squash the two commits into one?

PR #621, opened 2016-08-18
Add statistics graph

### LINKED ISSUE #601: Add graph for queries over time to statistics

irnnr:
Adds a graph / line chart of queries over time at the top of the new statistics module

Resolves #601


irnnr:
@timohund can you take a look at the failing tests? Not sure what you're testing there. Those functions were never meant to return concatenated strings. That actually made the template crash. I reverted that part already.

timohund:
@irnnr The pevious implementation returned strings and they were also shown in the backend, so this worked (The annotations bte. still indicated that a string will be returned). I'll have a look on in and fix it.

Btw. what should we do when the same query produces 4 hits at the first time and 6 hits at the seconde time (what can happen when the content changes), should we merge them to one entry?

timohund:
Closing, i'll open a new PR soon with your commits.

PR #628, opened 2016-08-22
[TASK] Move the extension folder before upload

timohund:


Build/Release/ter_tag_uploader.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+
+REPRONAME="ext-solr"
```
irnnr: REPO... there's an `R` too much :)

PR #630, opened 2016-08-22
[BUGFIX] Use the configured localField instead of foreignLabelField (#629)

### LINKED ISSUE #629: SOLR_RELATION uses wrong fieldname for $itemList
Classes/ContentObject/Relation.php->getRelatedItemsFromForeignTable is not working as expected because the value $itemList is not set correctly.

At the moment "$parentContentObject->data[$foreignTableLabelField]" (Line 293/294) is used, but at this point we don't want to have the content from the foreign table but the content from the local table, so "$parentContentObject->data[$this->configuration['localField']]" should be the right choice.

Domi202:
Resolves #629

### DIFF
--- a/Classes/ContentObject/Relation.php
+++ b/Classes/ContentObject/Relation.php
@@ -285,8 +285,8 @@ class Relation
             /** @var $relationHandler \TYPO3\CMS\Core\Database\RelationHandler */
         $relationHandler = GeneralUtility::makeInstance('TYPO3\\CMS\\Core\\Database\\RelationHandler');
 
-        $itemList = isset($parentContentObject->data[$foreignTableLabelField]) ?
-                        $parentContentObject->data[$foreignTableLabelField] : '';
+        $itemList = isset($parentContentObject->data[$this->configuration['localField']]) ?
+                        $parentContentObject->data[$this->configuration['localField']] : '';
 
         $relationHandler->start($itemList, $foreignTableName, '', $localRecordUid, $localTableName, $localFieldTca['config']);
         $selectUids = $relationHandler->tableArray[$foreignTableName];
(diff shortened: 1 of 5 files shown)


timohund:
I think we should add an integration test for the failing usecase that is working afterwards.

markuskobligk:
Will this be merged and released with the next version and when is the next version going to be released?

We have a huge problem in a current project with this and I'm going to use a patched fork until this is merged and released.

timohund:
@Domi202 
@markuskobligk 

Can you provide a test case or an example that fails with the current master? I think we should add this to:

1. Avoid in the future that this feature / use case breaks
2. Make sure that other use cases work.

markuskobligk:
What about this existing test?
Tests\Integration\IndexQueue\IndexerTest\canIndexItemWithDirectRelation()

Isn't this the desired test case?

markuskobligk:
In addition to a test, here is an example:
Table 1 is for events with a field location that referecnes the location of the event in a second table.
The second table (same extension) has some fields like title, city, zip, street

INdexing the city field of the second table fails within the indexing configuration for events:
```
city_stringS = SOLR_RELATION
city_stringS {
  localField = location
  foreignLabelField = city
}
```

timohund:
@markuskobligk 

Yes, Tests\Integration\IndexQueue\IndexerTest\canIndexItemWithDirectRelation() is an existing one but it's not failing, so it does not seem to cover this scenario. I think we should add one for this. 

How to run the tests is documented here: 

https://github.com/TYPO3-Solr/ext-solr/blob/master/CONTRIBUTING.md

I can have a look on it but i can not promiss when. Do you have the time to work on this?

markuskobligk:
@timohund 
I will give it a try, but not before middel/end of next week I think

timohund:
@markuskobligk thx

Domi202:
I just updated the integration test fixtures. Seems like the problem with the integration test was that both sides of the relation had the same field name "category".

markuskobligk:
@Domi202
I can confirm that, when both relation sides have the same field name it works (thats exactly why the patch works with localField). Thanks for updating this!

@timohund 
Can you you test/confirm this and merge it into the master?

timohund:
@markuskobligk 
@Domi202 

I've checked the patch and what you decribed and i think the patch looks good. Thanks for pointing this out and preparing a patch!

PR #636, opened 2016-08-24
[TASK] Fix CGL by a code fixer

georgringer:
Using php-cs it is easy to fix CGL issue.
It can be applied by using php-cs-fixer fix --config-file Build/.php_cs


dkd-friedrich:
LGTM

Ingo, what do you think? Any Concerns?

irnnr:
LGTM

Issue #639, opened 2016-08-30
Custom Page types are not support


DanielSiepmann:
TYPO3 offers the possibility to create custom page types beside the existing ones.
That's documented at https://docs.typo3.org/typo3cms/CoreApiReference/PageTypes/Index.html#create-new-page-type .

In our case we created two new page types with own indexing configuration:

```
plugin {
    tx_solr {
        enabled = 1

        index {
            fieldProcessingInstructions {
                publishDate_tDateS = timestampToIsoDate
            }

            queue {
                news < .pages
                news {
                    table = pages
                    initialization = WebVision\WvSite\IndexQueue\Initializer\Page
                    allowedPageTypes = 130
                    additionalWhereClause = doktype = 130 AND no_search = 0

                    fields {
                        pagetype_stringS {
                            value = News
                        }

                        image_intM = SOLR_RELATION
                        image_intM {
                            localField = media
                            foreignLabelField = uid
                            multiValue = 1
                        }

                        publishDate_tDateS = lastUpdated
                    }
                }

                projects < .news
                projects {
                    table = pages
                    allowedPageTypes = 135
                    additionalWhereClause = doktype = 135 AND no_search = 0

                    fields {
                        region_stringM = SOLR_RELATION
                        region_stringM {
                            localField = categories
                            additionalWhereClause = sys_category.parent = 18
                            multiValue = 1
                        }

                        country_stringM < .region_stringM
                        country_stringM {
                            additionalWhereClause = sys_category.parent = 1084
                        }

                        segment_stringM < .country_stringM
                        segment_stringM {
                            additionalWhereClause = sys_category.parent = 20
                        }

                        product_stringM < .segment_stringM
                        product_stringM {
                            additionalWhereClause = sys_category.parent = 1070
                        }

                        application_stringM < .product_stringM
                        application_stringM {
                            additionalWhereClause = sys_category.parent = 1071
                        }

                        business_area_stringM < .application_stringM
                        business_area_stringM {
                            additionalWhereClause = sys_category.parent = 1090
                        }

                        domain_stringM < .business_area_stringM
                        domain_stringM {
                            additionalWhereClause = sys_category.parent = 2
                        }

                        pagetype_stringS {
                            value = Project
                        }
                    }
                }
            }
        }
    }
}
```

At some places, the extension checks for the table that's indexed, which is `pages` and uses hardcoded the configuration for queue `pages` instead of the real configuration `news` or `projects` .

[...]

PR #640, opened 2016-08-30
WIP|FEATURE: Allow different configurations for pages table

DanielSiepmann:
- To keep indexing queue configuration for custom pages types.

Related: #639


DanielSiepmann:
This is not the full Pull Request, just one part of the fix.

This should fix the first issue not being able to index custom page types at all.

DanielSiepmann:
Would be nice if someone can check whether my tests are helpful and cover the necessary parts.
Hopefully I covered everything that's necessary for this fix.

yhermes:
Currently if you change content on an already indexed page and queue again, only the not changed pages get queued again.

timohund:
@DanielSiepmann Thx for your work on this. I'll try to have a look on this in the next days.

timohund:
@DanielSiepmann will have a look on this in the next days and we can proceed on this later.

DanielSiepmann:
We still need to tackle the issue that the page is still queued as `page` for mountpoints:
image
Also the added tests are not good enough, as they didn't show the issue with allowed doktype.

timohund:
Since there is a PR for #715 can you maybe check after merging if the problem still exists?

timohund:
I've rebased the patch and fixed some minor issues

I'll close this PR

@DanielSiepmann 
@yhermes 

Can you verfiy if everthing is working fine when you use (Keep in mind, you need to update the Apache Solr Server):

#778

Issue #647, opened 2016-09-08
Garbage Collector and Record Monitor when frontend editing


megamisan:
When changing record information using frontend editing actions (delete, hide, up, down), the SOLR Garbage Collector and Record Monitor are not triggered since the current ```TYPO3_MODE``` is ```FE```.

The Garbage Collector and Record Monitor hooks should be activated when ```TYPO3_MODE``` is ```BE``` AND when ```TYPO3_MODE``` is ```FE``` and ```$_POST['TSFE_EDIT']``` is set.

dkd-dobberkau:
Can you care for a PR?

irnnr:
Is this happening when modifying records in FE mode through Extbase?

megamisan:
It is happening when using the edit icons in Frontend. Specifically hide and delete buttons. I guess it should also happen when using the move up and move down buttons.
<img width="230" alt="editicons" src="https://cloud.githubusercontent.com/assets/12088094/18361254/713781e4-7601-11e6-9b2f-ff484caec058.png">

Extension feedit must be installed, admin panel activated and frontend editing options activated through admin panel or TSconfig.

The issue is observed on TYPO3 CMS 7 with PHP 5.6.

irnnr:
ok, thanks

PR #648, opened 2016-09-08
[BUGFIX] Have index updated when using frontend editing

megamisan:
DataHandler (TceMain) hooks are registered also in frontend mode when the TSFE_EDIT post parameter array is present.

Fixes: #647

### DIFF
--- a/ext_tables.php
+++ b/ext_tables.php
@@ -132,7 +132,8 @@ if (TYPO3_MODE == 'BE') {
         'solr::clearSolrConnectionCache',
         'ApacheSolrForTypo3\\Solr\\ConnectionManager->updateConnections'
     );
-
+}
+if ((TYPO3_MODE == 'BE') || ((TYPO3_MODE == 'FE') && isset($_POST['TSFE_EDIT']))) {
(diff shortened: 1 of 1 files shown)


ext_tables.php
```diff
@@ -132,7 +132,8 @@
         'solr::clearSolrConnectionCache',
         'ApacheSolrForTypo3\\Solr\\ConnectionManager->updateConnections'
     );
-
+}
+if ((TYPO3_MODE == 'BE') || ((TYPO3_MODE == 'FE') && isset($_POST['TSFE_EDIT']))) {
```
irnnr: Can you make two changes here: 
1) You don't need the braces around `TYPO3_MODE == 'FE'`
2) Please use triple equal signs for comparison `===`

megamisan: Done.

stmllr:
cHash generation has changed with latest TYPO3 release for new installations and now takes id into calculation.

irnnr:
ah right, thanks @stmllr

timohund:
I've fixed the broken tests that are related to the cHash change. Tests are running through now and since everything else looks good, i'll merge this PR.

PR #673, opened 2016-10-10
[TASK] Added integration test for frontend restricted page 

timohund:
* Adds an integration test for frontend restircted pages (fe_users)
* Tweeks some codeparts of the tested code
* Increases the scrutinizer timeout to make sure code coverage is present

### DIFF
--- a/Classes/Typo3PageIndexer.php
+++ b/Classes/Typo3PageIndexer.php
@@ -315,7 +315,10 @@ class Typo3PageIndexer
         $document->setField('rootline', $rootline);
 
         // access
-        $document->setField('access', (string)$this->pageAccessRootline);
+        $access = (string)$this->pageAccessRootline;
+        if (trim($access) !== "") {
+            $document->setField('access', $access);
+        }
         if ($this->page->page['endtime']) {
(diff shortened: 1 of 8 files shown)


Classes/Typo3PageIndexer.php
```diff
@@ -315,7 +315,10 @@ protected function getPageDocument()
         $document->setField('rootline', $rootline);
 
         // access
-        $document->setField('access', (string)$this->pageAccessRootline);
+        $access = (string)$this->pageAccessRootline;
+        if (trim($access) !== "") {
```
irnnr: This should never be empty. The access field always needs a value of at least `c:0` or `r:0`

timohund: I know, but then there there is a bug somewhere else because a PageAccessRootline generated with Rootline::getAccessRootlineByPageId (casted to string in Typo3PageIndexer::getPageDocument) produces an empty access string. When this gets past to solr it will we "", when passing nothing the document will be created with "c:0" which is the default from the schema. Therefore i think the check makes sence.

irnnr: ok, makes sense

Issue #690, opened 2016-10-18
[BUG] SOLR_RELATION: Recursive value resolution


dkd-friedrich:
By using the SOLR_RELATION option `enableRecursiveValueResolution` relations can be resolved recursive if the given `foreignLabelField` is a relation too.

This currently fails, since the wrong parameters for the recursive call of `getRelatedItemsFromForeignTable()` are set. We should fix this issue and add integration tests for this.

PR #691, opened 2016-10-18
[BUGFIX] Fix recursive value resolution in SOLR_RELATION

dkd-friedrich:
- Fix recursive value resolution
- Add integration tests for specific foreignLabelField and
enableRecursiveValueResolution settings

Fixes: #690


timohund:
Looks good, thanks for adding a test as well.

PR #696, opened 2016-10-21
Make rendering errors better readable in the backend

sascha-egerer:
Render indexing errors in a pre tag to make them better readable as
linebreaks and whitespaces are interpreted.


timohund:
Thx for providing a PR

PR #697, opened 2016-10-21
[BUGFIX] use TypoScript configuration object for plugin baseWrap

peterkraume:
The baseWrap function still used the old way to access the TypoScript configuration. This patch makes use of the TypoScript configuration object to make the baseWrap functional again.


peterkraume:
This patch should be applied to the release-5.1.x branch as well (in case you plan any bugfix releases for EXT:solr 5.1)

PR #704, opened 2016-10-27
[!!!][TASK] Remove old cli commands

georgringer:
Since there is now a CommandController available, the old CLI dispatcher
can be removed. Furthermore it did not work anymore since 8.0 because
of a breaking change in the core (#72368)

Fixes: #723


timohund:
I think we should talk about the 7.6 compatibility. I did an have an appointment with georg on wednesday. Please do not merge it before.

irnnr:
I agree, we should try to maintain compatibility with the previous major version

georgringer:
The code does not work on 8. To clean up code I am more in favor of removing it

PR #713, opened 2016-11-01
[BUG] Make definition of cookie in tx_solr_statistics compatible with…

thomashohn:
Make definition of cookie in tx_solr_statistics compatible with ses_id in fe_sessions


timohund:
@thomashohn Thx for your bug report and the fix!

PR #728, opened 2016-11-03
[BUGFIX] Set escapeOutput to false for ViewHelpers in Backend

georgringer:
As the Backend ViewHelpers outpout HTML, it's content must not be escaped.


georgringer:
@NamelessCoder sadly 7.6. is not yet deprecated ;)

NamelessCoder:
I know, and there's no rush to get rid of this feature in EXT:solr any time soon. But it doesn't hurt to be prepared, and it might be worth while to consider (not in this PR of course) making the EXT:solr versions of the option helper work the same as the expected core version will work.

Long story short, this patch works and is compatible with all TYPO3 versions!

PR #732, opened 2016-11-03
[TASK] Prevent SQL hammering in getConfigurationFromPageId

NamelessCoder:
This change implements a memoizer to remember each resolved
TS configuration array based on input arguments. During certain
RecordMonitor operations the SQL database gets hammered with
repeated requests to retrieve the exact same SQL resource (and
parse the result using semi-expensive logic). E.g. when indexing
multiple records on the same page.

Since TS output is the same for any page ID + language + path,
always, the problem can be solved by remembering the output
in the Util::getConfigurationFromPageId.

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -246,6 +246,13 @@ class Util
         $language = 0,
         $useCache = true
     ) {
+        static $configurationsByCacheKey = array();
+        $cacheId = md5($pageId . '|' . $path . '|' . $language);
+
+        if (isset($configurationsByCacheKey[$cacheId])) {
+            return $configurationsByCacheKey[$cacheId];
+        }
+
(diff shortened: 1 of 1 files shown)


timohund:
@NamelessCoder Thx for your pull request. I would propose in addition to swap the check and use a guard clause:

```php
if (isset($configurationsByCacheKey[$cacheId])) {
     return $configurationsByCacheKey[$cacheId]
}

// build cache
```

What do you think?

NamelessCoder:
Hi Timo - sounds perfectly fine to me! Some fear early returns, I don't :)

Will update this when I resume work tomorrow!

NamelessCoder:
Updated; switched to guard clause and early return.

Issue #739, opened 2016-11-04
Use fluid to render reports


georgringer:
Using fluid instead of a custom template solution improves the code quality and makes later changes easier. Furthermore the output has been improved by using proper tags like `<code>` for paths and versions

PR #740, opened 2016-11-04
[TASK] Use fluid to render reports

georgringer:
Using fluid instead of a custom template solution improves the code quality
and makes later changes easier.
Furthermore the output has been improved by using proper tags like <code> for paths and versions</code>

Resolves: #739

### DIFF
--- a/Classes/Report/AccessFilterPluginInstalledStatus.php
+++ b/Classes/Report/AccessFilterPluginInstalledStatus.php
@@ -24,8 +24,10 @@ namespace ApacheSolrForTypo3\Solr\Report;
  *  This copyright notice MUST APPEAR in all copies of the script!
  ***************************************************************/
 
+use ApacheSolrForTypo3\Solr\ConnectionManager;
 use ApacheSolrForTypo3\Solr\SolrService;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Fluid\View\StandaloneView;
 use TYPO3\CMS\Reports\Status;
 use TYPO3\CMS\Reports\StatusProviderInterface;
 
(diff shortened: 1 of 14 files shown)


Classes/Report/AccessFilterPluginInstalledStatus.php
```diff
@@ -63,7 +65,7 @@ class AccessFilterPluginInstalledStatus implements StatusProviderInterface
     public function getStatus()
     {
         $reports = array();
-        $solrConnections = GeneralUtility::makeInstance('ApacheSolrForTypo3\\Solr\\ConnectionManager')->getAllConnections();
+        $solrConnections = GeneralUtility::makeInstance(ConnectionManager::class)->getAllConnections();
```
NamelessCoder: This and other `::class` should probably not be part of the PR - although I agree they should be used!

Classes/Report/AccessFilterPluginInstalledStatus.php
```diff
@@ -94,27 +96,19 @@ protected function checkPluginInstallationStatus(SolrService $solrConnection
...
-                . '<li>Host: ' . $solrConnection->getHost() . '</li>'
-                . '<li>Port: ' . $solrConnection->getPort() . '</li>'
-                . '<li>Path: ' . $solrConnection->getPath() . '</li>
-				</ul>';
-
-            $message .= $this->getPluginDownloadMessage();
-
-            $status = GeneralUtility::makeInstance('TYPO3\\CMS\\Reports\\Status',
+            $standaloneView = GeneralUtility::makeInstance(StandaloneView::class);
+            $standaloneView->setTemplatePathAndFilename(
```
NamelessCoder: You could give this one a (set of) templateRootPath and then call `render($virtualAction)` - that would allow integrators to replace the template. Same for the case below :)

Classes/Report/AccessFilterPluginInstalledStatus.php
timohund: I would say both is ok, since we could change it when we work on the code anyway. Doing this in multiple pull requests is better for sure, but i think we could also be pracmatically and change it when we work in these parts.

Classes/Report/AccessFilterPluginInstalledStatus.php
timohund: Thanks for the hint. I would merge it, since the pull requests does a lot improvements and we can add this later.

Issue #766, opened 2016-11-16
[BUG] Page not queued when pages_language_overlay changed


timohund:
When a pages_language_overlay is changed the page in not added to the queue. By reading the code in the record monitor this should work.

What is the expected behaviour here?

PR #770, opened 2016-11-17
[BUGFIX] Page is not added to queue when page_language_overlay record changed

timohund:
* Fixed the bug
* Added several test cases

Fixes: #766


dkd-friedrich:
Just tested the change detection, changes of `pages_language_overlay` are detected now and the detection of other records seems to be unharmed.

Issue #788, opened 2016-11-29
Feature: Allow to access an auth protected solr server


pgampe:
This requires just a few lines of change because the integrated solr php client already has full support for this.
Pull request is on its way.

PR #789, opened 2016-11-29
[FEATURE] Supported authentication protected servers

pgampe:
Allow to access solr servers which require HTTP authentication.

Fixes: #788


Classes/ConnectionManager.php
```diff
@@ -254,6 +256,7 @@ public function getAllConnections()
                 $solrServer['solrPath'],
                 $solrServer['solrScheme']
             );
+            $this->checkAndSetAuthentication(end($connections), $solrServer);
```
irnnr: You could also assign the connection to a variable then pass it to you function and eventually push it onto `$connections`

Classes/SolrService.php
```diff
@@ -610,8 +610,9 @@ public function getSolrconfigName()
             $solrconfigXmlUrl = $this->_scheme . '://'
                 . $this->_host . ':' . $this->_port
                 . $this->_path . 'admin/file/?file=solrconfig.xml';
+            $response= $this->_sendRawGet($solrconfigXmlUrl);
```
irnnr: Is this supposed to be part of this PR?

Documentation/Configuration/Reference/TxSolrSolr.rst
```diff
@@ -69,3 +69,23 @@ path
 
 Sets the path portion of the URL. Make sure to have the path end with a slash (/).
 
+username
+--------
+
+:Type: String
+:TS Path: plugin.tx_solr.solr.username
+:Default: /
```
irnnr: If there's no default value just remove this line

irnnr:
Looks good, just some minor changes and questions.

timohund:
@pgampe  Thanks for your effort and for preparing a pull request.

I would agree to @irnnr  comments. In additionn i have some other remarks:

* The next version is 6.0, this is wrong in the documentation
* I would propose to avoid end($array) and instead spitup the call into asignment and then call the methode and finally add it to the array, or instead have $solrUser and $solrPassword as additional optional arguments of getConnection()

Documentation/Configuration/Reference/TxSolrSolr.rst
pgampe: oops, will do

pgampe:
Will look into the comments.

Classes/SolrService.php
irnnr: ok

irnnr:
as @timohund said, if you add the parameters to the constructor that's fine, too.

timohund:
@pgampe Thanks for your adaptions. The branch has merge conflicts with the master. Can you please rebase your branch?

pgampe:
Sorry,  missed the default stuff from Ingo.

timohund:
@pgampe Looks good to me, did you also check what happens when no authentication is configured.

@irnnr can you approve this change now?

pgampe:
The last time I checked, it worked fine without authentication, but I did not retest the latest rebases and master stuff. But from reading, it should only kick in if the username is set to an non-empty string.

PR #790, opened 2016-11-29
[FEATURE] Moved the actual processing of the records into a method pr…

thomashohn:
…ocessRecord in order to encapsulate the logic for updating tx_solr_indexqueue_item


timohund:
@thomashohn Looks good in general, can you please apply php-cs-fixer on the pull request?

Issue #607, opened 2016-08-16
[TASK] Query - remove by "fieldname:fieldvalue"


cihanin2code:
Hi,

It is possible to add filter-strings of form "fieldname:value" with Query->addFilter(). But there is no equivalent method for removing filters by giving the filter "fieldname:value". So a Query->removeFilterByString() would be very helpful:

```
    /**
     * @param string $filterString The filter to remove, in the form of field:value
     * @return void
     */
    public function removeFilter($filterString) {
        foreach ($this->filters as $key => $curFilterString) {
            if ($curFilterString === $filterString) {
                unset($this->filters[$key]);
            }
        }
    }
```

timohund:
Remove by filter allready exist with another scope. The methode Query::removeFilterByValue was added to implement the needed behaviour.

PR #791, opened 2016-11-29
[TASK] Added method Query::removeFilterByValue

timohund:
* Add's a method to remove a filter by the value "key:value"
* Adds tests
* Uses ::class when possible
* Replaces usage of array() with []

Fixes: #607

### DIFF
--- a/Classes/Query.php
+++ b/Classes/Query.php
@@ -24,6 +24,7 @@ namespace ApacheSolrForTypo3\Solr;
  *  This copyright notice MUST APPEAR in all copies of the script!
  ***************************************************************/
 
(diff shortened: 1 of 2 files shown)


Classes/Query.php
```diff
@@ -871,6 +872,23 @@ public function removeFilterByKey($key)
     }
 
     /**
+     * Removes a filter by the filter value. The value has the following format:
+     *
+     * "fieldname:value"
+     *
+     * @param string $filterString The filter to remove, in the form of field:value
```
irnnr: Why not split this up into two parameters `$key` and `$value`?
Who know if the colon will be changed? Or if at some point the API changes in other ways...

PR #795, opened 2016-12-01
[FEATURE] Allow overwrite of default timeout for Solr requests via ty…

thomashohn:
…poscript constant


irnnr:
I would like to avoid touching the SolrPhpClient library as much as possible. 
Can you look into another way to do this?

thomashohn:
@irnnr and @timohund I think I could move it up to ApacheSolrForTypo3\Solr\SolrService and then leave the SolrPhpClient library untouched ;-) I'll test that later today

thomashohn:
I have checked the code - and the change has to be made in Apache_Solr_HttpTransport_Abstract since all the calls to getDefaultTimeout in the underlying curl implementation hits this method. I have  encapsulated Typoscript fetching as much as possible - will update the pull-request later today

timohund:
@thomashohn 
I took your initial patch an moved the logic into SolrService. Can you please check if this is working for you as expected:

New PR #798

I'll close this pr.

Issue #794, opened 2016-12-01
[FEATURE] Overwrite _defaultTimeout


thomashohn:
It would be nice to be possible to overwrite the value of _defaultTimeout. Currently value is taken from default_socket_timeout in php.ini - however it would be to to be able to overwrite this value via a Typoscript like plugin.tx_solr.solr.defaultimeout if set. This is due to the fact the the default_socket_timeout from php.ini would affect all curl calls and not only the solr curl calls. 

Since we had a productive incident where it would have been very nice to alter this value we would hope that this pull-request could make it into the master branch for ext solr 6.0

PR #798, opened 2016-12-02
[TASK] Allow to configure connection timeout for solr

timohund:
* This PR takes the initial patch (See: #795)
* Adds a unit test
* Moves logic into SolrService
* Adds documentation

Fixes: #794


webfacer:
This doesn´t work. I already set the timeout in my TS configuration also the default_socket_time out is 60 and still getting the default. And i get still the default value of 2 which will be then devieded by 2 which will be in the end 1. Could some confirm this plz because it seems to have this issue since typo3 solr v6.1.3.
I think this is a bug. If not how can i change the timeout?

PR #802, opened 2016-12-05
Check if endtime is in the past in processDatamap.

Tuurlijk:


timohund:
@Tuurlijk What issue did you had? From reading the code i could imagine that an element was not removed when the endtime was in the past?

Tuurlijk:
@timohund That's correct. When a pages endtime is set to somewhere in the past, it was not removed from the index.

timohund:
I've tested the patch and could reproduce the bug and check your fix. Thanks for your pull request!

PR #807, opened 2016-12-06
[TASK] Extract logic from SolrService into Schema and SchemaParser to reduce complexity

timohund:
This PR:

* Extracts the schema parsing logic from the SolrService into a SchemaParser
* Introduces a schema class that holds the parsed information
* Adds unit tests for the SchemaParser and the StopWordsParser and SynonymsParser
* Use the Schema class to check the schema name


timohund:
@irnnr Thanks, good idea. I think on the long run it's good to extract more and more parts of the logic into seperat components and then into an own package.

PR #815, opened 2016-12-07
[FEATURE] Added more caching in Util::getConfigurationFromPageId

thomashohn:
I believe we can also "cache" these calls too. What do you think @NamelessCoder and @timohund ?

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -287,10 +287,7 @@ class Util
             $cache->set($cacheId, $configurationToUse);
         }
 
-        $configurationsByCacheKey[$cacheId] = self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId,
-            $language, $path);
-
-        return $configurationsByCacheKey[$cacheId];
+        return $configurationsByCacheKey[$cacheId] = self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
     }
 
     /**
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -274,22 +274,23 @@ class Util
 
         if ($initializeTsfe) {
             self::initializeTsfe($pageId, $language);
-            if (!empty($configurationToUse)) {
-                return self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
+            if (empty($configurationToUse)) {
+                $configurationToUse = self::getConfigurationFromInitializedTSFE($path);
             }
-            $configurationToUse = self::getConfigurationFromInitializedTSFE($path);
         } else {
-            if (!empty($configurationToUse)) {
-                return self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
+            if (empty($configurationToUse)) {
+                $configurationToUse = self::getConfigurationFromExistingTSFE($pageId, $path, $language);
             }
-            $configurationToUse = self::getConfigurationFromExistingTSFE($pageId, $path, $language);
         }
 
         if ($useCache) {
             $cache->set($cacheId, $configurationToUse);
         }
 
-        return $configurationsByCacheKey[$cacheId] = self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
+        $configurationsByCacheKey[$cacheId] = self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId,
+            $language, $path);
+
+        return $configurationsByCacheKey[$cacheId];
     }
 
     /**


Classes/Util.php
```diff
@@ -274,22 +274,23 @@ public static function getConfigurationFromPageId(
 
         if ($initializeTsfe) {
             self::initializeTsfe($pageId, $language);
-            if (!empty($configurationToUse)) {
-                return self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
+            if (empty($configurationToUse)) {
```
NamelessCoder: I would move this condition up one level since to avoid TSFE init when not required. Either that or I think a mistake may have snuck in here (see next comment below).

Classes/Util.php
```diff
@@ -274,22 +274,23 @@ public static function getConfigurationFromPageId(
...
             }
-            $configurationToUse = self::getConfigurationFromExistingTSFE($pageId, $path, $language);
         }
 
         if ($useCache) {
             $cache->set($cacheId, $configurationToUse);
         }
 
-        return $configurationsByCacheKey[$cacheId] = self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
+        $configurationsByCacheKey[$cacheId] = self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId,
```
NamelessCoder: These lines do not change any functionality. If a cosmetic change is desired please open a secondary pull request.

Classes/Util.php
```diff
@@ -274,22 +274,23 @@ public static function getConfigurationFromPageId(
...
-                return self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
+            if (empty($configurationToUse)) {
+                $configurationToUse = self::getConfigurationFromInitializedTSFE($path);
             }
-            $configurationToUse = self::getConfigurationFromInitializedTSFE($path);
         } else {
-            if (!empty($configurationToUse)) {
-                return self::buildTypoScriptConfigurationFromArray($configurationToUse, $pageId, $language, $path);
+            if (empty($configurationToUse)) {
+                $configurationToUse = self::getConfigurationFromExistingTSFE($pageId, $path, $language);
```
NamelessCoder: Using this function could break since TSFE is not initialised. I think this could be a mistake - that we want to use `buildTypoScriptConfigurationFromArray` instead of `getConfigurationFromExistingTSFE`. If there's a boost from using an existing TSFE and this code can potentially run in the frontend and/or works with a pre-initialised TSFE then we should check if one exists before using this method, and fallback to `buildTypoScriptConfigurationFromArray` if TSFE neither exists nor $initializeTsfe were true.

Classes/Util.php
thomashohn: I'll fix that one

Classes/Util.php
thomashohn: Hrmm - might be the case - but most calls of the method `getConfigurationFromPageId ` are called with `initializeTsfe = false` and thus it should have been a "problem" before also. I'll leave it up to @timohund to decide what should be done in this case

Classes/Util.php
thomashohn: Isn't it exactly a request for calling `self::initializeTsfe($pageId, $language);`? In most cases the method is called with false - so I don't think it's a mistake

thomashohn:
Hi @timohund/@NamelessCoder 

Ok - seems like a good idea - I could have an look at it since we are much in need of this. I suggest you close this PR and I'll create a new one :-)

Thomas

timohund:
@thomashohn $useCache = false needs to be taken into accunt for sure, thats a good point, but i think that's doable.

thomashohn:
Using the 2 level cache is rather expensive. So might i suggest that we keep both caches or skip using the 2 level cache instead?

thomashohn:
So the only calls where it is called with false is her:
`Util::getConfigurationFromPageId(0, 'plugin.tx_solr', false, 0, false);`
and that is explicitly removed - so i suggest removing the cache parameter and use the memory cache instead

timohund:
I think, i would keep it since this is also used in solrfal and i think there are cases when you want to force an uncached result. 

The two level cache, has a static in memory cache, and (if this does not produce a hit) the TYPO3 caching framework as secondlevel cache. I would not spread to much caching logic in the domain code, since this can be in controll of the cache class.

timohund:
I would say, we skip a second level cache when the cache is slower then the logic to build.

timohund:
@thomashohn I think i makes sence as you proposed that we add more of the results to the in memory cache. It took your patch and did some minor changes:

https://github.com/TYPO3-Solr/ext-solr/pull/816

Can you check if this is working in your system are you intended it? 
I'll close this pull request and it would be nice when you could add your thought's to the new pr.

PR #816, opened 2016-12-08
[FEATURE] Improve caching in getConfigurationFromPageId

timohund:
This pr:

* Improves the caching in getConfigurationFromPageId by adding the results allways to the in memory cache
* Extracts the generation of the configuration array into an own method

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -242,7 +242,7 @@ class Util
      * @param string $path The TypoScript configuration path to retrieve.
      * @param bool $initializeTsfe Optionally initializes a full TSFE to get the configuration, defaults to FALSE
      * @param int|bool $language System language uid or FALSE to disable language usage, optional, defaults to 0
-     * @param bool $useCache
+     * @param bool $useTwoLevelCache Flag to enable the two level cache for the typoscript configuration array
      * @return TypoScriptConfiguration The Solr configuration for the requested tree.
      */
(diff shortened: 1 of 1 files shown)


Classes/Util.php
```diff
@@ -252,44 +252,60 @@ public static function getConfigurationFromPageId(
...
+        if (isset($configurationObjectCache[$cacheId])) {
+            return $configurationObjectCache[$cacheId];
         }
 
         // If we're on UID 0, we cannot retrieve a configuration currently.
         // getRootline() below throws an exception (since #typo3-60 )
         // as UID 0 cannot have any parent rootline by design.
         if ($pageId == 0) {
-            return self::buildTypoScriptConfigurationFromArray(array(), $pageId, $language, $path);
+            return self::buildTypoScriptConfigurationFromArray([], $pageId, $language, $path);
```
NamelessCoder: Can we `return []` instead? Alternatively, if this method can be expensive, should we allow this to become cached? If so, we have to move the "useCache" decision up above this point and not return from this condition but assign $configurationArray instead (or cache-assign-and-return as is done below).

timohund: @NamelessCoder good point, i'll update the pull request.

timohund:
@NamelessCoder @thomashohn I've also added the first result to the cache. Can you check this in your system?

PR #824, opened 2016-12-10
[Feature] Cache call to getSolrConfigurationFromPageId in getSolrConf…

thomashohn:
…iguration


thomashohn:
In a normal situation it won't matter - but copying big installations where this method is called like 10-20.000 times it actually matters. It's ok if you just close it - I can always overwrite in my own code id needed :-)

PR #837, opened 2016-12-16
[FEATURE] Changed visibility of getSolrServersFromRegistry 

thomashohn:
from protected to public


thomashohn:
Another and better solution found :-)

Issue #838, opened 2016-12-16
[Feature] Cache sites in getAvailableSites in class Site


thomashohn:
When performing copy operations in the TYPO3 backend with solrfal enable there is a lot of "identical" calls to `getAvailableSites`. I gives a big performance boots here and other places if the sites fetched in getAvailableSites could be cache

PR #839, opened 2016-12-16
[FEATURE] Add Cache in getAvailableSites 

thomashohn:
Since solrfal and potentially other places have a large amount of identical calls to getAvailableSites it would make sense to cache the $sites

Fixes: #838


timohund:
@thomashohn Thanks for the patch!

Issue #683, opened 2016-10-12
[TASK] Hierarchical fields should have trailsing slash (v5.1.0)


AndreasA:
According to http://wiki.apache.org/solr/HierarchicalFaceting#Terms_Containing_Another_Term_in_the_Beginning

one should use trailing slashes for hierarchical field.
The reason is when using facet.prefix option in solr for instance, the hierarchy contains those fields:
"1-Books/OldBook"
"1-Books/OldBookParts"

if one uses facet.prefix=1-Books/OldBook
it would match both of those facets. 

If there is a trailing slash I can use
facet.prefix=1-Books/OldBook/

and it will not match "1-Books/OldBookParts"

timohund:
I thinks this is a valid point. Do you have the time to create a pull request?

AndreasA:
Not this week. Probably next week. Maybe the weekend (though I can't test it then)

AndreasA:
Sorry, for the delay.
Had to do a lot of things in my private time and work time.
Hopefully, I can implement it soon. It shouldn't be that hard to implement, just need the time and a testsystem where I can test solr :)

timohund:
@AndreasA Thx for your reply maybe our devbox is an option: https://github.com/TYPO3-Solr/solr-typo3-devbox by now you would need to start a docker container for Apache Solr 6.3 in parallel, but we'll add this in the next days.

AndreasA:
PR added

PR #851, opened 2016-12-21
[BUGFIX] hierarchical fields need to add trailing slash

AndreasA:
In order to be able to differentiate between 0-foo and 0-foobar
when using facet.prefix it is necessary to use a trailing slash
that way when using facet.prefix 0-foo/ only the 0-foo/ will
be found and not also 0-foobar/

Resolves: #683

### DIFF
--- a/Classes/Facet/HierarchicalFacetHelper.php
+++ b/Classes/Facet/HierarchicalFacetHelper.php
@@ -120,7 +120,7 @@ class HierarchicalFacetHelper
     {
         $menu = array();
 
-        $subMenuEntryPrefix = $level . '-' . $menuName . '/';
+        $subMenuEntryPrefix = $level . '-' . $menuName;
 
         foreach ($facetOptions as $facetOptionKey => $facetOption) {
             // find the sub menu items for the current menu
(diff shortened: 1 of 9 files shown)


timohund:
@AndreasA Thanks for the patch, i'll check it, beside that we need to keep in mind that this also requires a change in EXT:solrfluid. I'll take care of that.

timohund:
@AndreasA there is one failing integration test. I guess the assertion needs to be updated to index.php?id=1&amp;q=prices&amp;tx_solr%5Bfilter%5D%5B0%5D=pageHierarchy%253A%252F1%252F&amp; can you check it?

AndreasA:
I found an issue before I checked the test.
I fixed that issue, it was related to the URL and Hierarchical menus.

This should be fixed with the additional commit.
If this doesn't fix the hierarchy issue, I will check the results in regards to further changes.

AndreasA:
OK. Three tests still seem to be failing. Will take care of those tomorrow.

AndreasA:
One question: should we also change the variantId to have a trailing slash?
Though I think it is unnecessary as the variantId is only used with equals and who knows how often we would have to change that :)

timohund:
@AndreasA Thanks for your work on that. Regarding the variantId. Do you see a need for that? An option could be we the variantId is used to reflect a hierarchical structure, are you using it like this? I yes i would propose to open a seperate issue.

AndreasA:
No, I am not using it that way currently.
And if one intends to use it hierarchical one could always create a separate field, so I don't really see a need for that but maybe someone else does?

timohund:
I'll merge this pr and take care on the solrfluid adaption, if somebody sees a need on changing the variantId please open a new issue.

PR #860, opened 2016-12-30
[BUG] Added early return in processRecord for non configured solr-sites

thomashohn:


timohund:
@thomashohn Looks good to me, thanks for the pull request!

PR #863, opened 2017-01-04
Do not issue commits before deletes

RasmusWernerLarsen:
We're currently doing work on reducing the number of commits against our solr installation to improve scalability and cache efficiency.

Looking through the typo3 extension these two particular cases struck me as particularly bad. Essentially it seems the author was unsure of whether his deleteByQuery would delete non-committed documents and so issues an extra commit first, which is completely unnecessary since solr by default will delete both committed and pending documents.

I'm unsure of whether this may have been necessary at some point, but in 4.10+ it is not.


timohund:
Looks good to me, thanks for your pull request!

Issue #536, opened 2016-07-15
[BUG] Language overlay existence not taken into account while indexing pages


dkd-friedrich:
Patrick Broens wrote:
> Solr does not take into account if a language overlay (pages_language_overlay) exists for a page. It will always try to index a page on all Solr connections with the &L= parameter.
> 
> The Index Queue Worker Scheduler task will never reach 100% and show failures.
> 
> The log contains warnings for user _cli_scheduler like:
> `Core: Error handler (BE): PHP Warning: file_get_contents(http://domain.ext/index.php?id=1&L=2): failed to open stream: HTTP request failed! HTTP/1.0 404 Not Found in /path_to/domain_ext/typo3conf/ext/solr/Classes/IndexQueue/PageIndexerRequest.php line 144.`
> 
> Solr should check on existence of a page language overlay before indexing the page in that language. The Index Queue Worker task will then reach 100% and no failures will be shown, not in the task as in the TYPO3 log.
> 
> This is of course depending on the language fallback. In our case pages will get "404 - Page not found" when a page is not available in a certain language.

--------

https://forge.typo3.org/issues/62306

EXT:solr currently tries to index pages with all configured languages, if a page isn't translated and language fallback isn't configured and TYPO3 sending a 404 indexing will partly fail.

Value: 20
Effort: 20

PR #873, opened 2017-01-11
[FIX] avoid getting 404 page if translation of root page for language…

anjeylink:
when indexing records for non default language and root page wasn't translated it getting 404 page when init FE.

### DIFF
--- a/Classes/IndexQueue/Indexer.php
+++ b/Classes/IndexQueue/Indexer.php
@@ -184,7 +184,35 @@ class Indexer extends AbstractIndexer
      */
     protected function getFullItemRecord(Item $item, $language = 0)
     {
+        static $translationOverlaysList;
+
         $rootPageUid = $item->getRootPageUid();
+
+        // check if translation is available
+        if ($language > 0) {
+            if ($translationOverlaysList === null || !array_key_exists($rootPageUid, $translationOverlaysList)) {
+                $availableTranslations = [];
+
+                $translationOverlays = $GLOBALS['TYPO3_DB']->exec_SELECTgetRows(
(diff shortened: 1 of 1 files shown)


timohund:
@anjeylink Can you please create an issue and describe the problem there?

anjeylink:
@timohund where ? ;)

timohund:
@anjeylink here on github :)

anjeylink:
@timohund ok.
On our installation (TYPO3 8.5.1) we have a few system languages and different sites has different translations. Lets say root page with UID 1 has default language (UID 0 ) and Swedish language (UID 1), root page with UID 10  - has default language (UID 0) and Italian language (UID 2).

When it comes to indexing of news records, indexer trying to init FE for each root page and each available language. 
`Util::initializeTsfe($rootPageUid, $language);`

but if root page UID 10 doesn't have translation for Swedish language with uid 1, init FE cause throwing  404 page ( reasonText => 'Page is not available in the requested language.').

dkd-kaehm:
Can be closed? 
is it currently not checked in [master c936dea0dd29700773c6a62e429380f9373d5ac6](https://github.com/TYPO3-Solr/ext-solr/blob/c936dea0dd29700773c6a62e429380f9373d5ac6/Classes/IndexQueue/Indexer.php#L199-L213)

timohund:
Related issue is #536 

We should:

* Add a test
* Extract the logic in the patch into a method

timohund:
@anjeylink This only happens when you set the config.sys_language_mode to strict right?

anjeylink:
@timohund i think so, but i didn't test it.

Issue #868, opened 2017-01-09
[FEATURE] Add stdWrap support for plugin.tx_solr.solr values


frans-beech-it:
Would be a nice addition if `plugin.tx_solr.solr` values were processed as stdWrap values. That would make it possible to use typoscript options like https://docs.typo3.org/typo3cms/TyposcriptReference/DataTypes/Gettext/Index.html#getenv to fetch the config from System Environment Variable

frans-beech-it:
Example config I would like to use:

```
    solr {
        scheme = TEXT
        scheme.value = {$plugin.tx_solr.solr.scheme}
        scheme.override.data = getenv : SOLR_SCHEME

        host = TEXT
        host.value = {$plugin.tx_solr.solr.host}
        host.override.data = getenv : SOLR_HOST

        port = TEXT
        port.value = {$plugin.tx_solr.solr.port}
        port.override.data = getenv : SOLR_PORT

        path = TEXT
        path.value = {$plugin.tx_solr.solr.path}
        path.override.data = getenv : SOLR_CORE_NL
    }
```

timohund:
Maybe we can do this in the configuration object to have it available for all settings.

timohund:
We need to implement this for dev-master by using the TypoScriptConfiguration object and add an example to the typoscript settings.

PR #874, opened 2017-01-11
[FEATURE] Add stdWrap support for plugin.tx_solr.solr values

spoonerWeb:
Relates: #868
Releases: master


timohund:
@spoonerWeb Thx for the patch. I would propose for dev-master to implement this in the TypoScriptConfiguration object to simplify stdWrap usage for other configuration values.

Issue #870, opened 2017-01-10
An "avoid commit" setting


RasmusWernerLarsen:
I'm looking into implementing an "AvoidCommit" setting for ext-solr.

The idea would be that when this option is set, ext-solr will never issue any solr commit operations against the SOLR server, but instead rely on the solr servers autocommit feature (or say, an external job of some kind, maybe even manual action, though neither of those is our usecase).

The point is to improve scalability and performance, because in our usecase where we have approximately 500 individual sites sharing a single core, it's very difficult to control how many commits are issued and how they're distributed. The result is that, especially in maintenance situations, we can experience quite high commit rates and in particular commits with very few seconds between, this in turn causes concurrent warming searchers which is very harmful to solr performance. Also a commit will invalidate most of the high level caches, which is never a nice thing for performance if it happens all the time.

An alternative name for this setting could be "AutoCommitMode" because really that's what it comes down to. We want ext-solr to not even consider the issue of data reliability and data visibility, we want the AutoCommit interval to fully handle that.

Ofcourse, I'm not really suggesting that we conditionally override the "commit" call in the solrClient library or anything like that, if you really want to call the commit code, you should be able to, in case you have some manual administrative function that wants to hit it, but no ext-solr component should do it implicitly.

Looking for thoughts, ideas and criticism...

timohund:
@RasmusWernerLarsen Hi, i think this is a good idea and makes sence, i would propose to introduce the following setting (that defaults to true):

```
plugin.tx_solr.index.enableCommits
```

When you set it to false we would avoid the call's of commit(). What do your think?

RasmusWernerLarsen:
I like it... Will implement that.

PR #875, opened 2017-01-11
[FEATURE] Add plugin.tx_solr.index.enableCommits option.

RasmusWernerLarsen:
The option makes it possible to disable implicit solr commits, which
can improve scalability and performance when a solr core is shared
between many installations.

This affects the garbageCollector (used by the indexQueue) and the special reIndexTask.

This doesn't preclude manually triggering commits and the deleteByType function will still issue commits by default, but that can be explicitly disabled, and changing that would cause unnecessary API breakage.

I still need to test this a bit more, so please don't merge it yet.

Fixes: #870

### DIFF
--- a/Classes/GarbageCollector.php
+++ b/Classes/GarbageCollector.php
@@ -201,7 +201,7 @@ class GarbageCollector extends AbstractDataHandlerListener implements SingletonI
         $indexQueueItems = $this->getIndexQueue()->getItems($table, $uid);
         foreach ($indexQueueItems as $indexQueueItem) {
             $site = $indexQueueItem->getSite();
-            $solrConfiguration = $this->site->getSolrConfiguration();
+            $solrConfiguration = $site->getSolrConfiguration();
 
             // a site can have multiple connections (cores / languages)
             $solrConnections = $connectionManager->getConnectionsBySite($site);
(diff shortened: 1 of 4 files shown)


timohund:
@RasmusWernerLarsen can you please apply php-cs-fixer the build has shown some violations.

Classes/GarbageCollector.php
```diff
@@ -201,12 +201,15 @@ protected function deleteIndexDocuments($table, $uid)
...
         foreach ($indexQueueItems as $indexQueueItem) {
             $site = $indexQueueItem->getSite();
+            $solrConfiguration = $site->getSolrConfiguration();
 
             // a site can have multiple connections (cores / languages)
             $solrConnections = $connectionManager->getConnectionsBySite($site);
             foreach ($solrConnections as $solr) {
                 $solr->deleteByQuery('type:' . $table . ' AND uid:' . intval($uid));
-                $solr->commit(false, false, false);
+                if ($solrConfiguration->getEnableCommits()) {
```
timohund: I think, retrieving the setting could also be done outside the loop. What do you think?

RasmusWernerLarsen: Good point 👍

timohund:
Looks good to me, thanks for the pull request!

Issue #476, opened 2016-07-08
[TASK] Simplify classname usage by using ::class when possible


timohund:
Since php 5.5 calls like:

``` php
   \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance('TYPO3\CMS\Core\Imaging\IconRegistry');
```

can be changed to:

``` php
   use \TYPO3\CMS\Core\Imaging\IconRegistry;
   \TYPO3\CMS\Core\Utility\GeneralUtility::makeInstance(IconRegistry::class);
```

to make the code more readble

thomashohn:
@timohund I could go through the code and fix this

timohund:
We are done here. Thanks to @thomashohn !

PR #886, opened 2017-01-13
[TASK] Use ::class notation in getDumbMock

thomashohn:
Part of #476


thomashohn:
Just found 4 usages more to be fix and added that commit @timohund

PR #902, opened 2017-01-20
Task/master/convert array to brackets notation tests

### LINKED ISSUE #477: [TASK] Use short array syntax ['a'] instead of array('a') when possible
Since php 5.4 we can use the short array syntax 

``` php

  $foo = ['a'];
```

instead of

``` php
  $foo = array('a');
```

We can use this where possible to make the code more readable.

thomashohn:
partly fixes #477


timohund:
Lgtm. @thomashohn Thanks for the pull request!

Issue #355, opened 2016-04-18
Re-index child pages recursively after changing page title of parent page


markuskobligk:
After changing the title of a page it gets updated in the index queue and will be re-indexed.
All subpages of this page have to be re-indexed as well since the change of the page title on the parent page might change the url structure of the complete segment/tree. Depending on the URL rewriting configuration (especially the usage of the URL cache) this might lead to many broken links on the subpages.

After having a look at the record monitor I saw that the processDatamap_afterDatabaseOperations() hook calls two functions, if the updated record is a page:

if ($recordTable == 'pages') {
  $this->updateCanonicalPages($recordUid);
  $this->updateMountPages($recordUid);
}

I think this has to be extended by a third function call that updates the index queue of all subpages recursively, e.g. $this->updateChildPages($recordUid). This can be done with a few lines of code I think. The core has a function to determine all subpage uids for a given page uid.

My setup is TYPO3 7.6.5 with the latest TER version of the Solr extension.

Please let me me know if you agree with this and I will try to provide a pull request for this within the next days.

markuskobligk:
Hm, it seems that I looked into the master too late.
This is already fixed, right?

if ($this->isRecursiveUpdateRequired($recordUid, $fields)) {
  $treePageIds = $this->getSubPageIds($recordUid);
  foreach ($treePageIds as $treePageId) {
  $this->indexQueue->updateItem('pages', $treePageId);
  }
}

markuskobligk:
After trying the master-Branch version and taking a deeper look into the code I have re-open this.

The current implementation in the master only respects changes to the page properties "extendToSubpages" and "hidden". Changing a page title has also to be respected I think and  tx_realurl_pathsegment should be respected as well.

Waht do you think?

dkd-friedrich:
https://forge.typo3.org/issues/14094
https://forge.typo3.org/issues/29223
https://forge.typo3.org/issues/44015

Value: 40
Effort: 100

irnnr:
Needs a way to configure/define fields that trigger recursive updates

FamousWolf:
Wouldn't it be easier to solve this by using non-RealURL URLs when indexing and converting them to RealURL URLs in the search results? This will probably make displaying search results more resource heavy (not sure how much), but it would solve the issues with subpages when moving or renaming pages.

I'm don't know enough about the inner workings of the solr and realurl extensions to know if this is hard to do though.

thomashohn:
[...]

PR #913, opened 2017-01-22
[FEATURE] Enable recursive page Indexing when configured through new …

thomashohn:
…TS constant plugin.tx_solr.index.queue.pages.recursivePageUpdateFields.

Should solve the issue #355

### DIFF
--- a/Classes/AbstractDataHandlerListener.php
+++ b/Classes/AbstractDataHandlerListener.php
@@ -84,6 +84,40 @@ abstract class AbstractDataHandlerListener
     }
 
     /**
+     * Check if an page update will trigger an recursive update of pages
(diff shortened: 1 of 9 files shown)


timohund:
@thomashohn Thanks for the pull request. I would propose the followings changes:

* The queue Configuration can be anything, no only "pages" so we should pass the queue configuration, when fetching the fields. Otherwise this feature will only work for pages with the queue configuration "pages".
* I would propose to change the naming from plugin.tx_solr.index.queue.pages.recursivePageUpdateFields to plugin.tx_solr.index.queue.pages.recursiveUpdateFields because using this in a pages table implies that.
* I think we should cover this usecase with an integration test.

PR #924, opened 2017-01-27
[FEATURE] allow to toggle index clearing in ReIndexTask

tgriessbach:
* adds a checkbox to ReIndexTask
* calls cleanUpIndex() only if checked 

This allows to periodically reset the IndexQueue without killing the while Search Index.


Classes/Task/ReIndexTaskAdditionalFieldProvider.php
```diff
@@ -132,6 +132,13 @@ public function getAdditionalFields(
             'cshLabel' => ''
         );
 
+        $additionalFields['clearSearchIndex'] = array(
+            'code' => '<input type="checkbox" name="tx_scheduler[clearSearchIndex]" value="1" ' . (($task->shouldClearSearchIndex()) ? 'checked="checked"' : '') . ' />',
+            'label' => 'Clear Index for selected sites and record types',
```
irnnr: Can you move this to a localizable label please?

Classes/Task/ReIndexTask.php
```diff
@@ -56,6 +56,13 @@ class ReIndexTask extends AbstractTask
      */
     protected $indexingConfigurationsToReIndex = array();
 
+    /**
+     * Clear Index for selected sites and record types
+     *
+     * @var boolean
+     */
+    protected $clearSearchIndex;
```
irnnr: I'd suggest a default value of `false` just to make sure existing tasks are ok when they get deserialized after an update to a version containing this code.

irnnr (changes requested):

aWuttig (commented):
I think its better to use array(), because EXT:solr maybe support PHP 5.3 applications and this syntax is only supported since 5.4

http://php.net/manual/de/migration54.new-features.php

tgriessbach:
Thanks for the feedback, I changed it according to your suggestions.

irnnr:
@tgriessbach just noticed; if you want this to go into the 3.1 branch you need to make a PR for master first (and the release branches in between). Also, like @aWuttig already mentioned, since the 3.1 branch supports TYPO3 v6LTS, you need to use the old array syntax (does not apply for master).

Classes/Task/ReIndexTask.php
```diff
@@ -54,7 +54,14 @@ class ReIndexTask extends AbstractTask
      *
      * @var array
      */
-    protected $indexingConfigurationsToReIndex = array();
+    protected $indexingConfigurationsToReIndex = [];
```
irnnr: Can't use short array syntax for the 3.1 release branch.

irnnr (changes requested):

PR #926, opened 2017-01-30
[TASK] Refactor getInitialPagesAdditionalWhereClause

### LINKED ISSUE #917: Refactor getInitialPagesAdditionalWhereClause
The method getInitialPagesAdditionalWhereClause should also take $configurationName as an argument. In order to be able to implement this change - it's needs to wait for the changes proposed in #915 - since this will move additional functionality required where getInitialPagesAdditionalWhereClause is called in the Site class

thomashohn:
Refactoring of getInitialPagesAdditionalWhereClause to make it compliant with other
page configuration TS definitions. Also added unittest at documenttation.
Should fix #917


timohund:
@thomashohn Thanks thomas i would avoid to make all method static, because static methods are allways hard to test. I would propose to move getIndexingConfigurationName & getRecordIfIndexConfigurationIsValid & isValidTableForIndexConfigurationName to System/ConfigurationManager, because this is configuration related. Afterwards getIndexingConfigurationName and isValidTableForIndexConfigurationName could be public. This would also help to extract configuration related logic from the RecordMonitor and GarbageCollector.

timohund:
@thomashohn Thanks for the patch, now additionalWhereClause is useable for all pages configurations and the structure was improved.

Issue #862, opened 2016-12-31
[TASK] Allow to disable siteHash check


timohund:
The setting plugin.tx_solr.search.query.allowedSites = can be used by now, to enable a search on different sites. But this has currently the limitation that only sites from the same TYPO3 system can be used for search:

* \* and __all means: All sites of the same TYPO3 System
* __current_site means only the current site
* and it is possible to configure domains within the same system

I would suggest to make a (breaking) change here to allow the search on any site (also when it is located in another instance. Afterwards we would have the following meanings:

* \*   Any site (no sitehash check)
* __all (all sites from the instance)
* and it is possible to configure domains within the same system

Another approach would be to introduce a new setting like __any or __global but then the meaning of * would be confusing for me. Anyhow we need a possibility to disabled this check.

irnnr:
I don't think this is a good idea / I don't think this should be implemented. The purpose of the site hash is basically so that you could index multiple sites into one core/index.

Besides that there is EXT:solr_disablesitehash for exactly those use cases where someone would want to disable the site hash. https://github.com/TYPO3-Solr/ext-solr_disablesitehash

timohund:
@irnnr I know this extension. What about to move this functionallity into EXT:solr? i think this simpflies the maintenance and testing of this a lot.

irnnr:
What maintenance? There's not much to do, is it?

The site hash really is essential to quite a few things in the extension / it's used in quite some areas. So the "feature" - rather ability - to turn it off should be left to people / use cases where the user absolutely knows what they're doing. The extension is easy enough to install and use.

Why?:

I think the best explanation is the department scenario:  
A large installation with multiple sites for each department (university/company). They all index into the same core. By having the site hash you can ensure that while you're on a department's site you will only results for that particular department. When you search from the organization's umbrella/head/general site you could by default also search only that. However, you could implement a checkbox "search entire organization", that when checked would disable the site hash and thus returning results from all departments.

The site hash also ensures that in this scenario, when emptying/re-initializing an index, only documents for a particular site are affected.

Hope that makes sense.

PR #929, opened 2017-01-30
[!!!][TASK] Allow to disable siteHash check by setting query.allowedSites to *

timohund:
This PR changes the behaviour of `query.allowedSites`; the behavior `*` has been changed:

* Before: `*` was the same as `__all`, which means all sites in the system
* After: `__all` is still handled as `__all` sites in the system, but `*` now means every site (same as no check at all)

## Implementation Details

During the implementation the following things have been done

* Move SiteHash related logic from Util* to SiteHashService* and make them non static (See: http://wiki.c2.com/?AbuseOfUtilityClasses)
* Mark SiteHash related method in Util deprecated and announce removal
* Added Tests for Query and SiteHashService modifications
* If variants should be used across multiple instances the `variantId` needs to be unique by system (not by site!). As of now the `variantId` only contains (type/uid), to have a unique `variantId` across multiple systems, a system hash was added.

## Impact & Migration

* If you use `*` for query.allowedSites please use `__all` now unless you want to disable the siteHash check
* If you use variants, it is recommended to reindex your pages, since the hash was changed.

Fixes: #862

### DIFF
--- a/Classes/Domain/Site/SiteHashService.php
+++ b/Classes/Domain/Site/SiteHashService.php
@@ -1,0 +1,129 @@
+<?php
+
+namespace ApacheSolrForTypo3\Solr\Domain\Site;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2017- Timo Hund <timo.hund@dkd.de>
+ *  All rights reserved
+ *
+ *  This script is part of the TYPO3 project. The TYPO3 project is
+ *  free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  The GNU General Public License can be found at
+ *  http://www.gnu.org/copyleft/gpl.html.
+ *
+ *  This script is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  This copyright notice MUST APPEAR in all copies of the script!
+ ***************************************************************/
+
+use ApacheSolrForTypo3\Solr\Site;
+
+/**
+ * SiteHashService
+ *
+ * Responsible to provide sitehash related service methods.
+ *
+ * @author Timo Hund <timo.hund@dkd.de>
+ */
+class SiteHashService
+{
+
+    /**
+     * Resolves magic keywords in allowed sites configuration.
+     * Supported keywords:
+     *   __solr_current_site - The domain of the site the query has been started from
+     *   __current_site - Same as __solr_current_site
+     *   __all - Adds all domains as allowed sites
+     *   * - Means all sites are allowed, same as no siteHash
+     *
+     * @param integer $pageId A page ID that is then resolved to the site it belongs to
+     * @param string $allowedSitesConfiguration TypoScript setting for allowed sites
+     * @return string List of allowed sites/domains, magic keywords resolved
+     */
+    public function getAllowedSitesForPageIdAndAllowedSitesConfiguration($pageId, $allowedSitesConfiguration)
+    {
+        if ($allowedSitesConfiguration === '__all') {
+            return  $this->getDomainListOfAllSites();
+        } elseif ($allowedSitesConfiguration === '*') {
+            return '*';
+        } else {
+            return $this->getDomainByPageIdAndReplaceMarkers($pageId, $allowedSitesConfiguration);
+        }
+    }
+
+    /**
+     * Gets the site hash for a domain
+     *
+     * @param string $domain Domain to calculate the site hash for.
+     * @return string site hash for $domain
+     */
+    public function getSiteHashForDomain($domain)
+    {
+        static $siteHashes = [];
+        if (isset($siteHashes[$domain])) {
+            return $siteHashes[$domain];
+        }
+
+        $siteHashes[$domain] = sha1($domain . $GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'] . 'tx_solr');
+        return $siteHashes[$domain];
+    }
+
+
+    /**
+     * Returns a comma separated list of all domains from all sites.
+     *
+     * @return string
+     */
+    protected function getDomainListOfAllSites()
+    {
+        $sites = $this->getAvailableSites();
+        $domains = [];
+        foreach ($sites as $site) {
+            $domains[] = $site->getDomain();
+        }
+
+        $allowedSites = implode(',', $domains);
+        return $allowedSites;
+    }
+
+    /**
+     * Retrieves the domain of the site that belongs to the passed pageId and replaces their markers __solr_current_site
+     * and __current_site.
+     *
+     * @param integer $pageId
+     * @param string $allowedSitesConfiguration
+     * @return string
+     */
+    protected function getDomainByPageIdAndReplaceMarkers($pageId, $allowedSitesConfiguration)
+    {
+        $domainOfPage = $this->getSiteByPageId($pageId)->getDomain();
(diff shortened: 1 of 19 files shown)


Classes/Domain/Site/SiteHashService.php
```diff
@@ -0,0 +1,129 @@
...
+     *   __all - Adds all domains as allowed sites
+     *   * - Means all sites are allowed, same as no siteHash
+     *
+     * @param integer $pageId A page ID that is then resolved to the site it belongs to
+     * @param string $allowedSitesConfiguration TypoScript setting for allowed sites
+     * @return string List of allowed sites/domains, magic keywords resolved
+     */
+    public function getAllowedSitesForPageIdAndAllowedSitesConfiguration($pageId, $allowedSitesConfiguration)
+    {
+        if ($allowedSitesConfiguration == '__all') {
```
thomashohn: Use strict ===

Classes/Domain/Site/SiteHashService.php
```diff
@@ -0,0 +1,129 @@
...
+     * @return string site hash for $domain
+     */
+    public function getSiteHashForDomain($domain)
+    {
+        static $siteHashes = [];
+        if (isset($siteHashes[$domain])) {
+            return $siteHashes[$domain];
+        }
+
+        $siteHashes[$domain] = sha1($domain . $GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'] . 'tx_solr');
```
thomashohn: :-)

Classes/Domain/Variants/IdBuilder.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+namespace ApacheSolrForTypo3\Solr\Domain\Variants;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2010-2017 Timo Hund <timo.hund@dkd.de
```
thomashohn: 2017 missing >

Classes/Domain/Variants/IdModifier.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace ApacheSolrForTypo3\Solr\Domain\Variants;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2010-2017 Timo Hund <timo.hund@dkd.de
```
thomashohn: If new file 2017 and agin missing >

Classes/Domain/Site/SiteHashService.php
```diff
@@ -0,0 +1,129 @@
...
+        foreach ($sites as $site) {
+            $domains[] = $site->getDomain();
+        }
+
+        $allowedSites = implode(',', $domains);
+        return $allowedSites;
+    }
+
+    /**
+     * Retrieves the domain of the site that belongs to the passed pageId and replaces ther markers __solr_current_site
```
thomashohn: their?

Classes/Domain/Site/SiteHashService.php
```diff
@@ -0,0 +1,129 @@
...
+     * and __current_site.
+     *
+     * @param $pageId
+     * @param $allowedSitesConfiguration
+     * @return mixed
```
thomashohn: Is it mixed or string as return type?

Classes/Domain/Variants/IdBuilder.php
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Applies configured postProcessing hooks to build a custom variantId.
+     *
+     * @param string $variantId
+     * @param string $systemHash
+     * @param string $type
+     * @param integer $uid
+     * @return string
+     */
+    protected function applyHook($variantId, $systemHash, $type, $uid)
```
thomashohn: Maybe it would be better to use the new Signal concept in TYPO3? - I think hooks are only kept for compatibility

Classes/Domain/Variants/IdBuilder.php
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /**
+     * Returns a system unique hash.
+     *
+     * @return string
+     */
+    protected function getSystemHash()
+    {
+        $siteName = isset($GLOBALS['TYPO3_CONF_VARS']['SYS']['sitename']) ? $GLOBALS['TYPO3_CONF_VARS']['SYS']['sitename'] : 'none';
```
thomashohn: Im not sure TYPO3 works with this $GLOBALS['TYPO3_CONF_VARS']['SYS']['sitename'] - but I think I would throw an exception instead

Classes/Domain/Variants/IdBuilder.php
irnnr: signals don't work for every scenario. If you want to change data, you still need hooks.

Classes/Domain/Site/SiteHashService.php
```diff
@@ -0,0 +1,129 @@
+<?php
+
+namespace ApacheSolrForTypo3\Solr\Domain\Site;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2016 Timo Hund <timo.hund@dkd.de>
```
irnnr: `$year++`

Tests/Unit/Domain/Site/SiteHashServiceTest.php
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * @test
+     */
+    public function getSiteHashForDomain()
+    {
+        $oldKey = $GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'];
+        $GLOBALS['TYPO3_CONF_VARS']['SYS']['encryptionKey'] = 'testKey';
+
+        $service = new SiteHashService();
+        $hash1 = $service->getSiteHashForDomain('www.test.de');
```
irnnr: Please use `www.example.com` for such use-cases. `test.de` is Stiftung Warentest.

Classes/Domain/Variants/IdBuilder.php
thomashohn: Hi @irnnr and @timohund 

Not convinced - see https://usetypo3.com/signals-and-hooks-in-typo3.html and https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/core/Documentation/Changelog/master/Feature-79387-AddSignalToExcludeTablesFromReferenceIndex.rst - or have I missed something in this method?

Classes/Domain/Site/SiteHashService.php
timohund: Done

Classes/Domain/Site/SiteHashService.php
timohund: Just moved, the hash should be the same as before

Classes/Domain/Variants/IdBuilder.php
timohund: Done

Classes/Domain/Variants/IdModifier.php
timohund: Done

Classes/Domain/Variants/IdBuilder.php
timohund: I think signals and slots are more like "events" and the intension is normally a one way direction. It is possible to return data, but only by passing a variable by reference (what is not optimal i would say). I think with hooks you can directly return a value, what is needed in this case (Add btw. in the changes of the core also new hooks are added).

Classes/Domain/Variants/IdBuilder.php
timohund: I'll propose to throw an exception when sitename is not set.

Classes/Domain/Site/SiteHashService.php
timohund: Done

Tests/Unit/Domain/Site/SiteHashServiceTest.php
timohund: Done

Issue #485, opened 2016-07-11
[TASK]  Record Monitor cannot track records outside a site


dkd-friedrich:
A common scenario is to have a site and a storage folder for records parallel to it on the same level (f.e.)

Currently the Record Monitor tries to find the indexing configuration on a record's parent page (pid). For such a scenario the site cannot be found and thus no configuration can be determined, leading to the record not being updated in the Index Queue and also not being re-indexed.

---

https://forge.typo3.org/issues/31519
https://forge.typo3.org/issues/42521
https://forge.typo3.org/issues/43349
https://forge.typo3.org/issues/47110
https://forge.typo3.org/issues/55701

Outside the site root the indexing configuration is not available, but it should be possible to check all existing site roots. 

Value: 20
Effort: 13

irnnr:
Possible solution:

* detect if a sys folder is located within a tree with a root page
* if that is not the case add a (multi-)select field to sys folder page properties
* the select field shows the available sites configured for Solr indexing
* this way the user can select which configurations should apply to that folder

AndreasA:
In that regard it would also be good to have a setting that allows tables that have set TCA
```
'rootLevel' => true
```
set to be monitored and initialized.

Due to it being the pageId 0 setting
```
additionalPageIds = 0
```
will fail due the empty check 
```
if (!empty($this->indexingConfiguration['additionalPageIds']))
```

As workaround one could set 
```
additionalPageIds = 00
```
or similar but it would be nice if that were "easier"

irnnr:
that's a different issue, and IMHO indexing from root page should not be possible at all

AndreasA:
Well, I have a task that imports data via SOAP-interface therefore those data isn't editable, not even viewable (hideTable => true). So storing it at page 0 makes sense as that page will never be deleted.
However, it should be indexed in solr.
I could import it directly to solr but I also need that data as reference in other tables, so it is just easier to import it to the database and use default typo3 features for it (suggest etc.)

Therefore, there are cases where indexing such data can make sense. Probably not often the case but still.
In my case I am now importing the data now to the web root as that page is very unlikely to be deleted :)
And it is related to the additionalPageIds which is why I mentioned it here but I can create a own bug if that is better.

dkd-dobberkau:
I remember times, when we used a sysfolder holding that kind of general data. no need for packing the root page with payload.

irnnr:
[...]

PR #944, opened 2017-02-06
[TASK] RecordMonitor does not track records outside the siteroot.

timohund:
With the configuration "additionalPageIds" it was possible to add additional page ids to an indexing configuration.
When these external records have been updated in the backend by now the RecordMonitor does not recognize these updates.

This PR allows the RecordMonitor to determine all responsible site roots and add the entry to the queue when the queue get's intialized.

Fixes: #485

### DIFF
--- a/Classes/Domain/Index/Queue/RecordMonitor/Helper/ConfigurationAwareRecordService.php
+++ b/Classes/Domain/Index/Queue/RecordMonitor/Helper/ConfigurationAwareRecordService.php
@@ -44,11 +44,8 @@ class ConfigurationAwareRecordService
      * @param TypoScriptConfiguration $solrConfiguration
      * @return string Name of indexing configuration
      */
-    public function getIndexingConfigurationName(
-        $recordTable,
-        $recordUid,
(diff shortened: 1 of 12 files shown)


timohund:
Hi @thomashohn , hi @irnnr,

i just want to create this PR as a base for a discussion.

I’ve implemented in this PR the functionality that the RecordMonitor can track records outside the side root (monitor changes configured in <indexQueueName>.additionalPageIds .

Along with that i’ve refactored some code and make parts in the IndexQueue a lot simpler since this stuff is now done at one one place.

Beside that i have some things to discuss:

* Determining the responsible rootPages and configurations is a quite expensive task, but i think it is ok by adding a few caches on the right place.
* @thomas maybe you can check the impact in a real project with many site roots?
* I think we have a lot of tests and we would recognize big problems but still we should spend some time in manual testing
* When we reference a record in multiple siteRoots one record can also occur in multiple sites, therefore some bigger Changes in the IndexQueue and the RecordMonitor where needed

Issue #948, opened 2017-02-07
[TASK] Removed usage of Ext.grid.RowExpander.js - deprecated


thomashohn:
The usage of Ext.grid.RowExpander.js is deprecated with [TYPO3 8.x](https://docs.typo3.org/typo3cms/extensions/core/8-dev/Changelog/8.4/Breaking-77783-RemoveUnusedExtJSLibraries.html?highlight=rowexpander) - for the moment the version from TYPO3 7.6.x has been added as a temporary solution as part of #901

thomashohn:
@timohund Can you look into this one then 👍

PR #949, opened 2017-02-07
[TASK] Fix deprecations with TYPO3 8.x

thomashohn:
Removed $GLOBALS['PATH_solr'] , $GLOBALS['PATHrel_solr'], ExtensionManagementUtility::extRelPath('solr') and temp fix for #948

### DIFF
--- a/ext_localconf.php
+++ b/ext_localconf.php
@@ -3,14 +3,12 @@ if (!defined('TYPO3_MODE')) {
     die('Access denied.');
 }
 
-$GLOBALS['PATH_solr'] = \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::extPath('solr');
-
 # ----- # ----- # ----- # ----- # ----- # ----- # ----- # ----- # ----- #
 
 // Windows compatibility
 
 if (!function_exists('strptime')) {
-    require_once($GLOBALS['PATH_solr'] . 'Resources/Private/Php/strptime/strptime.php');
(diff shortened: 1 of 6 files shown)


ext_localconf.php
```diff
@@ -3,14 +3,12 @@
     die('Access denied.');
 }
 
-$GLOBALS['PATH_solr'] = \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::extPath('solr');
```
irnnr: Why?

irnnr:
What's the point of removing those shortcut globals?

thomashohn:
First of all - they are not really used at that many places and there they are used the "EXT:solr/..." notation can be used instead. If you have good reasons for keeping them - we can keep them :-)

thomashohn:
@irnnr Any news her?

timohund:
@thomashohn Thanks for your pull request! I'll try to have a look on it today. In general it is good and important that we fix it before the release for 8 LTS. Regarding the globals we also need to check solrfal and solrfluid if they are used there.

timohund:
@thomashohn @irnnr for me this looks good. I've created an Issue for version 7.0 to drop the compatibility code for TYPO3 7.6 LTS (#950)

timohund:
@thomashohn I've checked the patch with 7.6 LTS and the index inspector does not show up.

thomashohn:
@timohund Should work with 7.6.15 apparently EXT:extension name is not fully supported in addJS in 7.6.15

Issue #939, opened 2017-02-05
[TASK] Add indexes to the pages table


thomashohn:
I have been profiling the Solr extension with TYPO3 8.x via blackfire.io. 

In the "normal" case with creating pages/elements etc. performance is ok. The case where an existing page-tree is copied shows major performance issues in relation to the database.
The test consists of a page tree with 6 nodes, 366 tt_content elements and 4 file references.
As seen on the screen-shot there seems to be 2 missing indexes and a rewrite of the SQL statement in getDestinationMountPropertiesByRootLine.
<img width="889" alt="solr-missing-indexes" src="https://cloud.githubusercontent.com/assets/3414869/22627016/3a18fe3e-ebba-11e6-95b3-af39f9643c74.png">
Adding the following two indexes:
```
alter table pages add index content_from_pid_deleted (content_from_pid, deleted);
alter table pages add index doktype_no_search_deleted (doktype, no_search, deleted);
```
gives the following performance boost:
<img width="866" alt="solr-adding-indexes" src="https://cloud.githubusercontent.com/assets/3414869/22627026/52b0588e-ebba-11e6-9875-cc75ffd4db35.png">

A PR can be provided

irnnr:
This seems to affect only mount pages?

thomashohn:
No - since the RecordMonitor executes the `update` from the class `MountPagesUpdater` the SQL statement will be executed in all cases - causing the overhead - or do I miss something?

irnnr:
ok

thomashohn:
Merged

PR #954, opened 2017-02-09
[TASK] Added some important indexes to pages 

thomashohn:
Added som important indexes to the pages table to speed up certain solr  
operations while copying big sites solves #939


thomashohn:
I can try to get it into TYPO3 core - but normally they are not that keen on adding stuff that comes from a specific scenario in a specific extension - but i'll see what I can do - okay - if not I will suggest we add it.

dkd-dobberkau:
OK. I am not opposing this in general. but please add a note to the documentation then, about this change in the pages table.

thomashohn:
So one could argue that a index with doktype and delete could be generally good for TYPO3 but the other one I doubt.

Issue #940, opened 2017-02-05
[TASK] Add runtime Cache for isRootPage in Class Util.php


thomashohn:
I have been profiling the Solr extension with TYPO3 8.x via blackfire.io. 

In the "normal" case with creating pages/elements etc. performance is ok. The case where an existing page-tree is copied shows major performance issues in relation to the database.
The test consists of a page tree with 6 nodes, 366 tt_content elements and 4 file references.

<img width="1267" alt="solr-isrootpage-non-cached" src="https://cloud.githubusercontent.com/assets/3414869/22627130/287a57ba-ebbd-11e6-8864-f3f1e8557162.png">

The impact of the Run-time cache is quite impressive.

<img width="1268" alt="solr-isrootpage-cached" src="https://cloud.githubusercontent.com/assets/3414869/22627137/397baea6-ebbd-11e6-9849-985bd4bf950c.png">

A PR can be provided

PR #958, opened 2017-02-10
[TASK] Added runtime-cache to isRootPage in Util Class

thomashohn:
This also gains some performance when copying trees - should solve #940

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -504,9 +504,20 @@ class Util
      */
     public static function isRootPage($pageId)
     {
-        $page = BackendUtility::getRecord('pages', $pageId);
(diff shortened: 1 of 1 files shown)


timohund:
@thomashohn thanks for the patch. As follow up i would propose to move the code into Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php and mark it in Util as deprecated to simplfy the Util class in the long run.

thomashohn:
So merge this and it is and move it afterwards?

timohund:
@thomashohn yes, i'll check the pr, merge it and afterwards unify it as a followup.

Issue #941, opened 2017-02-05
[TASK] Add runtime Cache for getRecordIfIndexConfigurationIsValid


thomashohn:
I have been profiling the Solr extension with TYPO3 8.x via blackfire.io. 

In the "normal" case with creating pages/elements etc. performance is ok. The case where an existing page-tree is copied shows major performance issues in relation to the database.
The test consists of a page tree with 6 nodes, 366 tt_content elements and 4 file references.

<img width="1279" alt="solr-getrecordifindexconfigurationisvalid-noncache" src="https://cloud.githubusercontent.com/assets/3414869/22627193/4a767154-ebbe-11e6-80d9-f528f5db8e69.png">

The impact of the Run-time cache is quite impressive.

<img width="1282" alt="solr-getrecordifindexconfigurationisvalid-cached" src="https://cloud.githubusercontent.com/assets/3414869/22627195/523361e0-ebbe-11e6-8537-68b0b7c53ab7.png">

A PR can be provided

PR #959, opened 2017-02-10
Task/master/add runtimecache configuration aware record service

thomashohn:
This also gain some performance when copying trees - should solve #941


Classes/Domain/Index/Queue/RecordMonitor/Helper/ConfigurationAwareRecordService.php
```diff
@@ -112,7 +114,18 @@ protected function getRecordIfIndexConfigurationIsValid($recordTable, $recordUid
             return $this->getPageOverlayRecordIfParentIsAccessible($recordUid, $recordWhereClause);
         }
 
-        return (array)BackendUtility::getRecord($recordTable, $recordUid, '*', $recordWhereClause);
+        $cache = GeneralUtility::makeInstance(TwoLevelCache::class, 'cache_runtime');
```
timohund: I would also put the logic to get the record into an own method getRecord()

thomashohn:
New version coming up with getRecordForIndexConfigurationIsValid :-)

Issue #942, opened 2017-02-05
[TASK] Add runtime Cache for addItem in class Queue


thomashohn:
I have been profiling the Solr extension with TYPO3 8.x via blackfire.io. 

In the "normal" case with creating pages/elements etc. performance is ok. The case where an existing page-tree is copied shows major performance issues in relation to the database.
The test consists of a page tree with 6 nodes, 366 tt_content elements and 4 file references.

<img width="1265" alt="solr-additem-noncached" src="https://cloud.githubusercontent.com/assets/3414869/22627260/d999c0d8-ebbf-11e6-98e7-d990af764209.png">

The impact of the Run-time cache is quite impressive - please note that a major improvement comes also from the #940 too.

<img width="1269" alt="solr-additem-cached" src="https://cloud.githubusercontent.com/assets/3414869/22627261/de37b244-ebbf-11e6-9eb1-8e281ce1ea28.png">

A PR can be provided

PR #960, opened 2017-02-10
[TASK] Added run-time cache to addNewItem

thomashohn:
This also gains some performance when copying trees performance when copying trees - should solve #942


Classes/IndexQueue/Queue.php
```diff
@@ -351,7 +351,15 @@ private function addNewItem($itemType, $itemUid, $indexingConfiguration, $rootPa
             $additionalRecordFields = ', doktype, uid';
         }
 
-        $record = BackendUtility::getRecord($itemType, $itemUid, 'pid' . $additionalRecordFields);
+        $cache = GeneralUtility::makeInstance(TwoLevelCache::class, 'cache_runtime');
```
timohund: I would propose to move the code into a protected method "getRecordCached" not enhance the lengh of the original method.

thomashohn:
New version ready :-)

Issue #943, opened 2017-02-06
[TASK] Add runtime Cache to RecordMonitor


thomashohn:
I have been profiling the Solr extension with TYPO3 8.x via blackfire.io.

In the "normal" case with creating pages/elements etc. performance is ok. The case where an existing page-tree is copied shows major performance issues in relation to the database.
The test consists of a page tree with 6 nodes, 366 tt_content elements and 4 file references.
The problem seems to be that the TYPO3 core issues a lot of "nearly identical" `processDatamap_afterDatabaseOperations` operations for tt_content elements with status new.

![solr-recordmonitor-noncache](https://cloud.githubusercontent.com/assets/3414869/22638001/b3cf2c86-ec45-11e6-9a3a-f200d15622a6.png)

Until this is getting fixed/if it's going to be fixes - see https://forge.typo3.org/issues/79635 - the RecordMonitor could benefit from marking what has been processed or not. 
![solr-recordmonitor-cached](https://cloud.githubusercontent.com/assets/3414869/22638004/c0d8c39c-ec45-11e6-912f-4da3fe249860.png)

A PR can be provided.

PR #961, opened 2017-02-10
[TASK] Added runtime cache to RecordMonitor

thomashohn:
The method processDatamap_afterDatabaseOperations in RecordMonitor needs a to filter multiple processings of identical record to a a "bug" in TYPO3 core
This also gains some performance when copying trees performance when copying trees - should solve #943

### DIFF
--- a/Classes/IndexQueue/RecordMonitor.php
+++ b/Classes/IndexQueue/RecordMonitor.php
@@ -243,18 +243,9 @@ class RecordMonitor extends AbstractDataHandlerListener
         $recordTable = $table;
         $recordUid = $uid;
 
-        // Check if record has already been processed since DataHandler sends processDatamap_afterDatabaseOperations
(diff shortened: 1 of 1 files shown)


Classes/IndexQueue/RecordMonitor.php
```diff
@@ -242,6 +243,19 @@ public function processDatamap_afterDatabaseOperations(
         $recordTable = $table;
         $recordUid = $uid;
 
+        // Check if record has already been processed since DataHandler sends processDatamap_afterDatabaseOperations
```
timohund: We should move this code into an own method "isRecordAllreadyProcessed" which returns true or false and stepout when it returns true

thomashohn: I think hasRecordBeenProcessed is a better name - or?

thomashohn:
Another version coming up :-)

Issue #955, opened 2017-02-09
Go through BackendUtility::getRecord


thomashohn:
Go through all usages of  BackendUtility::getRecord. There are a couple of usages where the whole record is fetched instead of just the column needed for a check. 
This reduces the amount I/O intensive operations in Solr which matters in Cloud setups.

PR can be provided

PR #971, opened 2017-02-15
[TASK] Only fetch necessary data in BackendUtility::getRecord

thomashohn:
Should solve #955.


timohund:
@thomashohn Looks good, there is one new Issue by scrutinizer, can you please add a typecheck to the result of getRecord and throw an exception when it is not an array? (I know this was also missing before ;) )

timohund:
@thomashohn Thanks for the update. LGTM, will merge it when the build is done.

PR #973, opened 2017-02-15
[TASK] Mark addSortField and getSortingFields as deprecated

### LINKED ISSUE #603: Check usage of Query::addSortField
Currently no ussage of this method can be found. Is it used? Can we remove it?

thomashohn:
Should fix #603


timohund:
Looks good to me thanks

Issue #957, opened 2017-02-10
[TASK] Move all dateFormatting functions from Util to System/DateTime/FormatService


timohund:
To reduce the complexity of the Util class, we should move the DateFormatting functions to the responsible service (System/DateTime/FormatService). The old method should be marked as deprecated and to be removed in version 7. The util method should use the service to provide backwards compatibility.

The following functions a relevant.

* timestampToIso
* isoToTimestamp
* timestampToUtcIso
* utcIsoToTimestamp

thomashohn:
Hi Timo - I could have a look at this one?

timohund:
@thomashohn sure, would be nice, at the same time we could add a unit test for them.

PR #977, opened 2017-02-16
[TASK] Moved Time related methods to FormatService 

thomashohn:
Deprecated them in Util. Added some Unit tests too.

Fixes: #957


timohund:
Thanks for the patch @thomashohn

PR #1017, opened 2017-02-17
[TASK] Fix scrutinizer issues in HierarchicalFacetRenderer

### LINKED ISSUE #1009: [TASK] Fix scrutinizer issues in HierarchicalFacetRenderer

thomashohn:
Should fix #1009


timohund:
@thomashohn thanks for the fix. A next step could be to extract the stuff in if(!empty()) to a method.

PR #1020, opened 2017-02-17
[TASK] Fix scrutinizer issues in Queue

### LINKED ISSUE #1012: [TASK] Fix scrutinizer issues in Queue

thomashohn:
Should fix #1012. @timohund please have a look at the deprecations


timohund:
I'll have a look at that one later

irnnr:
I'd say if you're not sure whether a method is used, just deprecate it, but do not remove it yet. Someone else might use it.

thomashohn:
I suggest we deprecate both methods with 6.1 then

timohund:
@thomashohn can you remove the todo and do in both a GeneralUtility::logDeprecatedFunction(); i think then we can remove them in 7.0

PR #1065, opened 2017-02-19
[TASK] Fix scrutinizer issues in ReIndexTaskAdditionalFieldProvider

thomashohn:
Fixes: #1064


Classes/Task/ReIndexTaskAdditionalFieldProvider.php
```diff
@@ -111,8 +111,14 @@ public function getAdditionalFields(
         $task,
         SchedulerModuleController $schedulerModule
     ) {
-        $this->initialize($taskInfo, $task, $schedulerModule);
+        if ((!is_null($task)) && (!($task instanceof ReIndexTask))) {
```
timohund: The two fragements look same, can we move them into one method?

thomashohn: I'll look into that :-)

Issue #1077, opened 2017-02-20
[BUG] getIsRootPageId throws exception with 0 as parameter


thomashohn:
The method getIsRootPageId throws an exception when the root page can't be determined. 
The problem is that getRootPageId can return 0. So fix that getRootPageId always returns an integer and an early return in getIsRootPageId if parameter is 0

PR #1078, opened 2017-02-20
[BUG] getIsRootPageId must handle 0 

thomashohn:
The method getRootPageId should always return an int

Fixes:  #1077

### DIFF
--- a/Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php
+++ b/Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php
@@ -134,7 +134,7 @@ class RootPageResolver implements SingletonInterface
     {
         /** @var Rootline $rootLine */
         $rootLine = GeneralUtility::makeInstance(Rootline::class);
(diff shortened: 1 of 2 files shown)


thomashohn:
I'll look into that - or we can look at it Thursday..

timohund:
Yes :)

Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php
```diff
@@ -129,7 +134,7 @@ public function getRootPageId($pageId = 0, $forceFallback = false)
     {
         /** @var Rootline $rootLine */
         $rootLine = GeneralUtility::makeInstance(Rootline::class);
-        $rootPageId = intval($pageId) ? intval($pageId) : $GLOBALS['TSFE']->id;
+        $rootPageId = intval($pageId) ? intval($pageId) : intval($GLOBALS['TSFE']->id);
```
irnnr: I think you can just use tertiary short notation here `?:`

timohund:
@thomashohn Thanks for the pull request!

PR #1080, opened 2017-02-21
[TASK] Fix scrutinizer issues in AdministrationController

thomashohn:
Fixes: #1079


timohund:
@thomashohn thanks for the fix! I think the build was just broken because of an external issue. I've restarted the build.

PR #1082, opened 2017-02-21
[FEATURE] Optimise calls to getConfigurationFromPageId

thomashohn:
So this is an initial version solves the problem of expensive calls to getConfigurationFromPageId in the case where the TS can be reused when the solr root page contains all TS configuration.

I'm open for discussions how this should be implemented or not :-)
Feedback needed @timohund / @irnnr

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -35,6 +35,7 @@ use TYPO3\CMS\Core\TimeTracker\NullTimeTracker;
 use TYPO3\CMS\Core\TypoScript\ExtendedTemplateService;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Core\Utility\ExtensionManagementUtility;
+use TYPO3\CMS\Extbase\SignalSlot\Dispatcher;
 use TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController;
 use TYPO3\CMS\Frontend\Page\PageRepository;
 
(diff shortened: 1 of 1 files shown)


timohund:
In the current implementation we have the following problem:

* On every save operation in the backend the configuration is fetched for this page id. This operation is very expensive and also cache hits are not that likely because the same pages are not edited that often.

* I would be nice when this is faster

As discussed with @thomashohn we think the following implementation makes sence:

When a feature flag in the extension settings is active "useConfigruationFromClosestTemplate" (and only then):

* Resolve the closest page where a typoscript configuration is places on.
* Use this pageId to fetch the configuration and to build the cache.

A feature flag is needed because there are cases (e.g. when TypoScript conditions are used) where you allways want to evaluate the concrete page id.

To implement this we need to:

* Implement a class for the ExtensionConfiguration (System/Configuration/ExtensionConfiguration) with a method "getIsUseConfigruationFromClosestTemplateEnabled".
* Evaluate this flag in Util::getConfigurationFromPageId
* When the flag is active, get the pageId from the closest template and use it to build the configuration and to build the cache key.

Afterwards, when the flag is active, we will get the clostest template, but have more cache hits because the page id to build it will be most time the same one (the one from the closest template in the rootline)

timohund:
@thomashohn i'll close this one and open a new one with the changes that we have discussed yesterday.

Issue #490, opened 2016-07-11
[TASK] Migrate from devLog() to Logging-API


timohund:
devLog() is currently not usable in TYPO3 CMS v6. And in general imho we should use the Logging-API meanwhile.

I guess this task is important since without logging-functionality in some cases it's quite hard to drill down problems with a solr-installation.

timohund:
First step could be to register a devLog listener that writes into a log using the logging framework:

https://daniel-siepmann.de/Posts/2016-05-27-typo3-custom-devlog.html

thomashohn:
Since the ext:solrfal uses the standard TYPO3 logger framework - I would suggest replacing devLog with debug from TYPO3 logger framework. Any objections @irnnr or @timohund ?

irnnr:
sounds good

PR #1086, opened 2017-02-23
[!!!][TASK] Use new logging API instead of deprecated devLog

thomashohn:
This PR introduces a new SolrLogManager which uses the logger framework to write logs.

Since the devLog usage does not make any sence anymore the typoscript setting

plugin.tx_solr.logging.debugDevLogOutput 

was renamed to

plugin.tx_solr.logging.debugOutput

Fixes: #490

### DIFF
--- a/Classes/ConnectionManager.php
+++ b/Classes/ConnectionManager.php
@@ -24,6 +24,7 @@ namespace ApacheSolrForTypo3\Solr;
  *  This copyright notice MUST APPEAR in all copies of the script!
  ***************************************************************/
 
+use ApacheSolrForTypo3\Solr\System\Logging\SolrLogManager;
 use ApacheSolrForTypo3\Solr\System\Page\Rootline;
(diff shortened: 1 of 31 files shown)


timohund:
@thomashohn Thanks for the pull request. Looks good to me. The build currently fails due to changes in the core (independent from this pr) and should be fixed when https://github.com/TYPO3-Solr/ext-solr/pull/1088 is merged.

Regarding the logging i think we should think about the setting plugin.tx_solr.logging.debugDevlogOutput since the devLog is no more used, this setting does not make sence anymore. But the functionallity behind it was quite handy, maybe we change the setting from 'debugDevlogOutput' to 'debugOutput' and move the code to handle this, into the SolrLogManager.

What do you think?

timohund:
@thomashohn i'll merge the stuff and move the logic of the DevLogDebugWriter to the SolrLogManager

PR #1089, opened 2017-02-24
[BUGFIX] Replace `SITE_ROOT` with the actual value as cache identifier

### LINKED ISSUE #488: [BUG] Queue initialization returns wrong (duplicate) results for second site root in multisite environment
A recent project of ours is featuring a multisite/multitree setup, TYPO3 6.2.9 and EXT:solr 3.0. To ease deployment of our platforms we have our own set of CLI scripts, one of those scripts is used to re-initialize the solr index queue for all site roots.

``` php
$availableSites = Tx_Solr_Site::getAvailableSites();

$queue->deleteAllItems();

if (is_array($availableSites)) {
    foreach ($availableSites as $site) {
        if ($site instanceof Tx_Solr_Site) {
            $queue->initialize($site);
        }
    }
}
```

When inspecting the raw index queue (database table tx_solr_indexqueue_item) I noticed some items where added multiple times for the different site roots. This resulted in 404 errors spawned while indexing because of the sys_domain restriction in TYPO3.

After a while I noticed the method \Tx_Solr_IndexQueue_Initializer_Abstract::getPages was calling \Tx_Solr_Site::getPages without any arguments, thus the default 'SITE_ROOT' value of $rootPageId was used. This value was directly used as cache identifier for self::$sitePagesCache[] causing a result set from the "old" site root when initializing the queue for a next (different) site root.

I moved the condition where $rootPageId is compared to 'SITE_ROOT' outside of the condition where is evaluated if a cache entry is available, if $rootPageId is in fact this non unique value I replace it with the uid of the root page of the current site.

https://forge.typo3.org/issues/64957

arnoschoon:
Use the uid of the root page as unique cache identifier to prevent conflicts when calling the getPages() method multiple times from the same script.

Fixes #488

### DIFF
--- a/Classes/Site.php
+++ b/Classes/Site.php
@@ -337,18 +337,19 @@ class Site
      */
     public function getPages($rootPageId = 'SITE_ROOT', $maxDepth = 999)
     {
-        // when we have a cached value, we can return it.
-        if (!empty(self::$sitePagesCache[$rootPageId])) {
-            return self::$sitePagesCache[$rootPageId];
-        }
-
         $pageIds = [];
         $maxDepth = intval($maxDepth);
 
-        $recursionRootPageId = intval($rootPageId);
         if ($rootPageId == 'SITE_ROOT') {
-            $recursionRootPageId = $this->rootPage['uid'];
+            $rootPageId = $this->rootPage['uid'];
             $pageIds[] = (int) $this->rootPage['uid'];
+        }
+
+        $recursionRootPageId = intval($rootPageId);
+
+        // when we have a cached value, we can return it.
+        if (!empty(self::$sitePagesCache[$rootPageId])) {
+            return self::$sitePagesCache[$rootPageId];
         }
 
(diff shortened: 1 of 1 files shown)


timohund:
@arnoschoon Thanks for the patch. Do you have the time to add a regression test for that, that fails when you try to initialize multiple sites?

timohund:
@arnoschoon i will merge this and add an integration test as followup.

arnoschoon:
Thanks for merging @timohund, I'm not sure if I'll be able to write a test for this patch in the coming week. I'll keep an eye open on the issue tracker if I can participate with other issues. I'n currently integrating the extension for a client of ours in a CMS8 environment, so I'll try to provide patches if I encounter issues.

timohund:
@arnoschoon I've added a test in a followup pull request. Thanks for the fix. If you can work on other tasks feel free to work on a task in the issue tracker or from the milestone 6.1.

Issue #937, opened 2017-02-02
[FEATURE] Optimise calls to getConfigurationFromPageId


thomashohn:
We have the following situation:
1) On our root pages we have all our Solr Configuration and site TS config.
2) We often perform copies of sub-trees into new subtrees.

Problem: Performing copies of large sub-trees is very costly due to the fact the the TS-config has to be calculated for each page/sub-pages being copied. Since it takes about 1 second to parse the whole TS (yes we have a big TS config on the root)pages) - the performance is horrible and makes it impossible to perform copies.

We have the following suggestions
1) Either provide some kind of signal to the method getConfigurationFromPageId that can be used to say that the TS should always be taken from the pages rootPages (fastest method)
2) Add a check in getConfigurationFromPageId before calculation the TS, if not in the case that it's cached, if the page being processed has no record in sys_template meaning that the TS config from the parent page can be used from the cache for this page and thus saving a costly calculation.

What say you @irnnr and @timohund ?

timohund:
In the current implementation we have the following problem:

    On every save operation in the backend the configuration is fetched for this page id. This operation is very expensive and also cache hits are not that likely because the same pages are not edited that often.

    I would be nice when this is faster

As discussed with @thomashohn we think the following implementation makes sence:

When a feature flag in the extension settings is active "useConfigruationFromClosestTemplate" (and only then):

    Resolve the closest page where a typoscript configuration is places on.
    Use this pageId to fetch the configuration and to build the cache.

A feature flag is needed because there are cases (e.g. when TypoScript conditions are used) where you allways want to evaluate the concrete page id.

To implement this we need to:

    Implement a class for the ExtensionConfiguration (System/Configuration/ExtensionConfiguration) with a method "getIsUseConfigruationFromClosestTemplateEnabled".
    Evaluate this flag in Util::getConfigurationFromPageId
    When the flag is active, get the pageId from the closest template and use it to build the configuration and to build the cache key.

Afterwards, when the flag is active, we will get the clostest template, but have more cache hits because the page id to build it will be most time the same one (the one from the closest template in the rootline)

PR #1092, opened 2017-02-26
[TASK] Allow to use closest typoscript template to retrieve configuration

timohund:
By now, when the configuration is fetched the real pageId is used to retrieve it.
The drawback of this approach is, that in many systems this is an overhead,
because the configuration is only set on the rootlevel or some subpages
(except when e.g. conditions are used).

This pull request introduces an ext_conf_template setting that forces ext_solr
to use the closest page id with a typoscript configuration to retrieve the
configuration. The result is, that the configuration cache has more hits and
the performance in the backend is better

Fixes: #937

### DIFF
--- a/Classes/System/Configuration/ConfigurationPageResolver.php
+++ b/Classes/System/Configuration/ConfigurationPageResolver.php
@@ -1,0 +1,137 @@
+<?php
+namespace ApacheSolrForTypo3\Solr\System\Configuration;
+
+/***************************************************************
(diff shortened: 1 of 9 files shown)


timohund:
@thomashohn can you maybe check on your system how this behaves when this setting is active?

timohund:
@thomashohn I've fixed the comments you've mentioned in the review and added the runtime cache.

PR #1119, opened 2017-03-02
[TASK] FAQ from Slack-Channel Questions for Feb. 2017

dkd-kaehm:
Added: In which cases do I want to trigger indexing manually?

### DIFF
--- a/Documentation/FAQ/Index.rst
+++ b/Documentation/FAQ/Index.rst
@@ -43,8 +43,8 @@ Did you configure your solr connection as required?
 
 **In which cases do I want to trigger indexing manually?**
 
-- by changing any configuration file.
-- by modifying synonyms, stop words, protected words in BE>Search
+- after changing any configuration file.
+- after modifying synonyms, stop words, protected words in TYPO3 Backend -> Search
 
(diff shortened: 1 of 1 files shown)


Documentation/FAQ/Index.rst
```diff
@@ -41,6 +41,16 @@ Did you configure your solr connection as required?
 
 |
 
+**In which cases do I want to trigger indexing manually?**
+
+- after changing any configuration file.
+- after modifying synonyms, stop words, protected words in TYPO3 Backend -> Search
+
+Moreover by changing core/index configuration you need to reload the core to make the changes become active.
+To reload configuration you can either restart the whole Solr server or simply reload a specific core.
```
irnnr: I believe we added automatic core reload after editing synonyms, stop words, protected words. You might want to double check though. You can also point to the button in the backend module.

timohund: I think currently the reload is happening automatically when editing synonyms or stopwords, but the part above regarding synonyms and stopwords is just talking about reIndexing. At least for synonyms a reindexing is still required.

Issue #1115, opened 2017-03-02
[FEATURE] Add possibility to exclude Tables


thomashohn:
On complex installations with a lot of custom tables that are related to pages and tt_content the hook processDatamap_afterDatabaseOperations sends a lot of notifications for tables that you might never want to index. 

With the new Extension configuration it would be nice to have a setting called useConfigurationNeverIndexCustomtables which contains a list of tables that can be used for an early return before the existing check $this->hasRecordBeenProcessed

timohund:
I would say for this a configuration in ext_conf_template.txt "limitToTables" makes sence, when nothing is configured all tables get tracked, when a list ist set it can be used.

E.g like that:
```php
$extensionConfiguration->getIsAllowedTable($tablename);
```

Which allways returns true when nothing is configured and false, when something is configured and the tablename does not match.

thomashohn:
Makes sense 👍

irnnr:
I'd propose to call the option `monitorTables` to go with the Record Monitor name. What do you think?

timohund:
Sound more consistent

PR #1123, opened 2017-03-03
[FEATURE] Added extension configuration for MonitorTables

thomashohn:
Added useConfigurationMonitorTables

Fixes: #1115

### DIFF
--- a/Classes/System/Configuration/ExtensionConfiguration.php
+++ b/Classes/System/Configuration/ExtensionConfiguration.php
@@ -24,6 +24,8 @@ namespace ApacheSolrForTypo3\Solr\System\Configuration;
  *  This copyright notice MUST APPEAR in all copies of the script!
  ***************************************************************/
 
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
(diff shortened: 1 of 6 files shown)


Classes/System/Configuration/ExtensionConfiguration.php
```diff
@@ -73,6 +75,28 @@ public function getIsUseConfigurationTrackRecordsOutsideSiteroot()
...
+    public function getIsUseConfigurationMonitorTables()
+    {
+        $monitorTables = [];
+        $monitorTablesList = $this->getConfigurationOrDefaultValue('useConfigurationMonitorTables', '');
+
+        if (empty($monitorTablesList)) {
+            return $monitorTables;
+        }
+
+        $monitorTablesTemp = GeneralUtility::trimExplode(',', $monitorTablesList);
```
frans-beech-it: is `array_flip()` not easier here to just get the values as keys?

timohund: I guess you are right

PR #1133, opened 2017-03-07
[TASK] Allow self signed certificates when using https

tomasnorre:
This PR:

* implements the usage of self signed ssl certificates for indexing

Fixes: #1134


timohund:
@tomasnore Thx for the pull request

tomasnorre:
You're welcome.

verdure-peuser:
Since there is an option in core to globally set defaults for the used http connections it would be nice if those values would be respected. so it would be possible to enable those self-signed certs only in development context and not in production.

For example: $GLOBALS['TYPO3_CONF_VARS']['HTTP']['ssl_verify_peer']

PR #1139, opened 2017-03-07
Remove check when overwriting type

spoonerWeb:
Due to having no clue why this is forbidden the check should be removed.

### DIFF
--- a/Classes/IndexQueue/AbstractIndexer.php
+++ b/Classes/IndexQueue/AbstractIndexer.php
@@ -66,13 +66,6 @@ abstract class AbstractIndexer
                 continue;
             }
 
-            if ($solrFieldName == 'type') {
-                throw new InvalidFieldNameException(
-                    'Must not overwrite field "type".',
-                    1435441863
-                );
-            }
-
             $fieldValue = $this->resolveFieldValue($indexingConfiguration,
                 $solrFieldName, $data);
 


timohund:
@irnnr Can you maybe explain the thoughts behind the check. Do you think it should be removed?

irnnr:
Yeah, I don't want this to be changed. 

The reason this was put in place was that it caused a lot of issues with users overwriting the type field. It is used in a number of Solr and SQL queries that depend on the correct values to ensure consistency between the data in TYPO3's database and Sole's index.

If you need custom types, you can create a dynamic field `type_stringS`.

I'm trying to find the original issue.

irnnr:
original commit: https://github.com/TYPO3-Solr/ext-solr/commit/d875657b6645c208dfbac0bb4a0b827b950a3894

irnnr:
related: https://forge.typo3.org/issues/61364

Hans Höchtl:  
> The field 'type' as a special one as it's necessary for managing relation between solr-document and database table.
You could simply use a different dynamic-field for your purpose like mytype_stringS and copy the facet typoscript and change your template markers.

```
plugin.tx_solr {
    index{
        queue{
            news = 1
            news {
                table = tt_news

                fields {
                    mytype_stringS = TEXT
                    mytype_stringS.value = news

                }
            }
        }
    }
    search.faceting.facets.mytype_stringS {
         label = Type
         field = mytype_stringS
    }
}
```

irnnr:
I was under the impression we had a FAQ for this since it comes up every time and again, but unfortunately I couldn't find anything there either.

I added an issue to add a FAQ entry: #1172

Issue #1130, opened 2017-03-06
[TASK] Move logic from Item::updateIndexedTime to IndexQueue


timohund:
The logic to update the index time in the database is currently located int the Item class. Since the database access should be centralized in the IndexQueue it would make sence to move this logic to the indexQueue

Todo:

* Add a method updateIndexTimeByItem(Item) in Queue
* Let Item::updateIndexedTime use this new method and deprecated it (to keep backword compatibility)
* Change all places where $item->updateIndexedTime() to use the queue instead

PR #1140, opened 2017-03-07
[TASK] Move updateIndexedTime to Queue and rename

thomashohn:
Fixes: #1130

### DIFF
--- a/Classes/IndexQueue/Queue.php
+++ b/Classes/IndexQueue/Queue.php
@@ -1034,7 +1034,7 @@ class Queue
      *
      * @param Item $item
      */
-    public function updateIndexQueueByItem(Item $item)
+    public function updateIndexTimeByItem(Item $item)
     {
         $GLOBALS['TYPO3_DB']->exec_UPDATEquery(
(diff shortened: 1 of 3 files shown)


Classes/IndexQueue/Queue.php
```diff
@@ -1028,4 +1028,18 @@ public function markItemAsFailed($item, $errorMessage = '')
             ]
         );
     }
+
+    /**
+     * Sets the timestamp of when an item last has been indexed.
+     *
+     * @param Item $item
+     */
+    public function updateIndexQueueByItem(Item $item)
```
timohund: Hi thomas, i would propose to indicate in the name, that only the indextime is updated, otherwise somebody might think the whole items is getting updated. I would propose something like "updateIndexTimeByItem(Item $item)"

thomashohn: :-)

timohund:
@thomashohn Thanks for the pull request. Looks good to me, i'll merge it when the build is finished.

PR #1151, opened 2017-03-09
[TASK] Split Site

thomashohn:
This is currently WIP - first initial version wrapping stuff. 
Tests etc. must be adjusted etc.


timohund:
Looks good to me, i have one additional proposal:

* Maybe we add a method to the repository getSiteByRootPageId. Then we can also replace all direct instanciations of Site in the code (that SiteRepository is the only place where the Site object is created).

I think later we can think about to fetch the whole data(languages, label, ... in the repository) but since this would be a really big change and we need to measure the performance differences i think this could be done in a step later.

Classes/Domain/Site/SiteHashService.php
```diff
@@ -115,7 +116,8 @@ protected function getDomainByPageIdAndReplaceMarkers($pageId, $allowedSitesConf
      */
     protected function getAvailableSites()
     {
-        return Site::getAvailableSites();
+        $siteRepository = GeneralUtility::makeInstance(SiteRepository::class);
```
timohund: I would propose to make the siteRepository a constructor dependecy and use it directly. I think getAvailableSites abd getSiteByPageId have only been introduced as proctected methods because the old static way Site::getAvailableSites and Site::getSiteByPageId was not mockable in the test (because of the static).

timohund:
@thomashohn Thanks for the pull request, as discussed on slack i will add the tests for the SiteRepository as a followup.

PR #1159, opened 2017-03-13
[FEATURE] Force commit after indexing documents

### LINKED ISSUE #1158: [FEATURE] Force commit after indexing
A bit of a trade off after reading #870 where the point is to avoid committing. I would like to see an option to execute a commit after running the `IndexQueueWorker`.

In a very early release of EXT:solr there was a scheduler task for this, but the `autoCommit` feature of Solr is a better alternative, not even because a commit should only take place if things have changed and not periodically.

In a platform where we have replication set up between multiple Solr servers the client demands that even small changes are visible almost immediately. For that purpose I've created a small patch https://github.com/arnoschoon/ext-solr/commit/d6589551a5ee2d2039f4d1c4301dc6600adb5065 that adds an option to the task configuration and issues a `commit` after indexing documents.

I'm aware of the fact that this could seriously harm performance, but for these "large" setups we host Solr on dedicated machines for each client (private networks). I'm not sure yet if this option should be added as a checkbox in the back-end interface or (like #875) as a directive in TS. The `autoCommit` feature https://github.com/TYPO3-Solr/ext-solr/blob/master/Resources/Private/Solr/configsets/ext_solr_6_1_0/conf/solrconfig.xml#L94 is also a viable option, but what would be sensible values?

Any thoughts, ideas and criticism are welcome.

arnoschoon:
If a batch of documents is processed by the index queue worker send an explicit commit command to the Solr servers for that site.

Fixes #1158

### DIFF
--- a/Classes/Domain/Index/IndexService.php
+++ b/Classes/Domain/Index/IndexService.php
@@ -25,6 +25,7 @@ namespace ApacheSolrForTypo3\Solr\Domain\Index;
  *  This copyright notice MUST APPEAR in all copies of the script!
  ***************************************************************/
 
+use ApacheSolrForTypo3\Solr\ConnectionManager;
 use ApacheSolrForTypo3\Solr\IndexQueue\Indexer;
 use ApacheSolrForTypo3\Solr\IndexQueue\Item;
 use ApacheSolrForTypo3\Solr\IndexQueue\Queue;
@@ -114,6 +115,7 @@ class IndexService
         $errors     = 0;
         $indexRunId = uniqid();
         $configurationToUse = $this->site->getSolrConfiguration();
+        $enableCommitsSetting = $configurationToUse->getEnableCommits();
 
         // get items to index
         $itemsToIndex = $this->indexQueue->getItemsToIndex($this->site, $limit);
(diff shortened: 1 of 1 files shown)


timohund:
Sounds good for me. @thomashohn what do you think? Since you've propose the enableCommits setting.

RasmusWernerLarsen:
Yes, I think this makes sense, and has a reasonable usecase.

- The usecase for enableCommits=0 is to control commits for solr cores that are used by multiple installations, so you can do your own performance/searchability delay tradeoffs.
- The usecase here is, I guess, to give Typo3 full control over commit times, to minimize the time it takes for content to be searchable.

Only issue I see is that this might cause some performance regressions for people who upgrade, but do not consider/notice this change. Still, for anyone using just 1 installation per solr core, this shouldn't be a real issue, and I guess they are the vast majority.

timohund:
@RasmusWernerLarsen I think people who upgrade can still then set enableCommits = 0 and maybe adapt the solr configuration. I would propose to merge it and add a note to the release notes. I guess for 95% of the installations nobody will recognize a difference and people who run a system with much sites should read the releasenotes anyhow.

RasmusWernerLarsen:
Agreed, I see no issues with merging this.

timohund:
@arnoschoon @RasmusWernerLarsen Thanks the pull request and your thoughts

Issue #1177, opened 2017-03-17
[TASK] Refactor getIsUseConfigurationMonitorTables


thomashohn:
Refactor getIsUseConfigurationMonitorTables

PR #1178, opened 2017-03-18
[TASK] Use array_flip getIsUseConfigurationMonitorTables

thomashohn:
Fixes: #1177

### DIFF
--- a/Tests/Unit/System/Configuration/ExtensionConfigurationTest.php
+++ b/Tests/Unit/System/Configuration/ExtensionConfigurationTest.php
@@ -74,8 +74,8 @@ class ExtensionConfigurationTest extends UnitTest
             ['useConfigurationMonitorTables' => 'pages, tt_content']
         );
         $expectedResult = [
-            'pages' => true,
-            'tt_content' => true
+            'pages' => 0,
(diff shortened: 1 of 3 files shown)


Tests/Unit/System/Configuration/ExtensionConfigurationTest.php
```diff
@@ -74,8 +74,8 @@ public function testIsGetIsUseConfigurationMonitorTablesConfigured()
             ['useConfigurationMonitorTables' => 'pages, tt_content']
         );
         $expectedResult = [
-            'pages' => true,
-            'tt_content' => true
+            'pages' => 0,
```
timohund: Is this really a wanted change? Becuase pages should also be 1 from my thinking and also we change the values from boolean to int?

thomashohn: Maybe I just misunderstood the intention of this https://github.com/TYPO3-Solr/ext-solr/pull/1123#discussion_r105605392? But you can discard the PR if you mean it's not relevant.

timohund:
@thomashohn Thanks for the pull request and the adaptions!

Issue #1127, opened 2017-03-06
[TASK] Add constants for new username and password typoscript setting


timohund:
We should add a constant for the settings "usename" and "password" and beside that re-check if the connections gets initialized properly when the constant is set to an empty string.

PR #1182, opened 2017-03-22
[TASK] Add constant for username and password [6.0.x]

timohund:
This pr:

* Adds a constant for the username and password
* Adds a test if no authentication is done, when username is empty

Fixes: #1127


dkd-friedrich:
LGTM

Issue #1183, opened 2017-03-22
"Only variables should be assigned by reference" in IndexQueue\Indexer::preAddModifyDocuments()


LeoniePhiline:
Depending on your settings, this fails loudly or silently:

https://github.com/TYPO3-Solr/ext-solr/blob/master/Classes/IndexQueue/Indexer.php#L474

`$documentsModifier = &GeneralUtility::getUserObj($classReference);`

Fails with:

> "PHP Runtime Notice: Only variables should be assigned by reference in /srv/ext/system/www/trefwoord76.development.netcreators.info/typo3conf/ext/solr/Classes/IndexQueue/Indexer.php line 468"

(Line 468 in solr 6.0.0 - in the current master this is line 474.)

Please remove the ampersand sign. :)

PR #1192, opened 2017-03-24
[!!!][TASK] Fix assignment by reference in Indexer plus test

thomashohn:
Fixes: #1183


thomashohn:
For the Release notes a a not that the method getAdditionalDocuments now uses makeInstance instead of calling the getUserObj method. 
This could have an impact on installations where the configuration is done via file:reference - which is considered a bad practice.

Issue #1205, opened 2017-03-28
[BUG] Synonyms with umlauts behind frontend proxies


dkd-kaehm:
Behind frontend proxies with default settings it's impossible to delete
synonyms.

PR #1207, opened 2017-03-28
[FIX] Deleting synonyms with umlauts behind fronted proxies

dkd-kaehm:
We need to encode the url for deleting synonyms on Apache Solr
machines behind front-end proxies with default configs.

Closes: #1205


timohund:
@dkd-kaehm The pr conflicts because of the previous pr. Can you please rebase this pr?

PR #1213, opened 2017-03-29
[BUGFIX] GarbageCollector fails to check endtime correctly

### LINKED ISSUE #1212: GarbageCollector fails to check endtime correctly
If a record with configured endtime column is changed, the garbage collector is currently (EXT:solr 6.0 / master) removing it from the index, as it isn't checked if the endtime is actually set (`endtime > 0`). Thus all record changes lead to a removal from the index, but direct deletions should only be performed if a record is marked as hidden or gets deleted.

An additional check for endtime > 0 in `isEndTimeInPast()` should solve this issue, see following if statement, which should be improved:

```
if (isset($GLOBALS['TCA'][$table]['ctrl']['enablecolumns']['endtime'])) {
    $endTimeField = $GLOBALS['TCA'][$table]['ctrl']['enablecolumns']['endtime'];
    $endTimeInPast = $record[$endTimeField] < time();
}
```

markuskobligk:
fixes #1212

### DIFF
--- a/Classes/System/TCA/TCAService.php
+++ b/Classes/System/TCA/TCAService.php
@@ -103,7 +103,8 @@ class TCAService
 
         if (isset($this->tca[$table]['ctrl']['enablecolumns']['endtime'])) {
             $endTimeField = $this->tca[$table]['ctrl']['enablecolumns']['endtime'];
-            $endTimeInPast = $record[$endTimeField] < $this->getTime();
+            if($record[$endTimeField] > 0)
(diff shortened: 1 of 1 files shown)


markuskobligk:
Sorry for having three commits on this one, somehow my editor was "confused" about PSR-2 settings :-(

frans-beech-it:
Tnx for the fix. Just tested and it works as expected :+1:

timohund:
@markuskobligk @frans-beech-it thx for the support.

Issue #1215, opened 2017-03-30
[BUG] Newest TYPO3 8 Master breaks Initialize Solr connections in Cache Menu


thomashohn:
The newest TYPO3 8 breaks the "Initialize Solr connections" in the menu "Clear-cache"

Issue #950, opened 2017-02-09
[TASK] Remove 7.6 backwards compatibility


timohund:
* Remove from ext_emconf / composer.json
* Remove fix from here: https://github.com/TYPO3-Solr/ext-solr/pull/949/commits/0129e857a41280d750152fcf23c22ecd0c13d6fd#diff-0dc5ba2d5b4153529fdcedb0c2c0cf95R118
 See also: #949 
* Remove 7.6 compatibility from https://github.com/TYPO3-Solr/ext-solr/issues/1215
* Remove patching of testing framework

PR #1216, opened 2017-03-30
[BUG] Fix Initialize Solr Connection again

thomashohn:
Fixes: #1215

### DIFF
--- a/Classes/ConnectionManager.php
+++ b/Classes/ConnectionManager.php
@@ -35,6 +35,7 @@ use TYPO3\CMS\Core\Registry;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\TypoScript\ExtendedTemplateService;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\Utility\VersionNumberUtility;
 use TYPO3\CMS\Frontend\Page\PageRepository;
 
 /**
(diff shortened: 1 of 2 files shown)


timohund:
@thomashohn Can you please add a todo to remove the 7.6 stuff in EXT:solr 7.0.0

thomashohn:
Added todo

dkd-kaehm:
on CMS@dev-master Translation does not do the thing:

bildschirmfoto 2017-03-30 um 17 34 28

using:
`\TYPO3\CMS\Extbase\Utility\LocalizationUtility::translate('cache_initialize_solr_connections', 'solr')`
solves that and works in 7.6 to ;) .

of course all caches flushed incl. opcache and reloaded the backend.

thomashohn:
Weird - works fine at my end.
My suggestion would be to leave the code as it is - since

- 7.6 LTS support will soon be dropped.

- It has "always" been hardcoded.

- It's the second time I have to "fix" this with TYPO 8 - so I would rather be conform with the core and only pass the language reference.

This shows my installation - the translation of the label seems to work without any problems
screen shot 2017-03-31 at 06 07 50

timohund:
I've tested thomas patch and it works on both systems. @dkd-kaehm could it be that your TYPO3 dev-master was not uptodate? I would propose to merge it and remove the 7.6 compatibility right after the release.

timohund:
Added note to #950 for the removal in the next release. Thx @thomashohn & @dkd-kaehm

Issue #1217, opened 2017-03-31
[TASK] Remove method hasRecordBeenProcessed in RecordMonitor


thomashohn:
This method was introduced to avoid additional extra expensive executions of processDatamap_afterDatabaseOperations due to a bug in the Core. 
Since the bugfix made it into both TYPO3 v8 and TYPO3 v7 - this method and it's use should be removed before releasing the next Solr version.
See

- [Bug report on forge.typo3.org](https://forge.typo3.org/issues/79635)

- [Patch on review.typo3.org](https://review.typo3.org/#/c/51552/)

PR #1218, opened 2017-03-31
[TASK] Remove method hasRecordBeenProcessed in RecordMonitor

thomashohn:
The original cause has been solved in TYPO3 8 and backportet to TYPO3 7.

Fixes: #1217


timohund:
I would propose to wait until monday and when nothing get's reverted in the core to merge this. @thomashohn what do you think? Thanks for the pull request.

thomashohn:
I agree :) Could also be thrown in just before releasing Solr..

PR #1240, opened 2017-04-05
[BUGFIX] Move $TCA['tt_content'] from ext_tables.php to TCA/Overrides

frans-beech-it:


frans-beech-it:
this is needed to be compatible with https://github.com/FriendsOfTYPO3/compatibility6 will close this PR and create a new one that removes this legacy code for master.

PR #1260, opened 2017-04-07
Indexing records within protected area

devmes:
In the new release the extension does not index records within the protected fe_user area anymore.


devmes:
sorry, this merge request should point to the branch release-6.1.x

PR #1273, opened 2017-04-13
[BUGFIX] Load default configuration for DB records outside site root

hirnschmalz:
If a custom DB record is outside of the site root and should be
indexed (via `additioanlpageIds` configuration) the configuration
of the site root is loaded if there is no configuration present in
the rootline of the cusomt DB record.


timohund:
@hirnschmalz Thanks for the patch. I've tested your approach and found another problem when generating the documentId. I've created a new pr (#1299) that also contains you patch. Can you please check if this fixes the problem in your environment as well? I'll close this pr for now.

Issue #1277, opened 2017-04-25
solrfal: The page for the given page ID '-1' could not be found in the database and can therefore not be used as site root page.


thokel:
Hi,

when saving changes to content elements in a draft workspace solrfal throws an exception (see below). Changes are saved to the database though. Editing content elements in LIVE workspace is working fine.

**Expected behavior**
Saving changes to content elements in a draft workspace does not throw an exception.

**Stack trace**
```
Uncaught TYPO3 Exception
#1487171426: The page for the given page ID '-1' could not be found in the database and can therefore not be used as site root page. (More information)
InvalidArgumentException thrown in file
.../typo3conf/ext/solr/Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php in line 121.

16 ApacheSolrForTypo3\Solr\Domain\Index\Queue\RecordMonitor\Helper\RootPageResolver::getIsRootPageId(-1)
.../typo3conf/ext/solrfal/Classes/Queue/ConsistencyAspect.php:
00280:         if ($rootPageResolver->getIsRootPageId($rootPageId)) {

15 ApacheSolrForTypo3\Solrfal\Queue\ConsistencyAspect::getDetectorsForPageOrContentElement("tt_content", 46524)
.../typo3conf/ext/solrfal/Classes/Queue/ConsistencyAspect.php:
00258:             return $this->getDetectorsForPageOrContentElement($table, $uid);

14 ApacheSolrForTypo3\Solrfal\Queue\ConsistencyAspect::getDetectorsForRecord("tt_content", 46524)
.../typo3conf/ext/solrfal/Classes/Queue/ConsistencyAspect.php:
00215:         $detectors = $this->getDetectorsForRecord($table, $uid);

13 ApacheSolrForTypo3\Solrfal\Queue\ConsistencyAspect::issueCommandOnDetectors("recordUpdated", "tt_content", 46524)
.../typo3conf/ext/solrfal/Classes/Queue/ConsistencyAspect.php:
00177:             $this->issueCommandOnDetectors($method, $table, $id);

...
```

**We are using the following setup**
- TYPO3 7.6.16
- Solr 6.1.0 
- Solr FAL 4.1.0.
- PHP 7.0.7

**Steps how to reproduce the issue**
- Install solr + solrfal
- Switch to a workspace which is not LIVE workspace
- Edit any content element
- Save your changes

timohund:
We should port the fix to master as well

[...]

PR #1290, opened 2017-05-24
#1277 ignore workspace

alappe:
Can this be a simple fix for #1277? At least the backend isn't breaking anymore…


dkd-kaehm:
please rebase to 6.1

timohund:
and we should add a testcase for this @alappe can you do this?

alappe:
I can, I will approach it next week...

PR #1297, opened 2017-05-29
[BUGFIX] Ignore workspaces in RootPageResolver

alappe:
Related: #1277


timohund:
We should port this to dev-master as well

Issue #1282, opened 2017-05-04
[TASK] Refactor Backend Modules


dkd-kaehm:
- [x] Move Source-Files to the right from Extbase suggested places. #1269
- [x] Implement Backend Components for usage in Modules.
- [x] Split/Merge and move the modules to the module menu.
- [ ] adapt EXT:Tika ...

Please feel free to add the tasks

PR #1300, opened 2017-05-29
[TASK] Refactor backend modules

### LINKED ISSUE #1234: [BUG] Solr Root Page in user configuration is not cleared after reinitialization of Solr connections
## Preconditions

TYPO3 version: 8.7.0-dev, 6.2.25

Given an installed ext-solr extension with a configured and working connection for TYPO3 domain "Domain1" (Root Page ID = 1)

## How to reproduce

1) Copy TYPO3 domain "Domain1" and call it "Domain 2" (Let's say the new Page ID = 38)
2) Disable Solr on Domain 1 with 
```
plugin.tx_solr.enabled = 0
```
3) Reinitialize Solr Connections in toolbar menu

If you now change in Search module, you will see, that there is only one active connection to Domain 2 available (that's so far correct), but solr is still trying to establish a connection to Domain 1.

broken_solr_connections

## Initial analysis

Solr extension saves the current module data as a serialized object of `ApacheSolrForTypo3\Solr\Domain\Model\ModuleData` in backend user configuration (table `be_users` -> `uc`). Among other things, it contains the Solr Root Page`ApacheSolrForTypo3\Solr\Site`, the current user has successfully visited in search module for the last time. 

When search module is opened after the reinitialization of Solr connections again, the last active Solr Root Page is read out from the user configuration, regardless whether the corresponding Solr configuration and connection still exist.

## Expected behavior

One of the possible solutions would be to clear the stored Solr Root Page in module configuration for each backend user after reinitialization of Solr connections.

dkd-kaehm:
- Created new Main Module „Apache Solr“ as „searchbackend“ mainModuleName
- Moved Synonyms and Stop-Words in CoreOptimization
- Moved StopWordsModuleController and SynonymsModuleController in CoreOptimizationModuleController
  using Backend components.
- Moved IndexQueue Module
- Moved IndexMaintenance Module to Administration Module
- Moved Empty Index Button to Administration Module
- Moved Overview-, SearchStatistics-, IndexFields modules to InfoModuleController
- Deleted incl. templates:
  - AdministrationController
  - AdministrationModuleManager
  - AdministrationModuleInterface
  - ApacheSolrForTypo3\Solr\Backend\SolrModule\AbstractModuleController

Todo: set flag to allow using the choosed Page-UID between switching modules

Fixes: #1282, closes #1234


Classes/Controller/Backend/Search/IndexQueueModuleController.php
```diff
@@ -147,21 +167,10 @@ public function initializeIndexQueueAction()
                 FlashMessage::OK
             );
 
-            $this->addFlashMessagesByQueueIdentifier('solr.queue.initializer');
+            //$this->addFlashMessagesByQueueIdentifier('solr.queue.initializer');
```
timohund: Why is this commented out? Do we need it or can we remove it?

timohund: I think in addition we should think about how to avoid the traits, e.g. by moving the code into a shared abstract class or a service that can be used as object. Technically traits are the same as copy & paste.

I think with merging this we have a better useable and implemented backend!

timohund: I've tested the module with TYPO3 8.7 LTS. Looks good. One thing i think we should optimize is to keep the selected page/site when the module is changed. Is that possible?

Issue #1179, opened 2017-03-21
[TASK] Integrate EXT:solrfluid into EXT:solr


timohund:
Move all Classes/Tests and Documentation:

* Check that all tests are green
* Merge the documentation
* Make Search and SearchResultSet work as non singletons
* Delete the old Templatingcode / Templates and ViewHelper

PR #1308, opened 2017-06-01
[!!!][TASK] Replace templating with fluid templating (merge solrfluid)

timohund:
This PR:

* Adds the fluid templating (from solrfluid)
	* Solrfluid ViewHelpers are available in Classes/ViewHelpers/Frontend/ (which requires some changes in the templates)
* Drops the old templating
* Uses FluidStandalone when possible

How to migrate your project from EXT:solrfluid to EXT:solr 7.0.0:

* Make sure that resultSet is passed to every partial (was not the case for all partials before)
* Change the root namespace in your templates to 'xmlns:s="http://typo3.org/ns/ApacheSolrForTypo3/Solr/ViewHelpers/Frontend"' (before xmlns:s="http://typo3.org/ns/ApacheSolrForTypo3/Solrfluid/ViewHelpers")
* Move all Widget template from <yourTemplateFolder>/ViewHelpers/Widget/* to <yourTemplateFolder>/ViewHelpers/Frontend/Widget/
* Include "Search - Default Stylesheets" when you have used the default css before

Fixes: #1179


dkd-kaehm:
@timohund is [composers replace](https://getcomposer.org/doc/04-schema.md#replace) supported by CMS 8 LTS currently?
If yes, don't forget to add solrfluid in this section in composer.json

Resources/Private/Templates/Frontend/Search/Form.html
```diff
@@ -0,0 +1,9 @@
+<html xmlns="http://www.w3.org/1999/xhtml" lang="en"
+	  xmlns:f="http://typo3.org/ns/TYPO3/Fluid/ViewHelpers"
+	  xmlns:s="http://typo3.org/ns/ApacheSolrForTypo3/Solr/ViewHelpers/Frontend/">
+
+<f:layout name="Frontend/Main"/>
+
+<f:section name="main">
+	<f:render partial="Frontend/Search/Form" section="Form" arguments="{search:search, additionalFilters:additionalFilters}" />
```
dkd-kaehm: I think this will not work by "Use custom Fluid Templates", if user sets
`partialRootPaths.10 = EXT:own_extension/Resources/Private/Partials/Frontend/`
I prefer `<f:render partial="Search/Form" section="Form" arguments="{search:search, additionalFilters:additionalFilters}" />`
and using `.../Frontend/` everywhere in TypoScript configs

PR #1341, opened 2017-06-20
Fix logging of error when devlog is enabled

tribut:
`getTrace()` returns references to the functions called which is a problem, because devlog calls `serialize()` on the data, leading to a `Serialization of 'Closure' is not allowed` exception.

With the typo3 logging system (so tx_solr 6.1+) this is apparently not a problem, because it uses `json_encode` which just produces `{}` for closures.

Fixes: #1342


timohund:
@tribut Can you prepare a pull request for master as well?

tribut:
Sure, see #1343.

PR #1373, opened 2017-06-23
[TASK] allow 'hide default translation of page' on root page

maritwho:
If the root page is set to 'hide default translation of page' this causes a 404 in TypoScriptFrontendController with initializing the TSFE. To avoid this a language uid has to be thrown to the initialize function.

Fixes: #1374


Classes/IndexQueue/Indexer.php
```diff
@@ -554,7 +555,17 @@ protected function getSolrConnectionsByItem(Item $item)
             $siteLanguages[] = $solrConfiguration['language'];
         }
 
-        $translationOverlays = $this->getTranslationOverlaysForPage($pageId, $site->getSysLanguageMode());
+        $defaultLanguageUid = 0;
```
timohund: I would propose to move this into a method getDefaultLanguageUid()

maritwho: Ok, I will change this and update my pull request.

maritwho:
So now I updated my pull request. There is another change in line 571 I forgot in my first push. I implemented now a function getDefaultLanguageUid.

I will try to add a test for this in the next view days.

timohund:
Thanks for the optimizations of the patch. I think when we have a test for the use case we can merge it and should also port the fix to dev-master as well.

timohund:
@maritwho Are you still working on this? For me it looks like that your fixture needs to be located in  Tests/Integration/IndexQueue/Fixtures, not  Tests/Integration/Fixtures ?

maritwho:
@timohund Yes I am still working on this. I didn't have enough time to fix this. Thank you for your hint!

timohund:
@maritwho is there any update on this? Do you think you will have the time to continue on that?

maritwho:
I tried something in a separate branch on  my fork and travis. I was really busy the last weeks, so I had no time. But may be we can have a short talk at the T3DD17 in Malmö, so that I can finish this?

timohund:
@maritwho sure good idea, maybe in the coding night?

maritwho:
@timohund This pull request can be deleted. It's done in #1409

PR #1386, opened 2017-07-03
[BUGFIX] Missing parameter type in FrequentSearches::postInitializeTemplateEngine

pschriner:
Missing parameter type leads to warning in PHP7

### DIFF
--- a/Classes/Plugin/FrequentSearches/FrequentSearches.php
+++ b/Classes/Plugin/FrequentSearches/FrequentSearches.php
@@ -83,7 +83,7 @@ class FrequentSearches extends CommandPluginBase
      * @param \ApacheSolrForTypo3\Solr\Template $template
      * @return \ApacheSolrForTypo3\Solr\Template
      */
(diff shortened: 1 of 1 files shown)


Classes/Plugin/FrequentSearches/FrequentSearches.php
```diff
@@ -83,7 +83,7 @@ protected function performAction()
      * @param \ApacheSolrForTypo3\Solr\Template $template
      * @return \ApacheSolrForTypo3\Solr\Template
      */
-    protected function postInitializeTemplateEngine($template)
+    protected function postInitializeTemplateEngine(\ApacheSolrForTypo3\Solr\Template $template)
```
dkd-kaehm: please import FQDN in use as its done in https://github.com/TYPO3-Solr/ext-solr/blob/release-6.1.x/Classes/Plugin/PluginBase.php#L37
`protected function postInitializeTemplateEngine(Template $template)`

pschriner:
Thanks for the quick response

PR #1388, opened 2017-07-05
[FEATURE] Explicitly create data folder before fixing permissions

### LINKED ISSUE #1278: [FEATURE] Enable zero-configuration use of Docker image
When trying to use the Docker image "out of the box", i.e. without manually adding configuration / creating the volume (https://github.com/TYPO3-Solr/ext-solr/blob/master/Documentation/GettingStarted/Solr.rst#docker), you end up with permission problems on the `data/` directory (`/opt/solr/server/solr/data`).

This is due to the `VOLUME` directive inside the `Dockerfile` (https://github.com/TYPO3-Solr/ext-solr/blob/master/Dockerfile#L15) creating the folder with `root` user / group permissions. This is avoided by the steps detailed in the Getting Started. Unfortunately this means, for example, usage inside `docker-compose` is not straightforward:

```
version: "3"
services:
    solr:
      image: typo3solr/ext-solr:latest
      ports:
        - 8983:8983
      volumes:
        - storage-solr:/opt/solr/server/solr/data
volumes:
  storage-solr:
```

This, right now, results in SOLR running into permission problems creating the indexes on first start:
`Caused by: org.apache.solr.common.SolrException: java.nio.file.AccessDeniedException: /opt/solr/server/solr/data/core_*`

This could be avoided by explicitly creating the directory, before fixing the permissions:
https://github.com/TYPO3-Solr/ext-solr/blob/master/Dockerfile#L11

Additionally, this combination would mount volumes (e.g. data volumes, like in the `docker-compose` example above) with correct permissions, too.

As far as I can see, this should not affect any other use cases.

SomeBdyElse:
Fixes #1278 for 6.1.* by cherrypicking d9fa187 into the release-6.1.x branch.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,8 @@ COPY Resources/Private/Solr/ /opt/solr/server/solr
 
 USER root
 
-RUN chown -R solr:solr /opt/solr/server/solr/
+RUN mkdir -p /opt/solr/server/solr/data && \
+    chown -R solr:solr /opt/solr/server/solr/
 
 USER solr
 


timohund:
@SomeBdyElse Thanks for the backport

PR #1400, opened 2017-07-11
[BUGFIX] Relation handler should handle pages overlays correctly

frans-beech-it:
Table pages has diferent handling over overlays. Now the
ContantObject/Relation handles these correctly.


timohund:
@frans-beech-it Thanks a lot for the patch. I'll open a new pr soon with your changes and with an integration test that covers this case.

PR #1409, opened 2017-07-14
[FEATURE] Rootpage without defautllanguage

maritwho:
If the root page is set to 'hide default translation of page' this causes a 404 in TypoScriptFrontendController with initializing the TSFE. To avoid this a language uid has to be thrown to the initialize function.


Classes/IndexQueue/Indexer.php
```diff
@@ -575,6 +584,29 @@ protected function getSolrConnectionsByItem(Item $item)
     }
 
     /**
+     * @param $item
```
timohund: We should add typehints to the signature

Issue #1461, opened 2017-07-25
Fix NullTimeTracker usage


georgringer:
`NullTimeTracker` has been removed in the core with CMS9 as the regular TimeTracker has been improved with CMS8.

PR #1462, opened 2017-07-25
[BUGFIX] Fix usage of NullTimeTracker

georgringer:
The NullTimeTracker has been deprecated with CMS8 and removed with
CMS9. Switch to the regular TimeTracker.

Resolves: #1461


timohund:
I think when we want to keep the functionallity we should instanciate TimeTracker with first argument=false.

georgringer:
absolutly true @timohund, fixed it ;)

timohund:
Thx

PR #1464, opened 2017-07-26
add missing dkd logo

zlorfi:


dkd-dobberkau:
Fine with it, but someone else must merge.

Issue #1470, opened 2017-07-26
Fix clickmenu


georgringer:
reseting the connection currently does not work

timohund:
Thx, i think this ist same as #1467

Issue #1469, opened 2017-07-26
[TASK] Check failing tests in master


timohund:
Check why travis build fails with:

Unable to determine path to entry script. Please check your path or set an environment variable 'TYPO3_PATH_ROOT' to your root path.

Maybe related to testingframework or TYPO3 8.7.4 changes? We need to check it.

timohund:
Thanks to @helhum this could be fixed

PR #1472, opened 2017-07-26
[BUGFIX] Repair initialize solr connection in clickmenu

georgringer:
Fix the JS to handle the click menu item and add a
separate class handling the request.

Resolves: #1470


timohund:
Build currently fails because of #1469, we shoul fix that first. Will check it tomorrow (mit related to your changes)

PR #1481, opened 2017-07-27
Improve composer.json

helhum:
* Remove minimum-stability
* Remove optimize-autoloader
* Add extension-key

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,6 @@
   "name": "apache-solr-for-typo3/solr",
   "type": "typo3-cms-extension",
   "description": "Apache Solr for TYPO3",
-  "minimum-stability": "alpha",
   "license": "GPL-2.0+",
   "keywords": ["typo3", "TYPO3 CMS", "solr", "search"],
   "homepage": "http://www.typo3-solr.com",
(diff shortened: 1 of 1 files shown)


timohund:
@dkd-kaehm No. We need to check why the functional test fails, in general the change to the composer.json are good but for some reason the test fails. I'll check today why.

timohund:
@helhum @dkd-kaehm The failing test is really a bug, cause of a wrong namespace in EXT:solr. I'll create a pull request including @helhum changes and the fix.

helhum:
@timohund To detect such bugs early on, I would suggest to not allow direct merges to the repo (only over pull requests) and enforce passing builds for pull request merges.

You can configure this in your Github project settings for each branch: https://github.com/TYPO3-Solr/ext-solr/settings/branches

dkd-kaehm:
I thought this were 100% our workflow, is not?

timohund:
@helhum we allready do this. 

The problem in this case was, as i understand:

```
   "optimize-autoloader": true,
```

made the class available. But for sure it should be loadable with psr-2, so removing "optimize-autoloader": true just showed the (also previously) existing problem.

helhum:
@timohund 
> we allready do this

great!

If `"optimize-autoloader": true` caused the class to be available (which in fact should not happen, if the folder is scanned with a psr-4 directive), then this was broken for consumers of solr, as `"optimize-autoloader` is a root only directive.

Anyway, good the tests are green now and composer.json is cleaned up :)

PR #1493, opened 2017-07-30
[TASK] Wrap code in ext_locaconf/ext_tables

georgringer:
To avoid leaking variables the code should be wrapped.


timohund:
@georgringer Thanks for the patch. I think we should pospone this to a later release and check why the tests are failing.

3l73:
Closed in favor for #2808 which based on the current master.

Issue #1495, opened 2017-07-31
[BUG] Partials from Index Inspector are not set


dkd-kaehm:
StandaloneView in "AbstractFunctionModule" needs the Partials path.

PR #1496, opened 2017-07-31
[BUGFIX] Partials for Index Inspector are not set

dkd-kaehm:
added Partials path to StandaloneView in "IndexInspectorController"

Fixes: #1495


timohund:
@dkd-kaehm Looks good to me thanks for the patch

PR #1519, opened 2017-08-04
[TASK] Add styleci

georgringer:
Provide a configuration for styleci and a fix for all issues.

Best would be to enable styleci after the merge as service to do tests automatically


timohund:
We currently use php-cs-fixer and the results conflict with the changes from StyleCI. I would proposer to keep php-cs-fixer becaus you can use it in the devbox and on the commandline.

Issue #1405, opened 2017-07-12
[TASK] Extract acces to FrontendSession into FrontendSession object


timohund:
There currently multiple places where $GLOBLAS['TSFE']->fe_user->setKey and  $GLOBLAS['TSFE']->fe_user->getKey is used. These accesses can be extracted into an own FrontendSession class that can be located in "System/Session/"

Todo:

* Add a class System/FrontendSession that encapsulates the acesses to $GLOBALS['TSFE']->fe_user->setKey and getKey
* Extract access to session from LastSearchesService into FrontendSession
* Extract access to session from SearchResultSetService into FrontendSession

PR #1564, opened 2017-08-16
[TASK] Extract access to frontend user session

timohund:
This PR:

* Extracts the access to the frontend user session into System/Session/FrontendUserSession
* Simplifies the SearchResultSetService and LastSearchesService

Fixes: #1405

### DIFF
--- a/Classes/System/Session/FrontendUserSession.php
+++ b/Classes/System/Session/FrontendUserSession.php
@@ -1,0 +1,93 @@
+<?php declare(strict_types = 1);
+
+namespace ApacheSolrForTypo3\Solr\System\Session;
+
+/***************************************************************
+ *  Copyright notice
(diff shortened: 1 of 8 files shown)


Classes/System/Session/FrontendUserSession.php
```diff
@@ -0,0 +1,93 @@
...
+     */
+    public function setPerPage(int $requestedPerPage)
+    {
+        $this->feUser->setKey('ses', 'tx_solr_resultsPerPage', intval($requestedPerPage));
+    }
+
+    /**
+     * @return int
+     */
+    public function getPerPage()
```
dkd-kaehm: Do we not need PHP 7 return type here and below?  `public function getPerPage() : int`

Issue #1579, opened 2017-08-18
[TASK] Use SQLs AVG() instead of calculating in PHP method for statistics result


dkd-kaehm:
A method `StatisticsRepository::getAverageFromField()` can be replaced by doing it with SQLs AVG+groupBy 

WDYT about following addition to `StatisticsRepository::getPreparedQueryBuilderForSearchStatisticsAndTopKeywords()`: 
```
->select('keywords', 'num_found AS hits')
->add('select', $queryBuilder->expr()->count('keywords', 'count'), true)
->add('select', $queryBuilder->expr()->avg('num_found', 'hitsAvg'), true) // <-- added SQLs AVG
```
@mschwemer: can you evaluate this here?

PR #1573, opened 2017-08-17
[BUGFIX] Fix sorting and counts in statistics module

mschwemer:
The numbers of the top 5 and the search statistics did not match. This
change respects all search queries in the timeframe and sorts the merged
statistics rows according to their numbers

Fixes: #1572

Issue #1572, opened 2017-08-17
[BUG] Improve / fix statistics


mschwemer:
In the statistics module the top 5 results and the search result statistics are not in sync from regarding the numbers. Furthermore the sorting may be wrong, when result rows are merged.

The solution is to consider all search requests in the timeframe, merge the statistics rows, sort them  regarding the count and then slice the array to provide the desired result number.

PR is ready and will follow in a couple of minutes.

A backport to the 7.x would be cool.

PR #1582, opened 2017-08-18
[BUGFIX] Fix sorting and counts in statistics module

dkd-kaehm:
* fix "top 5" and "search statistics" queries for correct calculations
* remove PHP calculations where possible and replace it with SQL

fixes: #1579, closes: #1573, fixes: #1572

### DIFF
--- a/Classes/Domain/Search/LastSearches/LastSearchesRepository.php
+++ b/Classes/Domain/Search/LastSearches/LastSearchesRepository.php
@@ -108,19 +108,6 @@ class LastSearchesRepository extends AbstractRepository
     }
 
     /**
-     * Returns current count of last searches
-     *
-     * @return int
-     */
-    protected function count() : int
-    {
-        return $this->getQueryBuilder()
-            ->count('*')
-            ->from($this->table)
-            ->execute()->fetchColumn(0);
-    }
-
-    /**
      * Resolves next sequence id by given last searches limit.
      *
      * @param int $lastSearchesLimit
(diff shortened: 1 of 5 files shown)


dkd-kaehm:
@mschwemer could you please check if this works as expected.

mschwemer:
@dkd-kaehm Thanks. The counted numbers seem ok and deliver the same results for the TOP5 and the TOP100.

But the percentage is way too low. 
solr_search_percentage

Until recently the top100 summed up to 100%. Our top searches had values from 3 to 8 percent.

mschwemer:
Here is a screen  with numbers from master
solr_search_percentage_old

dkd-kaehm:
@mschwemer please check again, I fixed it. We must count the rows by root page id.
I guess you are using multiple site roots...

dkd-kaehm:
DBAL converts decimals to string for keeping high precision: ["For compatibility reasons this type is not converted to a double as PHP can only preserve the precision to a certain degree. Otherwise it approximates precision which can lead to false assumptions in applications."](http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/types.html#decimal)

So the test could fail on other DB platforms.

dkd-kaehm:
@mschwemer: could you please review this PR. 
Does it work as expected?

mschwemer:
@dkd-kaehm Thx for the reminder ... yes it works as expected.

dkd-kaehm:
Thanks Marcus, I'll do the same for 7.0

Issue #1575, opened 2017-08-17
[BUG] MySQL Query for search buckets is faulty


mschwemer:
The database query for calculating the buckets in the statistics module is partially faulty. In certain mysql server conditions, this leads to a graph like this: 

<img width="1047" alt="solr_search_graph" src="https://user-images.githubusercontent.com/103594/29425188-a6ea8fc2-8382-11e7-8265-bea4e7798c79.png">

For each day there are two rendered points. 

Reason is the function "getQueriesOverTime" in "EXT:solr/Classes/Domain/Search/Statistics/StatisticsRepository.php". There is a todo in there ...

But even for mySQL this does not work in every case. The mySQL docs warn that there might be different results: https://dev.mysql.com/doc/refman/5.7/en/date-and-time-functions.html#function_unix-timestamp

dkd-kaehm:
@mschwemer Thanks for reporting.
Will #1556 solve this issue?

mschwemer:
@dkd-kaehm I think so.

PR #1591, opened 2017-08-24
[BUGFIX] MySQL Query for search buckets is faulty

### LINKED ISSUE #1556: [TASK] add and use date column to tx_solr_statistics
This will add support for all with doctrine DBAL compatible platforms by using query in `StatisticsRepository::getQueriesOverTime()` method.

See: @todo comment in [StatisticsRepository::getQueriesOverTime()](https://github.com/TYPO3-Solr/ext-solr/blob/9f83ded9e52d08f7d50846bfe292a99866775eca/Classes/Domain/Search/Statistics/StatisticsRepository.php#L204)

dkd-kaehm:
* replace excl. MySQL functions with pure math in getQueriesOverTime()
* fixes faulty calculation, because bucket and calculated timestamp was calculated for wrong day
* closes #1556, because using new date column requires database migrations

Fixes: #1575

### DIFF
--- a/Classes/Domain/Search/Statistics/StatisticsRepository.php
+++ b/Classes/Domain/Search/Statistics/StatisticsRepository.php
@@ -201,8 +201,7 @@ class StatisticsRepository extends AbstractRepository
         $result = $queryBuilder
             ->addSelectLiteral(
                 'FLOOR(`tstamp`/' . $bucketSeconds . ') AS `bucket`',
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
@mschwemer could you please review this PR

mschwemer:
Hi,
I tested it successfully on our current data-set. It also looks good by reading.

Thanks for the fix!
Marcus

mschwemer:
BTW: Would be cool if this would be merged into a 7.x.x bugfix release

PR #1612, opened 2017-09-04
[BUGFIX] recursive value resolution - mm relations

### LINKED ISSUE #1608: RecursiveValueResolution does not work on n:m relations
When enableRecursiveValueResolution is used and the foreignTable value field references another table , only the first record of the foreign table is indexed.

IMHO the problem is the getRelatedItemsFromMMTable method in ApacheSolrForTypo3\Solr\ContentObject\Relation.

All fine, all related records will be returned.
https://github.com/TYPO3-Solr/ext-solr/blob/c4b7b22af2a8a6133c338cc3fac4f4ac19cc9fab/Classes/ContentObject/Relation.php#L186

running through all relatedItems to fetch related items. Still good.
https://github.com/TYPO3-Solr/ext-solr/blob/c4b7b22af2a8a6133c338cc3fac4f4ac19cc9fab/Classes/ContentObject/Relation.php#L187

but then: return after getting recursive value items for the FIRST records. All other records are not resolved. 
https://github.com/TYPO3-Solr/ext-solr/blob/c4b7b22af2a8a6133c338cc3fac4f4ac19cc9fab/Classes/ContentObject/Relation.php#L202

IMHO the resulting items of $this->getRelatedItems($contentObject) should be added to $relatedItems array.

AlexKvrlp:
respects all records of mm relations when enableRecursiveValueResultion is set.
Fixes #1608

### DIFF
--- a/Classes/ContentObject/Relation.php
+++ b/Classes/ContentObject/Relation.php
@@ -199,7 +199,7 @@ class Relation
                 $contentObject = GeneralUtility::makeInstance(ContentObjectRenderer::class);
                 $contentObject->start($record, $foreignTableName);
 
-                return $this->getRelatedItems($contentObject);
+                $relatedItems = array_merge($relatedItems,$this->getRelatedItems($contentObject));
             } else {
                 if ($GLOBALS['TSFE']->sys_language_uid > 0) {
                     $record = $this->getTranslationOverlay($foreignTableName, $record);


timohund:
@AlexKvrlp Can you please describe the problem with an example or add a testcase to the integration test that tests this scenario?

AlexKvrlp:
I will do my best.

**Models**
we have three models:
root model "taetigkeit" has an mm relation to model "behandlungsart".
"behandlungsart" has an n:1 relation to model "fachbegriff". 
"fachbegriff" has a property "value" which contains a string and is used as label.
"behandlungsart" uses the field "fachbegriff" as label.

Now we want to index "taetigkeit" with its "behandlungsarten".
Related indexing configuration looks like this

**Index Configuration**
plugin.tx_solr.index {
    queue {
        fooproject = 1
        fooproject {
            table = tx_kvrlpfooproject_domain_model_taetigkeit
            fields {
                behandlungsarten_stringM = SOLR_RELATION
                behandlungsarten_stringM{
                    localField = behandlungsarten
                    multiValue = 1
                    enableRecursiveValueResolution = 1
                }
		//other fields...
            }
        }
    }
}

**Behaviour without patch**
Only the first "behandlungsart" record is resolved when enableRecursiveValueResolution is enabled.  When enableRecursiveValueResolution is disabled and another "behandlungsart" field is used as label it works.

timohund:
@AlexKvrlp can you please fix the spacing issue reported by scrutinizer:

https://scrutinizer-ci.com/g/TYPO3-Solr/ext-solr/inspections/f19c920e-8f86-477f-932f-f2d0bb3337d4

?

timohund:
Hi @AlexKvrlp i've added two tests that resolve multiple relations (direct relation and mm relations): #1631 in both everything works as expected. Can you add a testcase for your use case?

timohund:
Closing because no response on question

frleb:
Hi, 
Can we reopen this issue ? Someone has pushed a fix on https://github.com/TYPO3-Solr/ext-solr/pull/1612
I think the actual code doesnt support nested MM relations. It only returns the first element instead of multiple elements.
Thanks.
FL

dkd-kaehm:
@frleb please open new one with description of problem and reference this issue.

PR #1623, opened 2017-09-06
[BUGFIX] Fix wrong parameter value

JNellissen-BM:
Change wrong assignment for group.field parameter


timohund:
@JNellissen-BM Thanks for the patch, do you have the time to add a unit test for that? Otherwise i'll have a look on that later.

Issue #1622, opened 2017-09-06
[TASK] Remove DatabaseUtility and replace its functionality with native Doctrine


dkd-kaehm:
* Remove DatabaseUtility and replace its usages by Doctrine

PR #1639, opened 2017-09-08
[TASK] Deprecate DatabaseUtility and replace it with native Doctrine

dkd-kaehm:
* deprecated DatabaseUtility class and replaced with doctrines transaction methods
* added AbstractRepository::getConnectionForAllInTransactionInvolvedTables()
* replaced usages of DatabaseUtility::*

Fixes: #1622


dkd-kaehm:
The scrutinizer issue and patch are strange, PhpStorm generates the doc-comment and `array_shift(` in combination with `funcName(string ...$tableNames)` definitely that what scrutinizer expects.

timohund:
@dkd-kaehm what about to assign it to a variable $firstTableName and cast it to a string, that makes also clearer what it is:

```php
$firstTableName = (string)array_shift($tableNames);
```

Issue #1370, opened 2017-06-22
[TASK] Use doctrine in ApacheSolrForTypo3\Solr\ContentObject\Relation


timohund:
DatabaseConnection a.k.a. $["TYPO3_DB"] has been marked as deprecated in TYPO3 v8 and will be removed in TYPO3 v9. Please use the newly available ConnectionPool and QueryBuilder classes.

We should migrate the Relation class to use the QueryBuilder.

timohund:
Split issue from #1165

mbresch-sf:
Okay, will fix this task.

dkd-kaehm:
@mbresch-sf Do you need help here? What is missing?

PR #1640, opened 2017-09-08
[TASK] Use doctrine in ApacheSolrForTypo3\Solr\ContentObject\Relation

dkd-kaehm:
The old MySQL-Only native function "FIELD(uid..." is not supported by other platforms,
therefore we need to order it in PHP

Fixes: #1370


dkd-kaehm:
@mbresch-sf please review

PR #1641, opened 2017-09-11
[TASK] test against CMS 9.x-dev

dkd-kaehm:
* increase nimut/testing-framework to 2.0
* mark temporary CMS 9.x-dev as 8.7.999 for compatibility with nimut...


dkd-kaehm:
fixed in #1774

Issue #1633, opened 2017-09-07
[TASK] Backport #1625 to 7.0.x branch


timohund:
We shourt set the default pathes also in the 7.0.x branch

#1625 Needs to be backported

PR #1625, opened 2017-09-06
[BUGFIX] Add default typoscript view configuration

jacobsenj:
Fixes #1624

PR #1642, opened 2017-09-11
[BUGFIX] Backport: Add default typoscript view configuration

dkd-kaehm:
since CMS 8.7.5 templates -, partials - and layout paths must be set

comes from: #1625

Fixes: #1633


dkd-kaehm:
closing for #1645

Issue #1644, opened 2017-09-11
[TASK] Backport #1611 into release-7.0.x


dkd-kaehm:
* Changes the expected id since the encryption key of the testing framework was changed (https://github.com/TYPO3/testing-framework/commit/ec72500b23d651862ca681d366393c428f957e4f).
* Allows master build to fail against 8.x-dev since there is another problem with FLUID fallbacks

PR: #1611 must be backported into release-7.0.x branch

PR #1645, opened 2017-09-11
[TASK] Backport: Check why build fails

dkd-kaehm:
This pr:

* Changes the expected id since the encryption key of the testing framework was changed.
* since CMS 8.7.5 templates -, partials - and layout paths must be set
* Sets the path to "module.tx_solr" because the test is running in the backend context
  * We need to find a solution to run it in a real frontend context

Comes From: #1611, #1616, #1625, 

Fixes: #1644, fixes: #1633


dkd-kaehm:
Please do not squash on merging.

PR #1646, opened 2017-09-11
[BUGFIX] wrong parameter value

dkd-kaehm:
* fixed wrong assignment for group.field parameter
* added missing unit tests for build() method

Fixes: #1623 

New PR, because contribution checkbox for #1623 is not set.


dkd-kaehm:
Oha pushing was errored, but anyway updated... closing this, because PR is updated

PR #1649, opened 2017-09-13
Adds index indexing_priority_changed

### LINKED ISSUE #1648: Optimizing table index for tx_solr_indexqueue_item
We have a huge table with over 80k entries in tx_solr_indexqueue_item table and we realized that there are over 5,000 calls in 12 hours which use a MySQL tablescan of 32k.

With adding an additional index the tablescan can be improved to 8k.

spoonerWeb:
Closes #1648.

### DIFF
--- a/ext_tables.sql
+++ b/ext_tables.sql
@@ -62,6 +62,7 @@ CREATE TABLE tx_solr_indexqueue_item (
 
 	PRIMARY KEY (uid),
 	KEY changed (changed),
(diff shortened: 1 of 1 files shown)


timohund:
@spoonerWeb Thanks for the report and the patch!

spoonerWeb:
@timohund Would be nice to have a backport to some branches. We use not the current version. ;) Thanks!

PR #1652, opened 2017-09-18
[TASK] Enable escaping in TranslateViewHelper and AbstractWidgetViewHelper

timohund:
This pr:

* Enabled the escaping for the Translate and AbstractWidgetViewHelper to avoid the output of unwanted content.

### DIFF
--- a/Resources/Private/Templates/ViewHelpers/Widget/ResultPaginate/Index.html
+++ b/Resources/Private/Templates/ViewHelpers/Widget/ResultPaginate/Index.html
@@ -20,10 +20,20 @@
 					<li class="previous">
 						<f:if condition="{pagination.previousPage} > 1">
 							<f:then>
-								<a href="{s:uri.paginate.resultPage(page: pagination.previousPage)}" class="solr-ajaxified"><s:translate key="paginate_previous">&laquo;</s:translate></a>
+								<a href="{s:uri.paginate.resultPage(page: pagination.previousPage)}" class="solr-ajaxified">
+									<f:if condition="{s:translate(key: 'paginate_previous')}">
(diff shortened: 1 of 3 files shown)


Resources/Private/Templates/ViewHelpers/Widget/ResultPaginate/Index.html
```diff
@@ -20,10 +20,20 @@
...
+								<a href="{s:uri.paginate.resultPage(page: pagination.previousPage)}" class="solr-ajaxified">
+									<f:if condition="{s:translate(key: 'paginate_previous')}">
+										<f:then><s:translate key="paginate_previous" /></f:then>
+										<f:else>&laquo;</f:else>
+									</f:if>
+								</a>
 							</f:then>
 							<f:else>
-								<a href="{s:uri.paginate.resultPage()}" class="solr-ajaxified"><s:translate key="paginate_previous">&laquo;</s:translate></a>
+								<a href="{s:uri.paginate.resultPage()}" class="solr-ajaxified">
```
dkd-kaehm: why not using default="&laquo;" instead of conditions?

Resources/Private/Templates/ViewHelpers/Widget/ResultPaginate/Index.html
```diff
@@ -20,10 +20,20 @@
...
+									<f:if condition="{s:translate(key: 'paginate_previous')}">
```
dkd-kaehm: why not using `default="&laquo;"` instead of conditions?

timohund: Because it will be escaped ;)

Issue #1656, opened 2017-09-19
Facets containing `+` are not handled correctly


alappe:
TYPO3 version: `7.6.x.`
ext:solr Version: `6.1.3`
Solr version: `6.0.0`
PHP version: `7.0.18`
Expected behavior: I can have a facet named `A+B` and filter for it.
Actual behavior: The `+` character is wrongly urldecoded as ` ` (space), thus resulting for a non-matching filter of `A B`.
Steps how to reproduce the issue: See above.

PR #1657, opened 2017-09-19
[BUGFIX] Use rawurldecode on facets to handle `+`

alappe:
The test case might be a bit weak and this PR does not regard any *en*coding of those values, since we're not using them in our projects, so I can't attest to it. However, this fixes my issues, so I'm looking for feedback…

Related: #1656


timohund:
@alappe Thanks for the pr. Are you using the old templating only or in combination with solrfluid?

alappe:
Together with solrfluid but mostly with js code doing ajax calls.

hirnschmalz:
Is there a reason why this PR hasn't made it into to code after almos 2 years?

timohund:
@hirnschmalz yes because it has not be rebased. The diff show's changes in **203** files. If you have the time it would be nice if you could rebase it and create a PR were the changes can be seen directly.

3l73:
Closed in favor for #2806 

New pull request based on current master with changes from this pull request.

PR #1663, opened 2017-09-22
[TASK] add singular form for results_found

DanielRuf:
The translations have to be updated accordingly and I did not know what I should write in `browse.vm`.

### DIFF
--- a/Resources/Private/Templates/Search/Results.html
+++ b/Resources/Private/Templates/Search/Results.html
@@ -37,8 +37,16 @@
 		</f:if>
 
 		<f:if condition="{resultSet.usedSearch.numberOfResults}">
+			
 			<span class="result-found">
-				<s:translate key="results_found" arguments="{0:resultSet.usedSearch.numberOfResults, 1: resultSet.usedSearch.queryTime}">Found %d results in %d seconds</s:translate>
+				<f:if condition="{resultSet.usedSearch.numberOfResults} == 1">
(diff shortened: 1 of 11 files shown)


Resources/Private/Templates/Search/Results.html
```diff
@@ -37,8 +37,16 @@
 		</f:if>
 
 		<f:if condition="{resultSet.usedSearch.numberOfResults}">
+			
 			<span class="result-found">
-				<s:translate key="results_found" arguments="{0:resultSet.usedSearch.numberOfResults, 1: resultSet.usedSearch.queryTime}">Found %d results in %d seconds</s:translate>
+				<f:if condition="{resultSet.usedSearch.numberOfResults -> f:count()} == 1">
```
timohund: usedSearch->getNumberOfResults() should allready return the count, why is this required? The test also shows an error that the argument of the count ViewHelper should not be an integer. I think this should be enought:

```php
<f:if condition="{resultSet.usedSearch.numberOfResults} == 1">
```

DanielRuf: Ah ok, I think I oversaw this. Can you push to my branch and make the needed change? Otherwise I'll do the change this evening.

DanielRuf:
I've done the requested changes and pushed a new commit.

DanielRuf:
Thank you too for the time that you took for my PR.

PR #1671, opened 2017-09-28
[BUGFIX] Rename allowSelfSignedCertificates and set verify_peer_name to false

### LINKED ISSUE #1670: [BUGFIX] Rename allowSelfSignedCertificates and set verify_peer_name to false too
Rename extension configuration option `allowSelfSignedCertificates` to `allowInvalidCertificates` 
 (optional, but imho a better name) and set the stream context option `verify_peer_name` to false too to prevent warnings (and subsequently and exception) like:
`PHP Warning:  file_get_contents(): Peer certificate CN=invalid.test.domain did not match expected CN=valid.production.domain`

Related to #1133

wazum:
To allow all invalid certificates (not only self signed).

Fixes #1670

### DIFF
--- a/Classes/IndexQueue/PageIndexerRequest.php
+++ b/Classes/IndexQueue/PageIndexerRequest.php
@@ -396,10 +396,9 @@ class PageIndexerRequest
             ],
         ];
 
-        if ($this->extensionConfiguration->getIsInvalidCertificatesEnabled()) {
+        if ($this->extensionConfiguration->getIsSelfSignedCertificatesEnabled()) {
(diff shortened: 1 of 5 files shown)


wazum:
Reverted most of the patch and just left the `verify_peer_name => false` part, another option just for that setting is seems unnecessary for me

timohund:
I would agree to the comments of @dkd-kaehm and not just add it, we should add an additional setting for this.

timohund:
Closing because no response for change request

PR #1672, opened 2017-10-04
Update Solr.rst

neonaut:
fixed typo


timohund:
@neonaut Thanks for the pr

PR #1674, opened 2017-10-09
[TASK] clean up old DatabaseConnection usages

dkd-kaehm:
remove all usages of old DatabaseConnection class


dkd-kaehm:
All occurrences and leftovers should be collected in this PR issue.

Issue #1681, opened 2017-10-11
[BUG] start attribute by widget.resultPaginate is not correct


pniederlag:
{pagination.displayRangeStart} is calculated as the number of the resultPage that is shown in the pagination control. (it will stay on 1 all the time if you have no more than 10 pages to show!)

however this number is also used as start attribute for the &lt;ol&gt; tag to indicate the number of the individual count on the result item in https://github.com/TYPO3-Solr/ext-solr/blob/dfb1ea4c1db892490e6e1fe1a7c7dc29b8f37e2d/Resources/Private/Templates/Search/Results.html#L64

Obviously this is not correct. It will (almost) always show 1 ... 10 (on page 1, page 2, page 3, ...) and only goes up when the range of pages shown is shifting (starting on 2 or higher).

Hope you get the point ;)

timohund:
@pniederlag Thanks for your report. Are you talking about version 7 or dev-master?

pniederlag:
i bumped into it on 7.0.2 but think it's still broken in master. will upload a patch proposal in a couple of minutes (pending for testing)

pniederlag:
please check, it fixes the problem on 7.0.2, did not look into master yet. If you are fine with the fix I can port it to master

timohund:
@pniederlag I've merged the patch. Can you create an additional pr for release-7.5.x please? I will also port it to master, because for master there is more todo, because of the pagination for grouping.

pniederlag:
https://github.com/TYPO3-Solr/ext-solr/pull/1718

Gimme a ping and I'll try to tackle master

timohund:
@pniederlag i think this is resolved and we can close the issue. If you have any objections please create a new one

PR #1682, opened 2017-10-11
[BUGFIX] broken count in pagination #1681

pniederlag:
fix start attribute for <ol> tag by calculating and adding a proper count into the pagination data.

Resolves: #1681


timohund:
@pniederlag sorry, no, currently i do not get the time to look on it.

timohund:
@pniederlag Thanks for the patch and sorry for the delay.

PR #1684, opened 2017-10-12
[BUGFIX] Convert function argument to int to match signature

pschriner:
The identifier field of AbstractProvider is defined as being of type string.
The RootPageResolver expects an int as argument for the call to getIsRootPageId(). This mismatch
causes a check for pageId 0 to fail


pschriner:
Question: How can I get a backport for this patch? We're running 6.1.3.

timohund:
@pschriner Thanks for the patch!  Could you also provide a pr for 6.5 and 7.5? Otherwise we could do that later.

pschriner:
https://github.com/TYPO3-Solr/ext-solr/pull/1704
https://github.com/TYPO3-Solr/ext-solr/pull/1705

Made two PRs - thanks for your work!

timohund:
@pschriner thanks

PR #1690, opened 2017-10-17
Feature/1574 enable tie parameter

mschwemer:
Hi,

as promised I added the tie parameter to the list of query parameters.

I'm looking forward to your feedback.

Marcus


timohund:
Hi Marcus,

thanks for your usefull patch. I'll have a look on it, as soon as i get the time to do that.
The build currently fails because of external dependencies and i guess it is not related to your change.

Timo

Issue #1711, opened 2017-10-27
[BUGFIX] missing array keys in facet options after manual sorting


mbresch-sf:
If you are using `faceting.facets.[facetName].manualSortOrder` the array keys of `OptionCollection->data[]` will gone.

Before sorting there is an array like
```
$data['filter1']
$data['filter2']
$data['filter3']
```
With the following option in TypoScript 
```
 faceting.facets.[facetName].manualSortOrder = filter3, filter2
```
you will get an array like
```php
$data[0]
$data[1]
$data['filter1']
```
I will provide a small fix soon.

PR #1712, opened 2017-10-27
[BUGFIX] missing array keys in facet options after manual sorting #1711

mbresch-sf:
now the array key is used


timohund:
@mbresch-sf Thanks for the patch. Can you add more details to the issue and is it possible to add a unit test for that?

mbresch-sf:
Hi @timohund,
I will try to add a unit test for that, yes.

I wrote an afterSearchHook to check available filters for a query group facet. If a filter is not available than I have to add this missing filter. In this case I used `OptionCollection->getByValue()` and this method works fine without manual sorting. But using manual sorting this method doesn't work.

timohund:
Created new pr #1793 with your changes + test. Thanks for the bug report and the fix!

PR #1721, opened 2017-11-03
[BUGFIX] Index news links with language parameter

dnlm:
Search results for news in other languages than default linked to the default language


timohund:
@dnlm Thanks for the pr, shouldn't this be &L={field:__solr_index_language} ? That's how it is configured for tt_news and i think how it was intended to use.

dnlm:
You are right, that's what it is for. It's even mentioned in the [documentation](https://github.com/TYPO3-Solr/ext-solr/blob/master/Documentation/Configuration/Reference/TxSolrIndex.rst#queue-indexconfig).
I think it would be a good idea though to include that in the news example to make it work for multilanguage sites out of the box.

PR #1735, opened 2017-11-14
[FEATURE] Make Phrase Search configurable

dkd-kaehm:
This enables Apache Solrs' phrase search feature.

### DIFF
--- a/Classes/Query.php
+++ b/Classes/Query.php
@@ -25,12 +25,15 @@ namespace ApacheSolrForTypo3\Solr;
  ***************************************************************/
 
 use ApacheSolrForTypo3\Solr\Domain\Search\Query\Helper\EscapeService;
+use ApacheSolrForTypo3\Solr\Domain\Search\Query\ParameterBuilder\BigramPhraseFields;
 use ApacheSolrForTypo3\Solr\Domain\Search\Query\ParameterBuilder\Faceting;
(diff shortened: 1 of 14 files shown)


Classes/Query.php
```diff
@@ -131,6 +134,36 @@ class Query
...
+     * @see https://lucene.apache.org/solr/guide/7_0/the-dismax-query-parser.html#pf-phrase-fields-parameter
+     */
+    protected $phraseFields;
+
+    /**
+     * Holds the bigram phrase fields with their associated boosts. The key represents
+     * the field name, value represents the field's boost. These are the fields
+     * for those Apache Solr should build the phrases from triplets and sentences.
+     *
+     * @var PhraseFields
```
timohund: I think this should be BigramPhraseField?

Classes/Query.php
```diff
@@ -131,6 +134,36 @@ class Query
...
+     * @see https://lucene.apache.org/solr/guide/7_0/the-extended-dismax-query-parser.html#extended-dismax-parameters
+     */
+    protected $bigramPhraseFields;
+
+    /**
+     * Holds the trigram phrase fields with their associated boosts. The key represents
+     * the field name, value represents the field's boost. These are the fields
+     * for those Apache Solr should build the phrases from triplets and sentences.
+     *
+     * @var PhraseFields
```
timohund: I think this should be TrigramPhraseField?

Issue #1589, opened 2017-08-24
[TASK] Clarify what should happen with faceting.facets.[facetName].singleOptionMode


dkd-kaehm:
In the Docs the TS option faceting.facets.[facetName].singleOptionMode is described but nowhere implemented.

What should we do?:
* Implement this feature
* Add the Hint in the docs that it is dropped and how to do the same instead

mbresch-sf:
I think the best way is to keep (get back) this option in 7.0. 

The "solrfluid" solution is fine too, but you should deprecate the singleOptionMode before dropping
this without any hint. :-)

timohund:
@dkd-kaehm @mbresch-sf 

I would proposer to add another partials for that, that shows how it works and to document that, otherwise we maintain this logic in EXT:solr and it is allready possible just by using the correct methods in the template.

PR #1740, opened 2017-11-20
[TASK] Remove singleOptionMode from documentation

timohund:
This pr:

* Add's an example typoscript configuration how to configure the singleOptionMode with fluid templating
* Removes the setting from the documentation
* Add's an faq entry how to implement this behaviour with fluid

Fixes: #1589


dogawaf:
Hello
I would suggest to cherry pick this commit for the next 7.5 release (if any) because the option was already deleted on this branch, but the doc remained incorrect.
Cheers.

PR #1744, opened 2017-11-21
[TASK] Fix deprecations in flexforms

georgringer:
Fixes deprecations for the results plugin:

Related issue:  #1750


timohund:
We need the same for the forms plugin, then we can close the related issue. Thanks for the patch.

Issue #1689, opened 2017-10-16
Multivalue field outputs string when there is 1 value. I expect always an array


jvanderkroon:
When there is 1 value in a stringM (or other multivalue field) the output is a string. When there are more value's the output is an array. 

For templating it would be nice if the output is always an array so i can make a foreach.

PR #1748, opened 2017-11-22
[BUGFIX] Always return value as expected

IchHabRecht:
If a field is an array, an array should be returned within Solr
document. This is necessary for e.g. be able to iterate over a property
in a Fluid file. If returning either an array or a string, an exception
in Fluid is thrown because it expects an array.

Resolves: #1689

### DIFF
--- a/Resources/Private/Php/SolrPhpClient/Apache/Solr/Response.php
+++ b/Resources/Private/Php/SolrPhpClient/Apache/Solr/Response.php
@@ -199,13 +199,6 @@ class Apache_Solr_Response
                 }
 
                 foreach ($originalDocument as $key => $value) {
-                    //If a result is an array with only a single
-                    //value then its nice to be able to access
-                    //it as if it were always a single value
-                    if ($this->_collapseSingleValueArrays && is_array($value) && count($value) <= 1) {
-                        $value = array_shift($value);
-                    }
-
                     $document->$key = $value;
                 }
 


timohund:
@IchHabRecht Thanks for the patch

masi:
The patch isn't clean. Instead of hacking the library the solr extension should make use of the collapseSingleValueArrays.

eg. solr/Classes/ConnectionManager.php

protected function buildSolrService($host, $port, $path, $scheme)
{
  $solrService = GeneralUtility::makeInstance(SolrService::class, $host, $port, $path, $scheme);
  $solrService->setCollapseSingleValueArrays(false);
  return $solrService;
}

It's not configurable as I have suggested in #1773, but is in line with PhpSolr if it should ever be maintained again.

IchHabRecht:
Hi @masi,

before I pushed the patch for PhpSolr Client, I had a look into the Git repository and saw there were changes done before. That is why I suggested this "quick + dirty" fix.

masi:
Hi @IchHabRecht,

well in that case... Wouldn't have believed to be it your patch otherwise ;)

timohund:
@IchHabRecht @masi would it be fine for you when we add a setting for that that turns of the collapsing by default. Then somebody who needs the old behaviour can turn it off? 

It think the goal in the long run should be to get rid of the SolrPHPClient because:

* It is pretty old and not maintained
* We have other issues with that (allow_url_fopen)

I think it is important that we do this in a way that we can continously test it and everybody who uses EXT:solr is able to upgraded with reasonable effort.

Currently i see two options:

* Replace it with a framework e.g. solarium
* Use guzzle and do the rest optimized for our needs

Currently i am more for the second option because we allready did a lot for parsing of the solr responses and i think solarium does not offer all that we need and we need to translate a lot to use that api.

But i think i any case we should reduce the amount of settings and other thinks that increas the coupling to SolrPhpClient. Therefore i also don't see anything wrong with the change of @IchHabRecht but adding a setting for it is also fine. It would see that different when SolrPHPClient is a very good maintained API that we could update continously.

timohund:
I've added an issue for the SolrPHPClient replacement:  #1778

IchHabRecht:
I personally don't need any setting for this. In regards of using fluid, it doesn't make any sense, to provide a string where I would expect to get an array. This needs special handling. I would really recommend to not provide any configuration here.

timohund:
@masi For what do you need a configuration to turn it off?

masi:
@timohund The default behaviour of the lib sucks. But as it has a way to change its behaviour I had used it. And perhaps someone needs it so he/she could turn it on again.

Anway, simply changing the default is IMHO still a bit nicer than hacking the feature out of the lib.

```
class Apache_Solr_Service
{
  protected $_collapseSingleValueArrays = false;
}
```
I won't insist on this solution. But I'm willing a path that reverts the hack, does the above and add a line of how-to-get-to-stupid-old-way back, eg with an XCLASS:

```
	protected function initializeAction() {
		parent::initializeAction();
		this->searchService->getSearch()->getSolrConnection()->setCollapseSingleValueArrays(true);
	}
```

As for Solarium, AFAIK it has many features built-in that the Solr-Ext also uses. To deduplicate the code it would probably be better to remove some parts of the Solr-Ext to make use of the stuff Solarium already offers and makes use of code that's also used from many more users . OTOH, This could jeoparidize the stability of the extension. The price to pay is that you have to reinvent the wheel again and again (for every new Solr feature).

timohund:
@masi sounds reasonable, can you provide a pull request for that?

Issue #1783, opened 2017-12-13
[BUG] Text in hierarchical text facet should contain slashes


timohund:
When you add a hierarchical field with text (e.g. by using the pathToHierarchyProcessor) it should be possible that the name of an item should contain a "/". Currently this is not possible because the character "/" is used as seperator and not handled correctly when it is escaped ( "\/")

PR #1784, opened 2017-12-13
[BUGFIX] Text in hierarchical text facet should contain slashes

timohund:
This pr:

* Fixes the bug
* Add's testcases

Fixes: #1783


markuskobligk:
Confirmed working.

Issue #935, opened 2017-02-01
[TASK] Upgrade license to GPLv3


irnnr:
Per the discussion in #930; Upgrade license to GPLv3.

PR #1792, opened 2017-12-27
[TASK] Update license to gpl version 3

timohund:
This pr:

* Updates the license to the gpl version 3.0

Fixes: #935

### DIFF
--- a/Classes/AbstractDataHandlerListener.php
+++ b/Classes/AbstractDataHandlerListener.php
@@ -10,7 +10,7 @@ namespace ApacheSolrForTypo3\Solr;
  *  This script is part of the TYPO3 project. The TYPO3 project is
  *  free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License as published by
- *  the Free Software Foundation; either version 2 of the License, or
+ *  the Free Software Foundation; either version 3 of the License, or
  *  (at your option) any later version.
  *
  *  The GNU General Public License can be found at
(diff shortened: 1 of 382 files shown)


dkd-dobberkau:
The GPL (GNU General Public License) versions 2 and 3 do have some significant compatibility issues. Here's what you should know about the conflicts between them:

## Key Conflicts Between GPL2 and GPL3

1. **One-way Compatibility**: Code licensed under GPL2 cannot be included in a GPL3 project unless it's specifically licensed as "GPL2 or later" (GPL2+). However, GPL3 code cannot be included in GPL2 projects at all.

2. **Tivoization Protection**: GPL3 added provisions against "tivoization" (named after TiVo's practices), which prevents hardware manufacturers from using technical measures to block modified versions of the software from running on their hardware. GPL2 lacks this protection.

3. **Patent Grants**: GPL3 includes explicit patent grants, while GPL2 has no explicit patent provisions.

4. **Digital Rights Management (DRM)**: GPL3 has specific clauses addressing DRM and user rights that aren't in GPL2.

5. **License Compatibility**: GPL3 was designed to be compatible with more licenses than GPL2, including the Apache License 2.0.

## Practical Implications

If you're working with both GPL2 and GPL3 code:
- You can incorporate GPL2+ code into GPL3 projects
- You cannot incorporate GPL3 code into GPL2 projects
- Software specifically licensed as "GPL2 only" cannot be combined with GPL3 code

This can create challenges when attempting to combine code from different open source projects that use different GPL versions.

If you're dealing with a specific licensing situation involving GPL2 and GPL3 code, consulting with a legal professional who specializes in open source licensing would be advisable.

dkd-dobberkau:
We will have a closer look at it. Thanks for your heads-up.

Issue #1794, opened 2017-12-28
Install solr doesn't support Java version 9


griebie:
The install-solr.sh script running under Java9 results in this error:

ERROR parent directory: (/opt) of install path (/opt/solr) is not writeable.
You have installed Java version 0. Please install Java 8 or newer.
Please install all missing requirements or fix any other errors listed above and try again.

mbresch-sf:
Will provide a patch soon.

mbresch-sf:
Java changes his version number to a "normal" one. ;-)

Now Java 9 has the version "9.0.1" instead of "1.9.0_1".

see https://blogs.oracle.com/java-platform-group/a-new-jdk-9-version-string-scheme for more informations

mbresch-sf:
Hi @griebie ,
is this patch working for you?

PR #1795, opened 2017-12-28
[BUGFIX] java 9 support

mbresch-sf:
now the old and the new version string schema are supported

Fixes: #1794


mbresch-sf:
Tested the install-script with Java 8.

Issue #1812, opened 2018-01-23
[BUG] SOLR_RELATION does not recognize sys_categories for translated pages


dkd-kaehm:
By trying to assign related sys_categories to solr-documents like this:

```
plugin.tx_solr.index.queue.pages.fields {
  someSubCategory_stringM = SOLR_RELATION
  someSubCategory_stringM {
    localField = categories
    foreignLabelField = title
    enableRecursiveValueResolution = 1
    multiValue = 1
    additionalWhereClause = parent=14
  }
}
```
the fields are added only on default language, all the translated pages don't have the desired field.  

[ApacheSolrForTypo3\Solr\ContentObject\Relation::getRelatedItems()](https://github.com/TYPO3-Solr/ext-solr/blob/2bca3b9064f744ae4d562c38cb2806269a2d7f2e/Classes/ContentObject/Relation.php#L132-L136)

```
...
        // pages has a special overlay table constriction
        // @todo this check can be removed when TYPO3 8 support is dropped since pages translations are in pages then as well
        if ($GLOBALS['TSFE']->sys_language_uid > 0 && $localTableName === 'pages' && Util::getIsTYPO3VersionBelow9()) {
            $localTableName = 'pages_language_overlay';
        }
...
```

**Expected behaviour** is that SOLR_RELATION fallbacks to `pages` table if `pages_language_overlay`does not have TCA configuration for `localField`.

PR #1816, opened 2018-01-26
[BUGFIX] fallback to pages on SOLR_RELATION on missing TCA for pages_language_overlay

dkd-kaehm:
missing column settings in TCA for pages_language_overlay must be fetched from pages TCA

Fixes: #1812

### DIFF
--- a/Classes/ContentObject/Relation.php
+++ b/Classes/ContentObject/Relation.php
@@ -122,35 +122,58 @@ class Relation
     protected function getRelatedItems(
         ContentObjectRenderer $parentContentObject
     ) {
-        $relatedItems = [];
 
-        list($localTableName, $localRecordUid) = explode(':',
-            $parentContentObject->currentRecord);
-
-        $localTableNameOrg = $localTableName;
-
-        // pages has a special overlay table constriction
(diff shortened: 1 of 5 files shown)


timohund:
@dkd-kaehm This fix, breaks another testcase, we need to check this as well.

Classes/ContentObject/Relation.php
```diff
@@ -122,37 +122,59 @@ public function cObjGetSingleExt(
...
         }
 
         return $relatedItems;
     }
 
+    /**
+     * @param string $localTableName
+     * @todo this can be removed when TYPO3 8 support is dropped since pages translations are in pages then as well
+     */
+    protected function usePagesLanguageOverlayInsteadOfPagesIfNeeded(string &$localTableName)
```
timohund: Why do you pass a reference here for a string instead of returning the valu and reassign it in the calling method?

dkd-kaehm:
There is another broken Scenario without Test coverage:

Scenario: Page is translated and category is assigned only to original page.
Actual behaviour:
Solr document for translated page has no defined field.
Expected behaviour: 
Solr document for translated page has the defined field.

I'm working on that.

PR #1822, opened 2018-01-31
[TASK] Allow using complex filter values introduced by round brackets

lgescobar:
By not escaping filter values starting with double quotes you can
define complex filter conditions also in request parameters.

I needed this function for a project and thought it could be useful for anybody else.

This is a very naive approach and I don't know if the filter values have to be checked or escaped again.

### DIFF
--- a/Classes/Query/Modifier/Faceting.php
+++ b/Classes/Query/Modifier/Faceting.php
@@ -206,7 +206,16 @@ class Faceting implements Modifier
                             $filterOptions);
                         $filterParts[] = $facetConfiguration['field'] . ':' . $filterValue;
                     } else {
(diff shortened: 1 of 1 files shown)


timohund:
@lgescobar Thanks for the pr. I think we should add a testcase for that and also merge it into master to have it available in latest versions as well. Could you proceed on that?

lgescobar:
I'll give it a try!

echavaillaz:
I think it's a good idea, you have to add code also in the ConfigurationService to allow complex clause on flexForm filters (the method is "getFilterFromFlexForm".

dkd-kaehm:
@echavaillaz @lgescobar 
This pull request is outdated, the sources are changed, so the rebase or cherry-pick are not possible. 
Could you please provide the new implementation for main branch? 

I'll close this pull request in favor of new issue: #3643

Issue #1823, opened 2018-01-31
[TASK] Migrate testing framework to nimut/testing-framework 3.x


timohund:
We should migrate the testing framework to version 3.x and enable the tests for TYPO3 9.0

PR #1824, opened 2018-01-31
[TASK] Update nimut/testing-framework to 3.x and allow to install with TYPO3 9.1

timohund:
This pr:

* Updates the testing framework to 3.x
* Fixes issues in the tests that are related to the update
* Use inject* Methods inject of injection Annotations since this is a compatible way for TYPO3 8 and 9
* Updates scrutinizer configuration to allow to exclude the typo3/cms-install package

Fixes: #1823

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -38,6 +38,12 @@ tools:
     runs: 2
     timeout: 2400
 
+checks:
+    php:
+        excluded_dependencies:
+            - typo3/cms-install
+        avoid_superglobals: false
+
 build_failure_conditions:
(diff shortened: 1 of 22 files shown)


.scrutinizer.yml
```diff
@@ -38,10 +38,24 @@ tools:
     runs: 2
     timeout: 2400
 
+checks:
+    php:
+        excluded_dependencies:
```
dkd-kaehm: 4 spaces instead of 2 used
below on build key to

Works on scrutinizer but some IDEs could begin to moan about

Issue #1658, opened 2017-09-21
[TASK] Improve instanciation of TSFE in initializeTsfe


timohund:
When we currently instanciate the TSFE in Util::initializeTsfe there is a problem with the language handling:

* When the typoscipt configuration is set to "config.sys_language_mode" strict and we intialize the TSFE for a language when no translation exists the 404 error handling kicks in and does a redirect.

To avoid that and have control over the error handling we should instanciate an own class that overrides the methods for the 404 handling and throws an exception instead. This makes it possible to handle this error and show an error message or re-initialize with the default language when this makes sence.

timohund:
PR #873 is related to that. We should add test cases for that as well.

timohund:
We need to refine the patch since it breaks the indexing. I think we can not use OverriddenTypoScriptFrontendController as typehint everywhere since it is not allways in our scope to create it.

PR #1826, opened 2018-02-05
[TASK] do not die on indexing process

dkd-kaehm:
add IndexingException, DocumentPreparationException to handle errors from TYPO3 CMS side properly
add OverriddenTypoScriptFrontendController extended from CMS::TypoScriptFrontendController
use allways OverriddenTypoScriptFrontendController instead of CMS::TypoScriptFrontendController
set constant EXT_SOLR_INDEXING_CONTEXT on __construct of IndexService to handle indexing process properly

Fixes: #1658


dkd-kaehm:
original TypoScriptFrontendController is replaced everywhere, because some classes(Classes and Tests) on Search context as well on Indexing context using same super class and the methods from super classes.
This does not possible to use original TSFE class for Searching context.

PR #1829, opened 2018-02-08
SolrConnection should have a string representation

bmack:
During to the `SolrService/SolrConnection` refactoring, the method `__toString()` is missing, but still called in 
`Classes/IndexQueue/FrontendHelper/PageIndexer.php` (line 303) causing in Fatal (but recoverable) PHP Errors while indexing.

The patch adds the missing method again, which works just fine for our use cases (TYPO3 v9, solr-master)

### DIFF
--- a/Classes/System/Solr/SolrConnection.php
+++ b/Classes/System/Solr/SolrConnection.php
@@ -222,4 +222,15 @@ class SolrConnection
 
         return $service;
     }
+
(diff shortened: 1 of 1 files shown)


timohund:
@bmack thanks for the pr, i think the failing test is not related to your commit, i'll have a look on it and merge it afterwards.

PR #1830, opened 2018-02-08
[TASK] Use Guzzle API for HTTP Crawling Requests

bmack:
Use TYPO3 v8's included crawling functionality with Guzzle to support cURL-based requests.


timohund:
Thanks for the patch, we should only make d2aedfa525103bfc48cec5af9c2aa200902a79fd part of this pr and check why the tests fail and fix them.

timohund:
@benni i'll close this pr and open a new one with your commit for guzzle. Thanks for the pr.

PR #1834, opened 2018-02-13
[DOCS] Add example about custom indexing

georgringer:
Add's a documentation, how to add documents to solr with custom php code

### DIFF
--- a/Documentation/Development/Indexing.rst
+++ b/Documentation/Development/Indexing.rst
@@ -83,6 +83,20 @@ If external data should be indexed or the RecordIndexer is not required, it is p
            $connection = $this->connectionManager->getConnectionByPageId($pageId, $language);
            $connection->addDocuments($documents);
        }
+       
+       
+       /**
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
Hi, 
I would recommend to mention the consistency. Everything can be indexed, no matter from which sources, but index must be consistent with real data.

Please add some scenarios to keep index consistent:

* purging all documents added this way and index all again from beginning  
or
* collect garbage non-existent docs from index, requires comparison of index and source

or something similar

timohund:
@georgringer Thanks for your patch
@dkd-kaehm Can you please add to the pr what you miss?

In general i would propose to extend ApacheSolrForTypo3\Solr\Domain\Search\ApacheSolrDocument\Builder to reduce the required code to create an own document

georgringer:
I have some code already for that. I would still prefer to be a bit lowlevel but of course you know it better than me!

PR #1850, opened 2018-02-20
[TASK] Catch NoPidException in RecordMonitor

georgringer:
To avoid exceptions shown to the user, this exection
must be caught.


timohund:
@georgringer i think we should at least log the exception. What did you do to see it in the backend?

georgringer:
I am doing some fancy stuff which deletes a record after persisting it using a DataHandler hook, therefore it is not possible for solr anymore to fetch this record. 

I saw the uncaught exception.

PR #1875, opened 2018-03-06
[BUGFIX] Ensure AjaxController returns a response

bmack:
In TYPO3 v9, all Routes should return a PSR-7 compatible Response object.

However "updateConnections"  does not do so, resulting in a PHP error.

Also valid for EXT:solr v8.0.1


timohund:
@bmack Thanks for the patch

Issue #1887, opened 2018-03-09
Suggests autocomplete box is attached to form in any case, results in a bad position


Mabahe:
`$searchBox = $form.closest('.tx-solr-suggest');` will not give a result because `closest` is traversing up through form ancestors, but the input is in the descendants, so it fallbacks to the form element in any case.
With the right selector `.find` the box never shows up, so the setting `appendTo: $searchBox,` is not a good approach here.

Additionally I suggest to use $searchBox.outerWidth() to not have paddings minimizing the width.

PR #1888, opened 2018-03-09
[BUGFIX] Show suggests box next to query input field

Mabahe:
Resolves: #1887


timohund:
@Mabahe Thanks for the patch

PR #1895, opened 2018-03-14
[BUGFIX] prevent CURLE_BAD_CONTENT_ENCODING

mtelgkamp:
When using an nginx server and libcurl 7.57.0 indexing did not work with libcurl error 61. This was because Content-Encoding "gzip" was requested but "normal" was returned. To prevent this I changed the header to send a Content-Type instead of Content-Encoding. This also works with apache. It should also work behind Varnish, but this needs to be tested.


timohund:
@mtelgkamp Thanks for your patch, i'll have a look on it.

Issue #1906, opened 2018-03-19
Dupe id "solr-pagination" with both pagebrowsers


Mabahe:
When the pagination is included twice the id is defined twice. Therefore a class should be used here.

PR #1907, opened 2018-03-19
[BUGFIX] Duplicate id attribute solr-pagination

Mabahe:
Resolves: #1906


timohund:
@Mabahe Thanks for the patch

Issue #1931, opened 2018-04-04
[TASK] Ignore wrong Arguments of GeneralUtility::makeInstance in scrutinizer


timohund:
There are several errors in scrutinizer caused by the passed arguments to GeneralUtitlity::makeInstance, we should ignore them since this is only related to the annotation and a false positiv.

PR #1932, opened 2018-04-04
[TASK] Ignore wrong arguments of GeneralUtility::makeInstance in scrutinizer

timohund:
This pr:

* Add's an ignore annotation to all calls to GeneralUtility::makeInstance that is passing arguments.

Fixes: #1931


dkd-kaehm:
would be nice if something like regex rules for methods in .scrutinizer.yml could be specified.
Shortly showed in the docs from scrutinizer, but nothing like this found

PR #1944, opened 2018-04-09
Add datetime to tx_news indexer

### LINKED ISSUE #1943: [BUG] Missing datetime index field in news examples
When using the index field examples for tx_news, the field datetime is missing.

spoonerWeb:
Two options:

* datetime as a string
* datetime as a date field

Closes #1943


timohund:
Thanks for the patch

Issue #1974, opened 2018-05-03
[TASK] RecordMonitor should log a warning instead of throwing an Exception when a record without pid is processed.


timohund:
Currently a NoPidException is thrown, when a record is processed without a valid pid. To allow that rare use case we should catch the Exception and create a log message with a warning instead.

PR #1975, opened 2018-05-03
[TASK] Catch NoPidException in RecordMonitor

timohund:
To avoid exceptions shown to the user, this exception must be caught and logged.

Fixes: #1974


timohund:
@georgringer I've made some adaptions to the patch. Can you check if this works as expected for you?

georgringer:
@timohund works

Issue #1984, opened 2018-05-04
Facet and sorting urls missing present piVars


Mabahe:
Since version 7 present piVars are not added to the facet and sorting urls again.

Till version 6.5 the Query LinkBuilder adds all prefix (default: tx_solr) GPvars to the build url https://github.com/TYPO3-Solr/ext-solr/blob/release-6.5.x/Classes/Query/LinkBuilder.php#L209

PR #1985, opened 2018-05-04
[BUGFIX] Add setting to be able to add additional persistent arguments

Mabahe:
Till solr version 6.5.x all parameters of the plugin namespace was
added to the facet or sorting url again.
This patch adds a setting to be able to enable this behavior again,
but only with a whitelist of argument names.

Resolves: #1984


Mabahe:
You are welcome for adding test. Thx for taking care

PR #2001, opened 2018-06-07
Removes broken reference to old EID suggest script

alappe:
Closes: #2000


irnnr:
we can also remove the comment above the removed line

Issue #1997, opened 2018-05-29
[BUG] No suggestions are shown for one result


jagodzinski:
Hi,
when only one result is given for a suggestion no results are shown because of this lines:

solr/Resources/Public/JavaScript/suggest_controller.js:76-80
                    // 1 result means not result for given input; then we don't show is as suggestion
                    // also prevents autosubmit when selecting input box
                    if (**result.suggestions.length <= 1**) {
                        result.suggestions = [];
                    }

When I change it to result.suggestions.length < 1 the suggestions are correct, but it triggers the form submit if you type the whole word (like it's written in the comment).

In older versions it's working as expected. 

We use **solr 8.0.3**

Kind regards
Jessica

stefanisak:
Try to remove this condition: https://github.com/TYPO3-Solr/ext-solr/blob/74d5531db90b4c46ccf65b05855e693578a37e0b/Resources/Public/JavaScript/suggest_controller.js#L78

and add `triggerSelectOnValidInput: false,` option e.g. here:
https://github.com/TYPO3-Solr/ext-solr/blob/74d5531db90b4c46ccf65b05855e693578a37e0b/Resources/Public/JavaScript/suggest_controller.js#L32

Now you should see all results without infinite submit loop.

Maybe this helps...

PR #2008, opened 2018-06-14
[BUGFIX] No suggestions are shown when only one suggestion available

timohund:
This pr:

* Remove the condition
* Add's the configuration triggerSelectOnValidInput: false,

Fixes: #1997


timohund:
@stefanisak @ jagodzinski can you please check if this patch is working in your installation? Then we can merge it to master.

Issue #1993, opened 2018-05-07
[BUG] Page not found leads to index error and timeout


elmarputz:
Hi, 
when indexing a website (either via scheduler or via manual indexer) with ext:solr, the indexer hangs if a page url returns a "not found" or "unavailable". This causes the PHP process to hang until it reaches the script timeout. The scheduler then shows the "running" label on the indexing process forever. The only fix i found until now is to remove the invalid ULRs from my project (by searching the Log entries). 

This is an example log entry which causes the behaviour: 

`Core: Exception handler (WEB): Uncaught TYPO3 Exception: #1301648404: This page (ID 6114) is of type "Shortcut" and configured to redirect to a page, which is not accessible (ID 6112). | TYPO3\CMS\Core\Error\Http\PageNotFoundException thrown in file /MY_PATH/typo3_src/typo3_src-7.6.27/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php in line 1723. Requested URL: https://mydomain.at/my-url/`

- TYPO3 Version: **7.6.27**
- ext-solr: **6.5.1**

Elmar

PR #2010, opened 2018-06-15
[TASK] Do not die on indexing process

timohund:
* add IndexingException, DocumentPreparationException to handle errors on Indexing process properly
* add OverriddenTypoScriptFrontendController extended from CMS::TypoScriptFrontendController
** use it in Util and in ext_localconf.php if Frontend-Helping things are triggered
** define EXT_SOLR_INDEXING_CONTEXT on indexing processes and handle then differently
** on __construct of IndexService and in ext_localconf
* add test case with dieing part of TYPO3 CMS for pages indexing

Fixes: #1658 
Fixes: #1993

### DIFF
--- a/Classes/Controller/Backend/Search/IndexQueueModuleController.php
+++ b/Classes/Controller/Backend/Search/IndexQueueModuleController.php
@@ -249,6 +249,8 @@ class IndexQueueModuleController extends AbstractModuleController
     /**
      * Indexes a few documents with the index service.
      * @return void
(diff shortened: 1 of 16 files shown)


timohund:
@dkd-kaehm i took the last state of the patch, that we remove and changed the typohints back from OverriddenTypoScriptFrontendController to TypoScriptFrontendController, we should test that patch extensive before we merge it.

timohund:
@elmarputz this should also fix the issue that you described in #1993, do you maybe have the time to test that?

Issue #1968, opened 2018-04-25
[BUG] Sorting item option fixedOrder broken since fluid rewrite


Mabahe:
Documentation: https://docs.typo3.org/typo3cms/extensions/solr/Configuration/Reference/TxSolrSearch.html#sorting-options-optionname-fixedorder

timohund:
@Mabahe this can be achieved by adapting the partial for one particular sorting. I would propose to move that option from the documentation.

Mabahe:
I'm ok with this.

PR #2014, opened 2018-06-15
[TASK] Remove option fixedOrder

Mabahe:
This option has no effect since version 7 (fluid rewrite)

It can be achieved by adapting the partial for one particular sorting and or using the option defaultOrder.

Resolves: #1968


Mabahe:
@timohund could you squash the commits on merge with a better commit msg?

Issue #126, opened 2015-11-13
[RecordMonitor] Monitor changes through Extbase repositories


irnnr:
When records are updated through Extbase they are not detected by the record monitor as these changes do not go through TCEmain.

Use signals in Extbase repositories to detect these changes.

Mabahe:
I need this in my current project, so I will 	shortly implement and share this. Did you already have a detailed idea?

irnnr:
nope, patches welcome!

hhoechtl:
Feel free to generalize my solution: https://gist.github.com/hhoechtl/cd57e7b68e8fe6642165

Mabahe:
Cool, thx, this is a good start.

irnnr:
Linking Marc Bastian's WIP patch: https://github.com/Mabahe/ext-solr/commit/0c3d6db8fcec0410666f1c70ecca997096543ac2

Mabahe:
I see a reference above your comment.

irnnr:
Saw that too after I added the link :)

Mabahe:
Former patch versions could be found here:
Based on ext-solr 8.1.x (TYPO3 8.7), tested and running in production: https://github.com/Mabahe/ext-solr/tree/feature/release-8.1.x/126-extbase-domain-observer
Based on ext-solr 10.0.x, rebased by @dkd-kaehm, but not tested by me: https://github.com/Mabahe/ext-solr/tree/feature/release-10.0.x/126-extbase-domain-observer

PR #2035, opened 2018-07-06
[WIP][FEATURE] Observe changes through Extbase repositories

Mabahe:
Resolves: #126

### DIFF
--- a/Classes/IndexQueue/DomainObjectObserver.php
+++ b/Classes/IndexQueue/DomainObjectObserver.php
@@ -1,0 +1,302 @@
+<?php
+namespace ApacheSolrForTypo3\Solr\IndexQueue;
+
+/***************************************************************
+ *  Copyright notice
+ *
+ *  (c) 2018 Marc Bastian Heinrichs <mbh@mbh-software.de>
+ *  All rights reserved
+ *
+ *  This script is part of the TYPO3 project. The TYPO3 project is
+ *  free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 3 of the License, or
+ *  (at your option) any later version.
+ *
+ *  The GNU General Public License can be found at
+ *  http://www.gnu.org/copyleft/gpl.html.
+ *
+ *  This script is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  This copyright notice MUST APPEAR in all copies of the script!
+ ***************************************************************/
+
+use ApacheSolrForTypo3\Solr\Domain\Index\Queue\RecordMonitor\Helper\ConfigurationAwareRecordService;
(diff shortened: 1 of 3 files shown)


Mabahe:
If added the WIP patch as PR to make it more visible. Better description and docu is missing for the moment

timohund:
@Mabahe Thanks for sharing and getting this started. I think it would be nice to get it into version 9.0 of EXT:solr. Do you have the time to add documentation? In addition it would be good to have this tested with our automated tests.

dkd-kaehm:
Notice 4 myself : D16239

WebsiteDeveloper:
@Mabahe @timohund any info on the progress of this PR would really love this feature

WebsiteDeveloper:
@timohund I would love to contribute some Tests for this
Any pointers on where to start?

timohund:
@WebsiteDeveloper Nice to here that. 
A good starting point is: 

https://github.com/TYPO3-Solr/ext-solr/blob/master/CONTRIBUTING.md#automated-tests
Or the .travis.yml in our project.

It is also possible to run the tests in a local travis docker container. If you need further assistance we can also talk on slack and maybe create a script for that.

WebsiteDeveloper:
Alright i will check it out :)
Any thoughts on then backporting this feature to older versions?

timohund:
As long at is not a breaking change i think in general we could backport that.

dkd-kaehm:
@Mabahe 
Could you please refresh the things for dev-master(future 11) @ TYPO3 9.5 + 10.4?

dkd-kaehm:
Rebased to current master. We should complete this.

Classes/IndexQueue/DomainObjectObserver.php
```diff
@@ -0,0 +1,302 @@
...
+
+    /**
+     * Affects new and updated objects
+     *
+     * @param DomainObjectInterface $object
+     */
+    public function afterUpdateObject($object)
+    {
+        $forcedChangeTime = 0;
+        list($_, $object, $forcedChangeTime) = $this->signalSlotDispatcher->dispatch(__CLASS__, 'preUpdateCheck', [$this, $object, $forcedChangeTime]);
```
dmitryd, line 139: This should be:

```php
[ , $object, $forcedChangeTime] = ...
```

dmitryd:
What about handling `afterInsertObject` for new records?

Mabahe:
I'm on updating this.

Mabahe:
@dmitryd @varioous new records are also affected by the updated event: `backend::persistObjects` calls `persistObject` calls `updateObject`. This is because `insertObject` inserts only the row without the relations; relations are handled in `updateObject` later on. Also the `Queue::updateItem` does an insert, if the item is missing. So IMO no extra handling for new records is needed. If I remember right it worked for new objects, or do you have noticed problems with new objects?

dmitryd:
@Mabahe Cannot remember, unfortunately. I have no code to test this right now. :(

Mabahe:
I will drop my branch https://github.com/Mabahe/ext-solr/tree/feature/extbase-domain-observer because of the bad name (branch missing). Also this patch is not compatible with the base branch. 

So will close this PR in favour of a new one using the right branch name.

PR #2038, opened 2018-07-16
[FEATURE] Allow open query in the DateRangeFacet

Anubiso:
This change allows to use the DateRangeFacet with an open end or open begin like:

?tx_solr[q]=*&tx_solr[filter][0]=created:-201701010000
?tx_solr[q]=*&tx_solr[filter][0]=created:201701010000-

Fixes: #2041


timohund:
@Anubiso thanks for the patch and adding a test for that

Issue #2054, opened 2018-07-19
Indexer::index poisons TSFE when call in FE action


stuartmcfarlane:
After creating a record in my application I reindex the record and redirect to the edit action like so:

```
        ...
        $ruinMyTSFEService = GeneralUtility::makeInstance(
            \Poisinous\RuinMyTSFEService::class
        );
        $ruinMyTSFEService->doIt(
            'tx_extensionname_domain_model_modelname',
            $newModel->getUid()
        );
        $this->redirect('edit', null, null, [
            training => $newModel
        ]);
```

The reindex seems to work just fine however, the subsequent redirect is broken.

Here is the code for the RuinMyTSFEService:

```
<?php
namespace Poisinous;

use ApacheSolrForTypo3\Solr\IndexQueue\Indexer;
use ApacheSolrForTypo3\Solr\IndexQueue\Queue;
use TYPO3\CMS\Core\Utility\GeneralUtility;

class RuinMyTSFEService
{
    public function doIt($type, $uid) {
        $indexQueue = GeneralUtility::makeInstance(Queue::class);
        $indexer = GeneralUtility::makeInstance(Indexer::class);

        $indexQueue->updateItem($type, $uid);
        $items = $indexQueue->getItems($type, $uid);
        foreach ($items as $item) {
            $indexer->index($item);
        }
    }
}
```

timohund:
@stuartmcfarlane What do you want to do with that code? And why do you need that?

stuartmcfarlane:
The records are changed in the FE with extbase:

```
public function createAction(MyModel $newModel)
{
    $this->myRepository->add($newModel);
    $this->persistenceManager->persistAll();

    $this->redirect('edit', null, null, [
        model => $newModel
    ]);
}
```

timohund:
I guess what you need is this: https://github.com/TYPO3-Solr/ext-solr/pull/2035 + continous processing of the index queue.

PR #2055, opened 2018-07-19
Site only Util::initializeTsfe's if TSFE is falsey

stuartmcfarlane:
This fixes the issue 'Indexer::index poisons TSFE when call in FE action #2054'


timohund:
@stuartmcfarlane Thanks for the patch. I think, that this patch has side effects, because you don't know if the existing intialize TSFE was initialized for the root page that was passed in that context. Therefore below the rootPageId is passed. I think this need extensive testing.

stuartmcfarlane:
Oops, sorry, this is not a solution

Issue #2088, opened 2018-08-15
Wrong root_pid saved in statistics table


mschwemer:
The root_pid, which is saved in the statistics table is always set to uid of the level "0" of the rootline of the search page. 

EXT:solr/Classes/Domain/Search/Statistics/StatisticsWriterProcessor.php ... line 80

But there are scenarios, where this setting is not correct. 

Our scenario is, that we have a multi-tree language setup under one domain. The search configuration is available on the subpages. 

The domain and pid ist configured via the `$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['solr']['sites']['781']['domains']`. The pid provided here is the pid of the search configuration. 

Now, if I go to the statistics module on that page, the module says "Notthing found ... did you enable statistics". If I go to the root page, the statistics module sais "No search configuration found".

The solution is to determine the 'root_pid' depending on the configuration.

I will provide a pull request in the next couple of minutes.

mschwemer:
TBH, I am not sure, whether it is a solution for all scenarios ...  At least another one, I can imagine, is, when a sys_domain  record is available and the attribute "is_site_root" is set in the page properties. Usually this is only the case on a "parent" page, but this is also possilble way down in the page tree.

PR #2089, opened 2018-08-15
[BUGFIX] Wrong root pid in statistics table

mschwemer:
The root_pid in the statistcs table is always the root_pid on the search page. There are secenarios where this is not appropriate. This patch resolves the 'root_pid' according to the configuration.

Resolves: #2088


timohund:
@mschwemer Thanks for pointing out this and for preparing a patch. Did you try to use SiteRepository::getSiteByPageId, with the current id of TSFE? This should be less code and also work, when the sites are only configured with TypoScript.

mschwemer:
I did not know that :-)

I changed the patch and used that method. It seems to work.

timohund:
Thanks for adapting that. Regarding the failing test it think we should add the SiteRepository as dependency in the constructor and mock it for the unit test. I can also have a look on that in the next days.

timohund:
I've added your changes to #2092 thanks for the patch

PR #2133, opened 2018-09-11
[FIX] Add only items if index configuration name exists

birgerstoeckelmann:
Ext-solr is adding items to the queue, even if an indexing configuration for this record type doesn't exist. Some extensions like in2code's Publisher use the index queue to update or add items. In the case of the Publisher the record is added to the queue only by the table name e.g. "tt_content" or "sys_reference". All items are added to the queue which results in multiple indexing errors when Solr tries to index those items:

1455530112: RuntimeException: The item indexing configuration "tt_content" on root page uid 1 could not be found! in /srv/www/.../Web/typo3conf/ext/solr/Classes/IndexQueue/Indexer.php:284

So  the class ConfigurationAwareRecordService should not return the table name by default as a valid configuration name. And an item should not be added to the queue without a valid configuration.


timohund:
@birgerstoeckelmann 

Thx for your pull request, can you please:

* rebase your pull request
* remove the commented part
* move you new part into an elseif() instead of adding a new if only inside the else part

birgerstoeckelmann:
@timohund 

Hey Timo, could you please merge this pull request or do I have to do something else?
Will this be merged into the 8.1 version either?

Kind regards
Birger

timohund:
Can you please create a clean mergeable pr? Your pull request containts changes in 122! files

PR #2147, opened 2018-10-01
WIP [BUGFIX] MM relation value resolution

cjanody:
getRelatedItemsFromMMTable returns only the first matching item when using enableRecursiveValueResolution. This is due to the use of a return in the foreach loop instead of merging the result of getRelatedItems.

Sample:
ModelA has n:m relations to modelB and ModelB n:m relations to ModelC

```
plugin.tx_solr {
  index.queue {
    A = 1
    A {
      table = tx_sample_domain_model_modela
      fields {
        title = title
        modelC_stringM = SOLR_RELATION
        modelC_stringM {
          localField = modelb
          multiValue = 1
          foreignLabelField = modelc.title
          enableRecursiveValueResolution = 1
          removeDuplicateValues = 1
        }
      }
    }
  }
}
```

Whitout the patch, only the first title of linked ModelC is returned

Thanks for reading.


timohund:
I would really propose to add a functional test that fails without the patch and passes with the patch.

timohund:
@cjanody your last commit add's a fixture, but not a testcase. Will you add a testcase in addtion?

cjanody:
@timohund I will add the testcase as soon as possible. Thks

dkd-kaehm:
This Pull-Request is still relevant.
@cjanody Would you like to continue or pass it on to someone else?

cjanody:
@dkd-kaehm I can pass it. I haven't had time to work on for a while... Thks

Issue #1773, opened 2017-12-07
Cannot set collapseSingleValueArrays of SolrPhpClient with TS configuration


masi:
The default is to collapse arrays with a single item to a string. I find this rather annoying especially with Fluid's for view helper.

I cannot think of any good reason to want it that way, but what di I know. So a new configuration variable in the plugin's TypoScript setting to change the behaviour would be nice.

timohund:
@masi Are you talking about this change, or the old behaviour?: 

https://github.com/TYPO3-Solr/ext-solr/pull/1748

timohund:
As far as i understand this should be fixed with #1748 if you want, you can backport it to the 7.5 branch?

masi:
Yes, but I don't like the solution. See #1748 for my proposal.

masi:
I don't like the old behaviour, so the suggesion of a TS setting is only to provide backwards compatibility for whoever wants that.

Do you want  a patch for that?

PR #2172, opened 2018-11-07
[BUGFIX:Backport] Always return value as expected (#1748)

dkd-kaehm:
If a field is an array, an array should be returned within Solr
document. This is necessary for e.g. be able to iterate over a property
in a Fluid file. If returning either an array or a string, an exception
in Fluid is thrown because it expects an array.

Backported from: #1748
Resolves: #1773


dkd-kaehm:
Its breaking change, even it is a BUG it changes behaviour.
Must be discussed. 
Imho we need to implement a TypoScript Setting for that.

timohund:
Why not simply update to the fixed version? I agree that this is a breaking change, but the setting would also then be obsolete in later versions

Issue #2193, opened 2018-12-12
Feature: Use facet uri viewhelpers without facet object


Mabahe:
In connection with grouping it would be nice to be able to use the facet uri viewhelpers without a dedicated facet object, but with plain strings.
You could create now a facet item link (add, set, remove) somewhere else in the results view.
Example: 
Beside 
`<a class="facet solr-ajaxified" href="{s:uri.facet.setFacetItem(facet: facet, facetItem: option)}">{option.label}</a>` 
you could create a set link now with this vh arguments:
`<a class="facet solr-ajaxified" href="{s:uri.facet.setFacetItem(facetName: 'type', facetItemValue: groupItem.groupValue, resultSet: resultSet)}">show more</a>`

PR #2194, opened 2018-12-12
[FEATURE] More flexible facet uri viewhelpers with optional facet object

Mabahe:
You could create now a facet item link (add, set, remove) somewhere else
in the results view where no facet object is available.

Beside {s:uri.facet.setFacetItem(facet: facet, facetItem: option)}
you could create a set link now with this vh arguments:
{s:uri.facet.setFacetItem(facetName: 'type', facetItemValue: 'pages',
resultSet: resultSet)}

Resolves: #2191


Mabahe:
uh, wrong issue reference. Could you open #2191 again and close #2193, thx

PR #2210, opened 2019-01-08
[BUG] Added siteHash as parameter to deleteRecordInAllSolrConnections

thomashohn:
Added siteHash to deleteRecordInAllSolrConnections


timohund:
Thx for the patch

PR #2216, opened 2019-01-16
[BUGFIX] Fix #2211: page browser links do not work in TYPO3 9.5 due to incorrect cHash

### LINKED ISSUE #2211: [BUG] TYPO3 9 - Wrong cHash on site using SiteConfiguration
On TYPO3 9, within a site using SiteConfiguration, all pagination or facet links are appended with a cHash.

```
https://mysite.local/global-search?tx_solr%5Bpage%5D=2&cHash=2083113c1e281718add2e591b3c7bf0f
```

This is new behaviour on TYPO3 9 introduced by the new routing handler. (see https://docs.typo3.org/typo3cms/extensions/core/Changelog/9.5/Feature-86365-RoutingEnhancersAndAspects.html#impact

The issue is also described on the forge: https://forge.typo3.org/issues/87120
It is clearly stated that useCacheHash=0 is now useless when generating a uri!

I open this issue in order to track the problem into milestone 9.
I know that siteHandling is not fully supported yet in dev-master, but this issue is very anoying, because everyone want to have pretty url, and we can not acheive this on v9 without a SiteConfiguration.

dmitryd:
TYPO3 9.5 now requires cHash on all links and appends it even if asked not to generate it. Thus the old way to generate links in EXT:solr does not work anymore. Links has to be generated for each set of arguments properly but still can be cached to be reused within the same request. This PR does exactly this.

### DIFF
--- a/Classes/Domain/Search/Uri/SearchUriBuilder.php
+++ b/Classes/Domain/Search/Uri/SearchUriBuilder.php
@@ -309,73 +309,18 @@ class SearchUriBuilder
      */
     protected function buildLinkWithInMemoryCache($pageUid, array $arguments)
     {
-        $values = [];
-        $structure = $arguments;
-        $this->getSubstitution($structure, $values);
-        $hash = md5($pageUid . json_encode($structure));
+        $hash = md5($pageUid . '|' . json_encode($arguments));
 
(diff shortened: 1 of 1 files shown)


timohund:
Thanks for the patch, i used that as a base for #2222

Issue #2217, opened 2019-01-16
Search form should not have id and L


dmitryd:
Form url is generated by a special viewhelper but `id` and `L` fields are still submitted with the search form. They can be safely removed without any problems to functionality.

Also `/de/suche/?tx_solr[q]=lorem` looks better than `/de/suche/?id=23&L=0&tx_solr[q]=lorem`.

n3amil:
i can confirm this is working. I had an issue too with the "L" parameter resulting in never ending 307 redirects where a "/" gets appended to the request url every new redirect. Resulting in something like this:
http://my.example.domain/en/search////////////////////?id=145&L=0 and "to many redirects" error in the browser.  
I replaced the "L" parameter with a lowercase "l" and it worked. (which is kinda interesting ;) )
But as @dmitryd  said, removing both the L and id parameter is working just fine for the form.

timohund:
I don't think that the id and L parameter can savely be removed, at least not now. When you don't have any speaking urls configured and as action "/index.php?id=4711&L=1" the id and L parameter get lost, when the form is submitted. This can be the case e.g. in an 8.7 LTS system that e.g. does not use any speaking urls.  Any proposals @Mabahe @dmitryd @n3amil

dmitryd:
Why do you think they will be lost? The action URL will always include them. Otherwise you would not have a proper speaking url in the form action parameter.

dmitryd:
Oh, I [got it](url
http://www.w3.org/TR/2011/WD-html5-20110525/association-of-controls-and-forms.html#form-submission-algorithm). The form has to be submitted by GET method and without speaking urls. In that case, yes, it will disappear.

Is that a real case? Does anybody run a web site with solr but without speaking urls?

timohund:
@dmitryd Thanks for your feedback. Here are migh thougs about it:

It is the default case when you install TYPO3 8 LTS + solr without any modifications. With the merged patch these arguments will not be added when you use the site management. Otherwise they get added but you can remove them in your custom partial if you need it. With realurl enabled, it's still working but just redundant.

I would say it should work in a default installation "outofthebox" and who is able to configure TYPO3 with realurl should also be able to remove that stuff from the default partial.

It could be optimized to have a more generic condition like "isSpeakingUrlUsed" and then not add the hidden parameters. Do you have a proposal for a chance or are you fine with the current status in 9 and master?

dmitryd:
[...]

PR #2218, opened 2019-01-16
[TASK] pretify search url by removing id and L from it

dmitryd:
These parameters are not needed in the form because they are handled by the URL.

Fixes: #2217


timohund:
Will be fixed with #2224

PR #2219, opened 2019-01-21
[Fix] add missing parameter 'method'

svwerder:
Correction of the missing passing of the "method" parameter


timohund:
@svwerder Thx for the pr. For what do you need that? Are you using the php api? If you create a patch like this it would be nice to implement that for the master branch and then backport it. Otherwise this change would be lost for newer versions. Can prepare a patch for master as well please?

PR #2221, opened 2019-01-28
[BUGFIX] Fix backend info module with missing core

### LINKED ISSUE #2220: Info module throws exception if solr cores not available for one language
In the info module the connection to all configured solr cores for the different languages is checked. If one core does not exist on SOLR server an exception is thrown. But this exception has to be catched by the info module to show, which solr cores are available and which not.

Currently this exception (SolrCommunicationException) is not catched. This crashes the info module if just one core is not available.

Found with TYPO3 8.7.x and ext:solr version 8.1.2

maritwho:
The Exception NoSolrConnectionFoundException is never thrown
in \ApacheSolrForTypo3\Solr\Search::search(), but the
SolrCommunicationException. This exception has to be catched
instead, otherwise the info module in backend crashes.

Fixes #2220


maritwho:
Ok, I see I have to adapt the tests first. Will do the next few days.

timohund:
Fixed with #2231 we need to catch both exceptions in the DocumentRepository

PR #2236, opened 2019-02-04
[TASK] Consider external and internal news URLs

fnagel:
Adding a `CASE` structure in order to use the `externalurl` and `internalurl` fields if needed. With this change, Solr for TYPO3 will not generate links to empty news detail views which only exist in the search results.

Related Slack discussion: https://typo3.slack.com/archives/C02FF05Q4/p1548323535173700 and ff


timohund:
@fnagel Thanks for sharing this and for the pr

fnagel:
Ahh, I have to thank you guys for maintaining this extension <3

Issue #2234, opened 2019-02-01
[TASK] When i use site management with TYPO3 9 in want to use the route of the page for indexing


timohund:
Currently the indexing url is build in the "PageIndexer" by building an url with own code. In this scenario the scheme, host and path if used and index.php is called with the id, L and MP parameter

When the Site Management is active we do not want to use the id and L parameter anymore. With the possibility of the site management and the router, we are able to build a proper speaking frontend url that can be used for indexing and contains the pageId and language information in the encoded path.

PR #2237, opened 2019-02-04
[TASK] Prepare different url generation when site management is active

timohund:
When the site management in TYPO3 9 is used, the language and pageId parameter are part of the encoded url. In previous versions of EXT:solr we build the frontend url by our own by adding the id, L and MP parameter.

This pr:
    
* Refactors the url generation by extracting the code into strategy classes
* Implements a SolrSiteStrategy that is able to build the indexing uri as before with index.php?id=xx&L=yy
* Implements a TYPO3SiteStrategy that uses the router to build the indexing uri
* Prepares a different indexing url handling with site management
    
**Important**: By now the old urls with id and L parameter are used also when site management is active since the urls that are generated by the router could produce urls that lead to a 404 error.
    
Related: #2234

### DIFF
--- a/Classes/Domain/Index/PageIndexer/Helper/UriBuilder/AbstractUriStrategy.php
+++ b/Classes/Domain/Index/PageIndexer/Helper/UriBuilder/AbstractUriStrategy.php
@@ -1,0 +1,156 @@
+<?php declare(strict_types = 1);
+
+namespace ApacheSolrForTypo3\Solr\Domain\Index\PageIndexer\Helper\UriBuilder;
+
+/***************************************************************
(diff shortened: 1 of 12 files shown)


timohund:
Using this can produce url's that lead to a 404. We need to clarify how this should be resolved (see also https://forge.typo3.org/issues/86595)

timohund:
After a discussion we decided to keep the refactoring and preparations from this pr, but by now allways return the old uri strategy with id and L param. Since site urls by now do not take fallbacks into account. When this is the case in the TYPO3 core, we can use the site urls for indexing as well.

Issue #2244, opened 2019-02-08
[BUG] KeepAllFacetsOnSelection is not evaluated when keepAllOptionsOnSelection is used


timohund:
**Describe the bug**
When a combination of keepAllFacetsOnSelection and keepAllOptionsOnSelection is used i would expect that both get evaluated. By now only the "keepAllOptionsOnSelection" is evaluated.

**To Reproduce**
1. Activate "keepAllFacetsOnSelection"
2. Activate "keepAllOptionsOnSelection"

**Expected behavior**

"keepAllFacetsOnSelection" should allways be evaluated and all facets and options should stay present.

 - EXT:solr Version: dev-master
 - Used Apache Solr Version: 7.6
 - PHP Version: 7.2

PR #2245, opened 2019-02-08
[BUGFIX] KeepAllFacetsOnSelection is not evaluated when KeepAllOptionsOnSelection is used.

timohund:
# What this pr does

Add's the excludeTags for both cases. This is required when both is activated at the same time.

# How to test

Enable both at the same time and check that the options and counts stay for the facets when you select an option.

Fixes: #2244


Mabahe:
No negative effect in two tested projects

Issue #2195, opened 2018-12-17
Feature: individual exclude tags for facets


Mabahe:
Support individual tagging and excluding filters, like described in https://lucene.apache.org/solr/guide/6_6/faceting.html#Faceting-LocalParametersforFaceting

It's already supported to tag filter queries from the settings, but one would need a setting to exclude this tag for selected facets.

mschwemer:
Hi,

thanks for the PR. It works very well. 

Greets
Marcus

dkd-kaehm:
leads on http://yonik.com/multi-select-faceting/
As example the facets with `plugin.tx_solr.search.faceting.facets.[facetName].operator = OR` all not selected options on used facet are always 0 currently.

dkd issue: 109334

timohund:
I think we have two issues here:

1. There seem to be a bug in the excludeTag creation if you have a mix of keepAllFacetsOnSelection and keepAllOptionsOnSelection. KeepAllFacetsOnSelection is not evaluated when keepAllOptionsOnSelection is set for a single facet. I'll create a new issue for that.

2. Beside that this pr proposes to introduce individuall exclude tags on a facet base. I think this is a very high level setting but it would be nice to support that. Since we allready generate exlucdeTags in ext solr i would propose to name the setting "addtionalExcludeTags" this makes more clear that there are allready other exclude tags in place.

I would handle the second part as intension and feature of this ticket and adapt @Mabahe pull request -

PR #2246, opened 2019-02-11
[FEATURE] Individual additional exclude tags for facets

timohund:
# What this pr does

* Add's the setting ```additionalExcludeTags``` to a facet to exclude custom tags from the count of that facet
* Add's the setting ```addFieldAsTag``` to a facet to force the creation of a tag for that facet

# How to test

Configure ```addFieldAsTag``` for the tagged facet and add that field with ```addtionalExcludeTags``` to a facet and see if the exclude tag will be applied

Fixes: #2195


Mabahe:
Thx for the docu and tests

PR #2276, opened 2019-02-28
✨ DDEV Configuration

WebsiteDeveloper:
This PR adds DDEV config to ease local development

See: https://ddev.readthedocs.io

### DIFF
--- a/.ddev/.gitignore
+++ b/.ddev/.gitignore
@@ -1,0 +1,8 @@
+#ddev-generated: Automatically generated ddev .gitignore.
+
+/import.yaml
+/docker-compose.yaml
+/db_snapshots
+/sequelpro.spf
+/import-db
+/.bgsync*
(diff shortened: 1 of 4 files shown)


timohund:
@WebsiteDeveloper Thanks for your PR, we allready have a ddev environment in a seperate repository: https://github.com/TYPO3-Solr/solr-ddev-site do you know that?

I would propose to have that in a seperate repository because:

* With that you have a complete running site with the introduction package.
* We can create new versions with different PHP/TYPO3 versions whatever

WebsiteDeveloper:
I didn't know that 
I still think that it provides some value for developing the extension itself. Which is the main usecase this intends to solve.

timohund:
@WebsiteDeveloper Ah ok, can you the please provide a snipped for the documentation how that can be used?

WebsiteDeveloper:
sure will do :)

dkd-kaehm:
Would propose to add the hint about https://github.com/TYPO3-Solr/solr-ddev-site existence and its usage instead of billow out this Repo. ddev is not а hub of the universe.

Issue #2279, opened 2019-03-01
[BUG] Exclude logger from serialize on save for scheduler task


achimfritz:
logFilePath is saved in Database

same as described in
s. https://forge.typo3.org/issues/86785 (and related issues

**Used versions (please complete the following information):**
 - TYPO3 Version: [e.g. ^9.5]
 - EXT:solr Version: [e.g. ^9.0]

PR #2280, opened 2019-03-01
[BUGFIX] Exclude logger from serialize on save for scheduler task

achimfritz:
Exclusion of the logger instance from the record during save to database.
Prevent hard-coded logfile paths in database record.
Set the logger at runtime if the task is executed.

Fixes: #2279

### DIFF
--- a/Classes/Task/AbstractSolrTask.php
+++ b/Classes/Task/AbstractSolrTask.php
@@ -94,8 +94,8 @@ abstract class AbstractSolrTask extends AbstractTask {
     public function __sleep()
     {
         $properties = get_object_vars($this);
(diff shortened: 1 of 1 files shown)


Classes/Task/AbstractSolrTask.php
```diff
@@ -94,8 +94,8 @@ public function getSite()
     public function __sleep()
     {
         $properties = get_object_vars($this);
-        // avoid serialization if the site object
-        unset($properties['site']);
+        // avoid serialization if the site and logger object
```
WebsiteDeveloper: Maybe you can correct the typo "if" too :)

Issue #2286, opened 2019-03-12
[BUG] Deprecated method getCoreName is still used in IndexAdministationModuleController


timohund:
**Describe the bug**

The deprecated method getCoreName is still used in IndexAdministationModuleController

**Expected behavior**

Method usage should be replaced with getCorePath if it makes sence.

**Used versions (please complete the following information):**

 - TYPO3 Version: 9.5.5
 - EXT:solr Version: 9.0.2
 - Used Apache Solr Version: 7.6.0
 - PHP Version: 7.2

**Additional context**
Automated test should be added since this issue could be detected with that.

PR #2288, opened 2019-03-12
[BUGFIX] Remove usage of deprecated method getCoreName in IndexAdministrationModuleController

timohund:
# What this pr does

* Replaces the usage of the deprecated method ```getCoreName()``` with ```getPrimaryEndpont()->getCore()```

# How to test

No deprecation notice is triggered anymore. When a core is reloaded or an index is emptied.

Fixes: #2286


can0199:
Bug occures still in solr version 9.0.2 please update this too

Issue #2238, opened 2019-02-05
[TASK] Combine site handling in EXT:solr with TYPO3 core


timohund:
**What should be done in the scope of this task?**

Currently we have the class **Site** in EXT:solr to handle everything that is related to site's with EXT:solr.

TYPO3 9 introduced the site handling in the TYPO3 core and we now want to combine those concepts in a reasonable way.

By now we have:

* ApacheSolrForTypo3\Solr\Site: Class that represents a site in EXT:solr for historical reasons at that place. Should be moved to ApacheSolrForTypo3\Solr\Domain\Site\Site
* ApacheSolrForTypo3\Solr\Domain\Site\SiteRepository: Responsible to instanciate a site for given pages
* ApacheSolrForTypo3\Solr\Domain\Site\SiteHashService: Responsible to build site hashes for a site

Thinks we should do:

1. Move ApacheSolrForTypo3\Solr\Site to ApacheSolrForTypo3\Solr\Domain\Site\Site in order to have all site logic in this package
2. Move additional build logic from the Site class to the SiteRepository to have as less as possible logic in the site class itself
3. Add subclasses of Site "LegacySite" and "TYPO3ManagedSite" that have the same interface.
4. Make sure that "LegacySite" contains the old site logic that is used for now with solr sites
5. Make sure that "TYPO3ManagedSite" contains the new logic that can be maintained with the site management from now on.

timohund:
1. and 2. could allready be done with EXT:solr 9

tbd-rsm:
As suggested via Slack, I'll drop this gist here for a temporary custom QnD approach on SiteManagement (strict fallback_mode) handling for the current state of Ext:solr:

https://gist.github.com/tbd-rsm/000fcd9ab46b09da9741775cc3264929

Maybe this can point someone into the right direction.

PR #2289, opened 2019-03-13
[WIP][FEATURE] Use TYPO3 site configuration for solr connection info

frans-beech-it:
# What this pr does

Add the option to configure your solr connection info through the TYPO3 site configuration instead of `typoscript` and/or `$GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['solr']['sites']....`

# How to test

Configure site with the `Site management tool` and check if cores can be found

Fixes: #2238

# Todo
- Check if we can hide "initialise conneciton" when site config is used
- Add missing fields to site configuration
- Add proper tests
- Fix solr core -> path resolving
- Cleanup todo in code (remove/deprecate unused methods)
- Think about how we handle username and password (site configuration could be public accessable)
- Documentation of Legacy and TYPO3Managed site mode

### DIFF
--- a/Classes/ConnectionManager.php
+++ b/Classes/ConnectionManager.php
@@ -72,13 +72,19 @@ class ConnectionManager implements SingletonInterface
     protected $pagesRepositoryAtExtSolr;
 
     /**
+     * @var SiteRepository
+     */
+    protected $siteRepository;
+
+    /**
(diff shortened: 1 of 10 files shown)


timohund:
Will continue in #2291

PR #2291, opened 2019-03-15
[FEATURE] Use TYPO3 site configuration for solr connection info

timohund:
# What this pr does

With version 9 the concept of sites has been introduced in TYPO3. A site in TYPO3 can be used to configure everything that is related to domains and languages.

In the solr extension we also had a site concept before. The site concept in EXT:solr collects all connection information from the TypoScript configuration and stores that in the sys_registry. In addition the configuration ```$GLOBALS['TYPO3_CONF_VARS]['EXTCONF']['solr']['sites']```could be used to configure domains, when no TYPO3 domain record was configured.

Since the addtional domain record configuration was only developt because this was  not supported by the core, we should now get rid of that configuration. 

In addition the parsing the TypoScript and storing the connection information was only required to have a good performance (because TSFE needs to be instanciated from the backend). Therefore we should also remove the sys_registry approach and now configure everything that is Apache Solr related in the site configuration.

This pr:

* Refactors the SiteRepository to retrieve 
   * LegacySite for and old (TypoScript & EXTCONF & Domainrecord) 
   * Typo3ManagedSite for new sites managed by the TYPO3 site management
* Refactors the connection manager to remove the retrieval of the connection configuration to the site which allows us to use the old and new configuration handling of connection configuration
* Add's the configuration to the site configuration that allows to configure
   * Scheme, host, port and path at the global site level
   * Corename on the language level

# How to test

Configure site with the `Site management tool` and check if cores can be found

Fixes: #2238

### DIFF
--- a/Classes/ConnectionManager.php
+++ b/Classes/ConnectionManager.php
@@ -27,13 +27,10 @@ namespace ApacheSolrForTypo3\Solr;
 use ApacheSolrForTypo3\Solr\Domain\Site\Site;
 use ApacheSolrForTypo3\Solr\Domain\Site\SiteRepository;
 use ApacheSolrForTypo3\Solr\System\Logging\SolrLogManager;
-use ApacheSolrForTypo3\Solr\System\Page\Rootline;
 use ApacheSolrForTypo3\Solr\System\Records\Pages\PagesRepository as PagesRepositoryAtExtSolr;
 use ApacheSolrForTypo3\Solr\System\Records\SystemLanguage\SystemLanguageRepository;
 use ApacheSolrForTypo3\Solr\System\Solr\Node;
 use ApacheSolrForTypo3\Solr\System\Solr\SolrConnection;
-use TYPO3\CMS\Backend\Routing\UriBuilder;
-use TYPO3\CMS\Backend\Toolbar\ClearCacheActionsHookInterface;
 use TYPO3\CMS\Core\Registry;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\TypoScript\ExtendedTemplateService;
@@ -41,10 +38,7 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Frontend\Page\PageRepository;
 
 /**
- * A class to easily create a connection to a Solr server.
- *
- * Internally keeps track of already existing connections and makes sure that no
- * duplicate connections are created.
+ * ConnectionManager is responsible to create SolrConnection objects.
  *
  * @author Ingo Renner <ingo@typo3.org>
  */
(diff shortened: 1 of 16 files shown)


rschaufler:
@timohund @frans-beech-it  I can confirm that this approach is currently working with TYPO3 9.5 Site Handling. I've used the following steps:

- NO domain record created
- NO constants defined for connection purposes
- NO EXTCONF used for connection purposes
- NO initialization of Solr connections used (no entry in `sys_registry` created)
- Configured the solr connection ONLY via the Site Configuration

I am able to successfully index `pages` and query them in the frontend. Note that this is currently only a single language instance, so i didn't check multi language support.

The only issue i have at the moment is that i cannot configure different connections depending on the current environment. How about a solution based on variants which is like the one that is used for the `Entry Point [base]` in the Site Configuration?

timohund:
@rschaufler Thx for testing. Is is possible to you %env variables for that in the yaml file? I would try to keep the configuration in the UI as easy as possible.

Beside that there are a few things left. to have that ready:

* Tests
* Deprecation logging of legacy logic
* Documentation

rschaufler:
@timohund thanks for the hint. Using env variables works like described here:
https://docs.typo3.org/typo3cms/CoreApiReference/ApiOverview/SiteHandling/UsingEnvVars.html

There are two issues with it:

- when using shared hosting solutions this could be a problem if connection data has authentication credentials, because in that case one could be forced to handle env variables via htaccess for example which is publicly accesibly (i think like you pointed out in your original description).
This issue can be avoided though by using https://github.com/helhum/dotenv-connector which makes it pretty easy to load a custom .env file when otherwise possbilities are limited
- In the site configuration `solr_scheme_read` currently is a selectbox, so i have no possiblity to use an env variable there (at least when using the UI).
I propose to make it simply an input field like it is the case for the other configuration values.
Same issue goes for `solr_port_read` which is an input type "number".

How could i give you additional pr's for proposals based on this pull request? Would you possible merge it first in order to be worked on in the future?

hirnschmalz:
`SOLR_RELATIONS` are not properly handled in a multi language setup. An ugly workaround at the moment is to set `$GLOBALS['TSFE']->sys_language_uid = $language;` in `\ApacheSolrForTypo3\Solr\Util::initializeTsfe` until the TYPO3 Site Configuration is respected.

timohund:
For the username and password i would propose the following approach:

* Only allow to configure that with the yaml and not by the UI add to the documentation to propose to configure an ENV variable for that or to make sure that the site configuration is located in a folder outsite the webroot.

@rschaufler When your hosted forces you to configure env variables by htaccess and that is public available i would say this is a problem of your hoster. I would propose to make it as hard as possible to configure an insecure environment here (with the credentials). Therefore i would not add the username and password to the ui and recommend to use env vars if this is required. Do you have another approach to store these values in a secure location that works all setups?

timohund:
I've removed the todos and will add seperate issues for them, to get that merged a split up the remaining things

Issue #2293, opened 2019-03-19
[BUG] Integer in FlexForm gets quoted


manuelselbach:
**Describe the bug**
If you set an integer value in the FlexForm of the a filter, this will be quoted and thus not valid to filter e.g. on uid.

**To Reproduce**

1. Add a Filter in the search plugin
2. select `uid`
3. add a page uid in "Value"

-> In ApacheSolrForTypo3\Solr\System\Service\ConfigurationService:135 the value gets quoted.

**Expected behavior**
Integer values should not be quoted thus the filter query is valid to filter on integer fields of an Solr document.

**Used versions (please complete the following information):**
 - TYPO3 Version: 8.7.22
 - EXT:solr Version: 8.1.2

manuelselbach:
👍

PR #2294, opened 2019-03-19
[BUGFIX] Do not quote integer values for flexform filters

timohund:
# What this pr does

* Removes the quoting if a numeric value for the filter was configured in the flexform.

# How to test

Configure a numeric filter in the flexform and check the frontend.

Fixes: #2293


timohund:
@manuelselbach Can you please review review this?

PR #2300, opened 2019-03-27
[TASK] migrate calls to PageRepository->getRootLine calls

infabo:
use RootlineUtility directly as proposed by the deprecation message

# What this pr does

migrate deprecated calls to PageRepository->getRootLine

# How to test


timohund:
We need to rebase that and check why the tests fail

PR #2304, opened 2019-04-18
[WIP][BUGFIX] Use the SiteMatcher to create url for indexing pages

georgringer:
Instead of building URLs by hardcoded strings like `index.php?id=`, the `SiteMatcher` and router must be used to create urls

What could be changed:
- Not sure about using `BackendUtility` in FE; that should be avoided
- As the mointpoint info is not used in the method, more could be restructured


timohund:
@georgringer I guess want you want to do should/is done in https://github.com/TYPO3-Solr/ext-solr/blob/master/Classes/Domain/Index/PageIndexer/Helper/UriBuilder/TYPO3SiteStrategy.php

georgringer:
ah ok, hm I guess I will close the PR then

Issue #2295, opened 2019-03-19
[BUG] TYPO3 9.5: language handling in indexer is broken


dmitryd:
**Describe the bug**

In TYPO3 before 9.x it was ok to cache TSFE in the indexer using the root page id and language id like this (Classes/Util.php):

```
        $cacheId = $pageId . '|' . $language;


        if (!isset($tsfeCache[$cacheId]) || !$useCache) {
```

In TYPO3 9.x this no longer works correctly because language is not a property of TSFE anymore, it is a property of the context object. I found that records often are not indexed correctly if using `CONTENT` typoscript for rendering irre records (think of irre records in news). Even if solr thinks it tries to index them for the default language and use TSFE with language=0, context may have language id=1. The code inside `ContentContentObject` will try to apply overlay using functions from `PageRepository`. `PageRepository` uses context object and it may have value, which does not correspond to the value in TSFE. So overlay will return the empty value and record will be discarded. For me it happened like this:

* record 1 is indexed with the default language
* TSFE is created with language id=0 and cached
* context is initialized with language=0
* record 1 is indexed with language=1
* TSFE is created with language id=1 and cached
* context is initialized with language=1
* record 2 is indexed with the default language
* TSFE is *not* created but taken from cache, language id=0
* context is *not* initialized but still uses language=1 from the previous run
* ContentContentObject applies overlay with language=1 => unsets the record in the default language, no more content can be indexed

So each connected irre record gets just one single element.

**To Reproduce**
Steps to reproduce the behavior are more or less described above

 - TYPO3 Version: 9.5.5
 - EXT:solr Version: 9.0.2

** Possible fix **

I found that not disabling TSFE cache in Util.php solves the problem.

gerchomat:
Same problem here

gerchomat:
If you work with the old language-condition syntax and set the config.sys_language_uid it works.
`[globalVar = GP:L = 1]
    config.sys_language_uid = 1
[global]`

SomeBdyElse:
I can confirm this bug. Adding `[globalVar = GP:L = 1] config.sys_language_uid = 1 [global]` did not help for me. Once I applied that TS condition, translated records in the translated core looked fine. However, they also started to show up in the default language Solr core. 
I can confirm the behavior that @dmitryd describes. The TSFE settings from one language seems to carry over to the next one.

dmitryd:
[...]

PR #2312, opened 2019-05-16
[BUGFIX] Fix language handling during indexing

timkarliczek:
# What this pr does
Fixes indexing of translated records that make use of `SOLR_RELATION`

* Set language id in language context during indexing
* Set language id in TypoScriptFrontendController during indexing
* replace calls to $GLOBALS['TSFE']->sys_language_uid with calls to the language context
* However setting `$GLOBALS['TSFE']->__set('sys_language_uid', $language);` was still necassery. Otherwise when TSFE is taken from the Cache the language for `SOLR_RELATION` was always the default language.

#### Translated pages are not indexed
_Can already be fixed with a updated TypoScript configuration (check below)_
When Typo3 9 is used with a site config the non speaking URLs with L=1 (which EXT:solr is currently using for indexing) had no effect
The content was always delivered from the default language. This is more a configuration issue than a bug of EXT:solr.

#### SOLR_RELATION for translated records return wrong value
_My changes should solve this_
In my TypoScript indexing config for tx_news_domain_model_news i use `SOLR_RELATION`. There always language uid 0 was used.
Like dmitryd described in #2295  the wrong language id is used here. So language overlay is not performed correctly.
Disabling cache works, but slows down the indexing process significantly.

I reproduced this behaviour with
TYPO3 Version: 9.5.7
EXT:solr Version: 9.0.2

# How to test
Add the following to your TypoScript Config to fix indexing of pages. This also works before my changes.
```
[request.getQueryParams()['L'] == 0]
    config.sys_language_uid = 0
    config.language = de
[end]

[request.getQueryParams()['L'] == 1]
    config.sys_language_uid = 1
    config.language = en
[end]
```

- In Typo3 9 with site configuration and e.g. language 0 = de and language 1 = en create a translation of a page and try to index the site in both languages. This should now work.
- Use the example configuration for `tx_news_domain_model_news` (provided by EXT:solr)
 - Create translations for at least two news records
 - Assign categories and translate them as well
 - Index the records
 - The indexed news records should contain the title of the translated categories

Fixes: #2295

### DIFF
--- a/Classes/ContentObject/Relation.php
+++ b/Classes/ContentObject/Relation.php
@@ -26,6 +26,7 @@ namespace ApacheSolrForTypo3\Solr\ContentObject;
 
 use ApacheSolrForTypo3\Solr\System\Language\FrontendOverlayService;
 use ApacheSolrForTypo3\Solr\System\TCA\TCAService;
+use ApacheSolrForTypo3\Solr\Util;
 use Doctrine\DBAL\Driver\Statement;
 use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Database\Query\QueryBuilder;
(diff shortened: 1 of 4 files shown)


timkarliczek:
I created this pull request on branch master. The current master does not seem to be compatible with Typo3 9 (because there is no Typo3 version with 9.5.99 or higher). I created a new pull request based on branch release-9.0.x and will close this one. So this fix can be tested properly with Typo3 9 and merged to release-9.0.x when verified.

Issue #2150, opened 2018-10-08
TSFE not properly initialised when indexing records via additionalPageIds


rschaufler:
Used version: 8.1.2

When indexing records that are configured via `additionalPageIds`, the `$GLOBALS['TSFE']` is not properly initialised.
Example:

Page ID root: 1
Page ID records (not within root): 20

when indexing records from pid 20, the indexer tries to find an indexing configuration on pid 20 first, see https://github.com/TYPO3-Solr/ext-solr/blob/8.1.2/Classes/IndexQueue/Indexer.php#L280

This leads to `$GLOBALS['TSFE']` being initialised for page ID 20. Since there is no configuration, the root page of this item will be used to find the configuration, see https://github.com/TYPO3-Solr/ext-solr/blob/8.1.2/Classes/IndexQueue/Indexer.php#L282

While this returns the proper configuration, the TSFE is still initzialized for page ID 20 which is wrong.
TSFE should be initialised for page id 1 instead.
Due to a broken TSFE, the URL to the record can't be rendered properly, especially when using realurl.

As a temporary solution i modified https://github.com/TYPO3-Solr/ext-solr/blob/8.1.2/Classes/Util.php#L138 as follows:

```
    /**
     * Loads the TypoScript configuration for a given page id and language.
     * Language usage may be disabled to get the default TypoScript
     * configuration.
     *
     * Forces re-initialization of TSFE
     *
     * @param int $pageId Id of the (root) page to get the Solr configuration from.
     * @param string $path The TypoScript configuration path to retrieve.
     * @param bool $initializeTsfe Optionally initializes a full TSFE to get the configuration, defaults to FALSE
     * @param int $language System language uid, optional, defaults to 0
     * @param bool $useTwoLevelCache Flag to enable the two level cache for the typoscript configuration array
     * @return TypoScriptConfiguration The Solr configuration for the requested tree.
     */
    public static function getConfigurationFromPageId($pageId, $path, $initializeTsfe = false, $language = 0, $useTwoLevelCache = true)
    {
        $pageId = self::getConfigurationPageIdToUse($pageId);

        static $configurationObjectCache = [];
        $cacheId = md5($pageId . '|' . $path . '|' . $language . '|' . ($initializeTsfe ? '1' : '0'));

        // force proper TSFE re-initialisation
        if ($initializeTsfe) {
            Util::initializeTsfe($pageId, $language);
        }

       [...]
```

froemken:
[...]

PR #2323, opened 2019-06-06
[BUGFIX] Initialize TSFE, if conf of page was cached

froemken:
With first call of getConfigurationFromPageId()
the parameter initializeTsfe is checked and processed.
With a second call to getConfigurationFromPageId()
we now check, if TSFE has to be initialized also while
retrieving data from cache.

Resolves: #2150

# What this pr does

It initializes TSFE also, if configuration was fetched from configurationObjectCache

# How to test

A perfect instruction you can find here: https://github.com/TYPO3-Solr/ext-solr/issues/2150

Fixes: #2150

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -140,6 +140,9 @@ class Util
         static $configurationObjectCache = [];
         $cacheId = md5($pageId . '|' . $path . '|' . $language . '|' . ($initializeTsfe ? '1' : '0'));
         if (isset($configurationObjectCache[$cacheId])) {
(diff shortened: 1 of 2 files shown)


froemken:
Please do not merge...I will try to add a UnitTest first...

Mabahe:
Stumbled over this, too. But in my case it is the second level cache. If $initializeTsfe is true and there is a hit in the second level cache, TSFE needs to get initialized, too. Will add an additional issue and PR

Issue #2328, opened 2019-06-28
[BUG] TSFE is not properly initialised if configuration has a hit in 2nd level cache


Mabahe:
Stumbled over a similar problem like described in #2150, but with the second level cache.
As a result you get different TSFE objects when indexing records via additionalPageIds outside the normal page tree for the first indexing (without TwoLevelCache hit) and second indexing (with TwoLevelCache hit).

To fix this, TSFE needs to get also initialized, if $initializeTsfe is true and there is a hit in the 2nd level cache.

dkd-kaehm:
@Mabahe which versions are affected?

timohund:
@Mabahe can you please check if this issue is fixed for you with #2329 or if not, do you have the time to prepare a pr?

Mabahe:
This is not related to #2329, since the 1stLevel cache is static var in Util::getConfigurationFromPageId(). Yes, will push a PR.
Perhaps the cache usage in  Util::getConfigurationFromPageId() could be optimized after this.

Mabahe:
@dkd-kaehm in this project we are on 7.5. but also master is affected

PR #2329, opened 2019-07-03
[BUGFIX] Fill firstLevelCache with secondLevelCache result

IchHabRecht:
If the function \ApacheSolrForTypo3\Solr\System\Cache\TwoLevelCache::get is used, currently every time the secondLevelCache is fetched again as the result is never stored in the firstLevelCache. This is very expensive as it means fetching the result from database (default backend cache) and unserialize the information (as it is stored serialized by default).

### DIFF
--- a/Classes/System/Cache/TwoLevelCache.php
+++ b/Classes/System/Cache/TwoLevelCache.php
@@ -109,7 +109,10 @@ class TwoLevelCache
             return $firstLevelResult;
         }
 
-        return $this->secondLevelCache->get($cacheId);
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
@IchHabRecht: is that related to #2328?

IchHabRecht:
Hi @timohund,

I'm not sure about #2328 as I don't exactly understand the problem nor I can reproduce it currently (in our project). But I would guess it is not related at all because this patch just fills the first level cache and doesn't prevent hitting the second level cache.

PR #2342, opened 2019-08-12
[TASK] Increase Solr Version to 7.7

dkd-kaehm:
Increase Version to Solr v7.7


timohund:
We updated to 8.2 with EXT:solr 10

Issue #2301, opened 2019-04-03
[BUG] Filter in flexforms make facets not working


georgringer:
If a filter is set in the FlexForms, the result is reduced to this filter. However if an additional facet is selected, this is not applied anymore. The facet is still shown as used facet. 

If the filter is defined in TypoScript
```
plugin.tx_solr {
    search.query.filter {
        restrictedType = type:tx_xxxxx
    }
}
```
everything works 

**Used versions (please complete the following information):**
 - TYPO3 Version:9.5.5
 - EXT:solr Version: dev-master

RinyVT:
I think this is because the useFilterArray uses a key to set the filter:
https://github.com/TYPO3-Solr/ext-solr/blob/master/Classes/Domain/Search/Query/AbstractQueryBuilder.php#L96

When selecting filters in the flexform the keys are array-integers, 1 filter gets the index 0.
But the facets from the url also use array-indexes:
https://github.com/TYPO3-Solr/ext-solr/blob/master/Classes/Query/Modifier/Faceting.php#L157

So the first facet in the url gets key 0 and in this function there is a check if the key has already been set:
https://github.com/TYPO3-Solr/ext-solr/blob/master/Classes/Domain/Search/Query/AbstractQueryBuilder.php#L351

So the facet is not applied. In my case changing line 157 of Faceting.php from:
`$facetFilters[] = $tag . '(' . implode($operator, $filterParts) . ')';`
to:
`$facetFilters[$tag] = $tag . '(' . implode($operator, $filterParts) . ')';`
fixes the problem but I'm not sure if this is a good solution.

snusson:
Any plans, when this is gonna be fixed?

PR #2343, opened 2019-08-14
[BUGFIX] Add facet name to facet filters

georgringer:
To be able to combine facets from _GET and from flexforms, the key of the facets must be set to a proper name. Otherwise the index would be overriden.

Resolves: #2301


beechit:
@timohund Could you backport this issue also to version 9?

timohund:
@beechit i've created a pr: #2464 can you please test that then we can merge it to release-9.0.x

PR #2360, opened 2019-08-29
[BUGFIX] Initialize TSFE, if conf of page was cached

Mabahe:
With first call of getConfigurationFromPageId()
the parameter initializeTsfe is checked and processed.
With a second call to getConfigurationFromPageId()
we now check, if TSFE has to be initialized also while
retrieving data from cache.

* Add UnitTests for Util
* Set error_reporting to 0 in Util class
* Change pageIds for each Util test because of static method cache

Resolves: #2150

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -140,6 +140,9 @@ class Util
         static $configurationObjectCache = [];
         $cacheId = md5($pageId . '|' . $path . '|' . $language . '|' . ($initializeTsfe ? '1' : '0'));
         if (isset($configurationObjectCache[$cacheId])) {
+            if ($initializeTsfe) {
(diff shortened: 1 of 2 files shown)


Mabahe:
Backport of #2323 to release-9.0.x

timohund:
@Mabahe The test's fail with TYPO3 8 can you please have a look on that?

Mabahe:
Jep, ok to remove them, since they are green in master and this is a backport onyl?

dkd-kaehm:
> Jep, ok to remove them, since they are green in master and this is a backport only?

@Mabahe Try to remove `->shouldBeCalledOnce()`

PR #2366, opened 2019-09-02
don't remove content that is visible to the user (9.0.x)

### LINKED ISSUE #2348: [BUG] Escaped xml-like strings are stripped form the search
**Describe the bug**
The HtmlContentExtractor strips everything that looks like xml, even something like `<something else>`. There is a similar issue in #1541 but that did only resolve some cases.

**To Reproduce**
Add content like `<number>` to a page and index it, it will be stripped.

**Expected behavior**
All literal content should be searchable

 - TYPO3 Version: 8.7.27
 - Browser: chrome
 - EXT:solr Version: 7.5.3 but also reproducable in 8.1.2
 - Used Apache Solr Version: 6.6.2
 - PHP Version: 7.1.31
 - MySQL Version: 5.7.27

Nemo64:
same as #2349 but for 9.0.x
fixes #2348

### DIFF
--- a/Classes/HtmlContentExtractor.php
+++ b/Classes/HtmlContentExtractor.php
@@ -128,9 +128,8 @@ class HtmlContentExtractor
     public function getIndexableContent()
     {
         $content = self::cleanContent($this->content);
-        $content = html_entity_decode($content, ENT_QUOTES, 'UTF-8');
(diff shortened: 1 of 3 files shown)


timohund:
@Nemo64 There are failing unit tests on travis we need to check why they fail and fix that

Issue #1626, opened 2017-09-06
TypoScript setting keepExistingParametersForNewSearches is not implemented


netcoop:
The feature is more than useful but appears to be lost in refactoring.

What still exists:

- TypoScript setting 'plugin.tx_solr.search.keepExistingParametersForNewSearches' in the templates
- documentation
- method 'getSearchKeepExistingParametersForNewSearches' in TypoScriptConfiguration.php, but it is not called anywhere.

PR #2369, opened 2019-09-03
[FEATURE] Re-implement 'keepExistingParametersForNewSearches' (#1626)

netcoop:
# What this pr does

Re-implement the feature 'keepExistingParametersForNewSearches': When doing a new search, existing parameters like filters will be carried over to the new search.
The TypoScript setting and documentation already exist from an earlier version (feature got lost in a rewrite)

# How to test

Select one or more facets, type new search words and search. With this patch, the selected facets remain active for the new search.

Fixes: #1626


timohund:
@netcoop Thanks for the pull request! There is one failing functional test, do you have the time to have a look on that?

netcoop:
Thank you for the quick response. I also noticed the error after submitting the PR. Apparently something changed between TYPO3 v8 and 9, in 8 the controllerContext was available in the viewHelper, but now it isn't anymore. It's this line:
`$arguments = $this->controllerContext->getRequest()->getArguments();`
so I need to figure out how to get hold of the search arguments now. Let me know if you immediately know the answer, otherwise I'll try to figure it out later today.

netcoop:
I decided to go for GeneralUtility::_GPmerged('tx_solr'), other options seem more likely to change in the near future.

Issue #2374, opened 2019-09-04
[BUG] getAllRequirementsMet returns true when one requirement is met


netcoop:
**Describe the bug**
RequirementsService->getAllRequirementsMet() with multiple requirements returns true as soon as 1 requirement is met.

PR #2376, opened 2019-09-04
[BUGFIX] Fix getAllRequirementsMet returns true too soon

netcoop:
RequirementsService->getAllRequirementsMet() returns true as soon as one requirement is met (#2374)

# What this pr does

With this PR true is only returned when ALL given requirements are fullfilled, not just one

Fixes: #2374


timohund:
@netcoop Can you add a unit test, that fails without the fix and passes with the fix?

netcoop:
I was looking into that already but didn't get the unit testing working... need some up-to-date howto

timohund:
@netcoop The contribution guide lines describe how to run them: https://github.com/TYPO3-Solr/ext-solr/blob/master/CONTRIBUTING.md#automated-tests

netcoop:
Thanks, will give that a try soon!

Issue #2364, opened 2019-09-02
[TASK] Add Extension configuration allowLegacySiteMode and evaluate it


timohund:
**What should be done in the scope of this task?**

We should add an extension configruation "allowLegacySiteMode" that need's to be explicitly enabled to use domain records and typoscript to configure EXT:solr.

The implementation of that flag should be used to:

* Only show the global initialize connections flag, if the legacy mode is enabled
* Show the initialize connections flag per tree only when the legacy mode is enabled
* Throw an exception. If the old logic is triggered (SiteRepository::buildLegacySite) but the legacy site handling is not enabled.

PR #2388, opened 2019-09-06
[FEATURE] Add legacy site mode : exception on invalid site conf combi

dkd-kaehm:
Throws an exception by booting legacy site conf, 
if the legacy site mode is disabled.

Fixes: #2364


timohund:
I've added https://github.com/TYPO3-Solr/ext-solr/issues/2408 as followup to remove Util::legacySiteModeIsEnabled

Issue #2425, opened 2019-09-24
[BUG] Info Module shows only indexed documents from root page (current master)


codemonkey1988:
**Describe the bug**
When using the info module in current master, the tab "Index Documents" shows only documents from root page.

**Expected behavior**
It should show the indexed documentes related to the page chosen in the pagetree.

**Used versions (please complete the following information):**
 - TYPO3 Version: 9.5.9
 - Browser: Firefox
 - EXT:solr Version: dev-master
 - Used Apache Solr Version: 7.6.0
 - PHP Version: 7.3

PR #2426, opened 2019-09-24
[BUGFIX] Use requested page uid in info module

codemonkey1988:
# What this pr does

Make sure always to use the page uid from selected pagein pagetree when fetching indexed documents in the backend info module.

# How to test

1. Setup TYPO3 9.5 instance iwth one site configuration
2. Using current master, create 2 pages with content
3. Index both pages
4. Switch to Solr info module and go for the tab "Index Documents"
5. Before this patch, you will always get the documents from the root page, now you will get the documents for current page

Fixes: #2425


timohund:
We need to check that, as far as i remember there was a good reason not to use requestedPageUID

PR #2458, opened 2019-10-07
[TASK] Proofread documentation and release notes

timohund:
# What this pr does

* Fixes some parts of the documentation and release notes
* Deprecates some more unneeded methods


Documentation/Backend/ConnectionManager.rst
```diff
@@ -3,10 +3,90 @@
...
+
+* Solr core name
+
+If you have a more complex setup or need to configure username and password please configure solr by editing the yaml file.
+
+The fallback mechanism allows you to overwrite settings on the language level.
+
+The fallback mechanism work like that:
+
+Each setting has the following structure ``solr_{$setting}_{$scope}"```. The score can be read or write. Every setting can be overwritten for the scope *write*, if nothing is configured it will fallback to the *read* setting. Every setting can be defined on the language level, if it is not configured on the language level it fallsback to the global setting.
```
Smove: The `score` can be?

PR #2464, opened 2019-10-10
[BUGFIX:BACKPORT] Add facet name to facet filters (#2343)

timohund:
# What this pr does

To be able to combine facets from _GET and from flexforms, the key of the facets must be set to a proper name. Otherwise the index would be overriden.

Fixes: #2301


baschte:
@timohund 
Any informations when this fix will be merged into version 9? 👍

dkd-kaehm (approved):

PR #2485, opened 2019-10-21
[FIX] EXT:solrconsole version in matrix for next addon release

dkd-kaehm:
There is small typo in Versionmatrix about solrconsole.
The next release of solrconsole should be 3.0. 

Byside of that the **solrdebugtools** added to the matrix.


timohund:
Thx, i did some small changes in #2490

Issue #2417, opened 2019-09-18
[BUG] Pages secured with fe_group cant be indexed


Navi2016:
**Describe the bug**
When having protected pages in the backend, indexing works fine as long as the page itself isn't secured by a group.

Example:
- parent page 1 (has group id 400, extend to subpages enabled)
-- subpageA (no groups)
--- subSubPageA (no groups)

This works fine.

Now this example:
- parent page 1 (has group id 400, extend to subpages enabled)
-- subpageA (no groups)
--- subSubPageA (also has a group, id 500)

Now subSubPageA cant be indexed and will fail with a Failed to execute Page Indexer Request. See log for details.

**Possible solution**
It appears its going wrong in:
typo3conf/ext/solr/Classes/Access/Rootline.php:163

$currentPageRecord = $pageSelector->getPage($pageId);

the $currentPageRecord is empty because it's using a query with access fields like fe_group must be 0 etc.

Adjusted code to:

$currentPageRecord = $pageSelector->getPage($pageId, true);

and indexing seems to work for those pages as well.

Is this a bug and are there any downsides to this fix?
I hope this does not mess with showing pages unprotected in the search results, will check this of course.

**Used versions**
 - TYPO3 Version: 7.6.39
 - EXT:solr Version: 6.5.1
 - Used Apache Solr Version: 6.6.6
 - PHP Version: 7.2.22
 - MySQL Version: 5.5

**Additional context**
I know the used solr extension is older but its the last version to work with TYPO3 7.6 and the code seems unchanged in newer versions.

Might be related to:
https://github.com/TYPO3-Solr/ext-solr/issues/2341

timohund:
Fixed with #2481

PR #2492, opened 2019-10-24
[BUGFIX] Indexing user protected pages does not work as expected

Tuurlijk:
# What this pr does

Disables the user permission check when building the access rootline since this is required to build an access rootline for user restricted pages

# How to test

Test to index pages with and without user restrictions

Fixes: #2417


timohund:
@Tuurlijk thanks for the backport! The build is currently failing because of some travis ci issues on the 6.5.x branch. I'll fix them and re-run the build then and merge it when everything is fin

timohund:
Will be merged with #2496

PR #2496, opened 2019-10-25
[BUGFIX] Indexing user protected pages does not work as expected

timohund:
What this pr does

Disables the user permission check when building the access rootline since this is required to build an access rootline for user restricted pages
How to test

Test to index pages with and without user restrictions

Fixes: #2417


Tuurlijk:
Hi @timohund. Thx for merging. Can you please do a release for 6.5.x?

Issue #2501, opened 2019-11-13
[BUG] Hierarchy is not indexed completely


DrWh0286:
**Describe the bug**

Imagine there is just one sys_category selected in a page, which is not in the first level. In this case the indexed value in a field, which is called e.g. taxonomy_strangM, holds just the first level of the hierarchy.

Example field configuration:

```
taxonomy_stringM = SOLR_RELATION
taxonomy_stringM {
	localField = categories
	foreignLabelField = uid
	multiValue = 1
}
```

**To Reproduce**
1. Add the field as configured above to the solr fields
2. select only one category in categories field in TYPO3 BE, which is in a deeper level (> 1).

This should cause a solr document field value like: ["0-9/","1-9/36/"], but the value is just ["0-9/"] instead.

**Expected behavior**
Normally we would expect a solr document field value like: ["0-9/","1-9/36/"], but we get just ["0-9/"] instead.

**Used versions (please complete the following information):**
 - TYPO3 Version:9.5.11
 - Browser: Chrome (but doesn't matter)
 - EXT:solr Version: 9.0.2
 - Used Apache Solr Version: 7.6.0
 - PHP Version: 7.3.10
 - MySQL Version: 5.7.28

dkd-kaehm:
@maritwho 
I cant reproduce described behaviour, I tried following:

```
plugin.tx_solr.index {
  fieldProcessingInstructions.taxonomy_stringM = categoryUidToHierarchy
  queue.pages.fields {
    taxonomy_stringM = SOLR_RELATION
    taxonomy_stringM {
      localField = categories
      foreignLabelField = uid
      multiValue = 1
    }
  }
}
```
With TypoScript config from above I get expected Values in `taxonomy_stringM` field. 
By commenting out of line 2 `fieldProcessingInstructions.taxonomy_stringM = categoryUidToHierarchy` the document gets following value `"taxonomy_stringM":["9"]`.

Moreover: the changes from #2840 break the indexing process, due of invalid Solr documents. The Apache Solr does not accept those documents.

This issue will be automatically closed on 22 Feb. 2021.

PR #2502, opened 2019-11-13
[BUGFIX] Fix wrong hierarchy for only one relation

maritwho:
If a field has just one related record the field processor service
detects this as a single value field. In case of hierarchy this might
be wrong, because a hierarchy can always have multiple values caused
by a depth > 1.

# What this pr does

Fix the bug, that e.g. just one sys_category is selected the indexed value in a field, which is called e.g. taxonomy_strangM, holds just the first level of the hierarchy. The configuration of the field looks like:

```
  taxonomy_stringM = SOLR_RELATION
  taxonomy_stringM {
    localField = categories
    foreignLabelField = uid
    multiValue = 1
  }
```
# How to test

Add the field as configured above to the solr fields and select only one category in categories, which is in a deeper level (> 1). This should cause a solr document field value like: ["0-9/","1-9/36/"], but the value is just ["0-9/"] instead.

Fixes: #2501

### DIFF
--- a/Classes/FieldProcessor/Service.php
+++ b/Classes/FieldProcessor/Service.php
@@ -83,16 +83,19 @@ class Service
                         $fieldValue = $processor->process($fieldValue);
                         break;
                     case 'pathToHierarchy':
+                        $isSingleValueField = false;
                         /** @var $processor PathToHierarchy */
                         $processor = GeneralUtility::makeInstance(PathToHierarchy::class);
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
Closing in favor of #2840, 
because no maintainer pushing allowed on pluswerk fork and therefore not able to rebase the stuff.

PR #2510, opened 2019-12-09
If we have root page in the root page, we always have only last page …

andron200683:
If we have root page in the root page, we always have only last page in rootLineArray!

# What this pr does

If we have root page in the root page

# How to test
Add new site configuration for some page with different solr paths and move this page into the root page, after that you need a switch to tab "info" in SOLR administration panel and compare root page path and children root page path


bmack:
Ha, I created the same change here https://github.com/TYPO3-Solr/ext-solr/pull/2628

dkd-kaehm:
Closing in favor of #2628. 
@andron200683 Thanks for assistance.

Issue #2525, opened 2019-12-31
[FEATURE] Composite value for foreignLabelField in SOLR_RELATION


saitho:
**Is your feature request related to a problem? Please describe.**

I want to setup a SOLR_RELATION to a record which does not have a title. The title rather consists of two columns (first name and last name).

**Describe the solution you'd like**
I suggest extending the SOLR_RELATION object for allowing a list of fields which will be concatenated into a string:

```
people_stringM = SOLR_RELATION
people_stringM {
  localField = tx_myext_people
  relationTableSortingField = sorting
  foreignLabelField = first_name, last_name
  multiValue = 1
}
``` 

The glue would also be configurable as "foreignLabelFieldGlue", defaulting to a " " (space).

**Describe alternatives you've considered**
Using a UserFunc, however that means I can't use a multivalue string here.

**Target versions**
10

saitho:
> **Describe alternatives you've considered**
> Using a UserFunc, however that means I can't use a multivalue string here.

Just realized I can split that using SOLR_MULTIVALUE. Used that for a more complex query via UserFunc. I'll still leave this issue and PR up for discussion if it's useful to allow label composites. ;)
If not I'll switch to SOLR_MULTIVALUE + UserFunc for that too.

saitho:
I don't need that anymore (also can't remember if I solved my issue with SOLR_MULTIVALUE... has been some time since I worked on that and couldn't find anything in code). Closing issue.

PR #2526, opened 2019-12-31
[FEATURE] Allow composite label and concatenate values by glue (#2525)

saitho:
# What this pr does

Allows creating a composite label by defining multiple label fields which are concatenated by a configurable glue. If no glue was configured it will use a blank space.

# How to test

Add some categories to a page and this to your TypoScript:
```
plugin.tx_solr.index.queue.pages.fields {
  categoryTest_stringM = SOLR_RELATION
  categoryTest_stringM {
    localField = category
    foreignLabelField = uid, title
    foreignLabelFieldGlue = -
    multiValue = 1
  }
}
```

Index the page. Look into the Solr document.
It should have a "categoryTest_stringM" field with values being the UID and title of the category separated by -.

Fixes: #2525

### DIFF
--- a/Classes/ContentObject/Relation.php
+++ b/Classes/ContentObject/Relation.php
@@ -183,27 +183,37 @@ class Relation
 
         $relatedRecords = $this->getRelatedRecords($foreignTableName, ...$selectUids);
         foreach ($relatedRecords as $record) {
-            if (isset($foreignTableTca['columns'][$foreignTableLabelField]['config']['foreign_table'])
-                && $this->configuration['enableRecursiveValueResolution']
-            ) {
-                if (strpos($this->configuration['foreignLabelField'], '.') !== false) {
-                    $foreignTableLabelFieldArr = explode('.', $this->configuration['foreignLabelField']);
-                    unset($foreignTableLabelFieldArr[0]);
-                    $this->configuration['foreignLabelField'] = implode('.', $foreignTableLabelFieldArr);
+            $itemParts = [];
+            foreach (explode(',', $foreignTableLabelField) as $fieldName) {
+                $fieldName = trim($fieldName);
+
+                if (isset($foreignTableTca['columns'][$fieldName]['config']['foreign_table'])
+                    && $this->configuration['enableRecursiveValueResolution']
+                ) {
+                    if (strpos($this->configuration['foreignLabelField'], '.') !== false) {
+                        $foreignTableLabelFieldArr = explode('.', $this->configuration['foreignLabelField']);
+                        unset($foreignTableLabelFieldArr[0]);
+                        $this->configuration['foreignLabelField'] = implode('.', $foreignTableLabelFieldArr);
+                    }
+
+                    $this->configuration['localField'] = $fieldName;
+
+                    $contentObject = GeneralUtility::makeInstance(ContentObjectRenderer::class);
+                    $contentObject->start($record, $foreignTableName);
+
+                    $itemParts = array_merge($itemParts, $this->getRelatedItems($contentObject));
+                } else {
+                    if (Util::getLanguageUid() > 0) {
+                        $record = $this->frontendOverlayService->getOverlay($foreignTableName, $record);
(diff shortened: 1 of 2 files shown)


Classes/ContentObject/Relation.php
```diff
@@ -306,42 +316,51 @@ protected function resolveRelatedValue(
         if (Util::getLanguageUid() > 0 && !empty($foreignTableName)) {
             $relatedRecord = $this->frontendOverlayService->getOverlay($foreignTableName, $relatedRecord);
         }
+        $values = [];
```
saitho: My use case did not cover this part of the code. That's why I was unable to find a test case for that.
However since this also uses the foreignTableLabelField setting it would probably break when allowing multiple values. So I adjusted that as well. Maybe you can help me with a test case for this one. ;)

Documentation/Configuration/Reference/TxSolrIndex.rst
```diff
@@ -573,7 +573,18 @@ The current record's field name to use to resolve the relation to the foreign ta
...
-Usually the label field to retrieve from the related records is determined automatically using TCA, using this option the desired field can be specified explicitly. To specify the label field for recursive relations, the field names can be separated by a dot, e.g. for a category hierarchy to get the name of the parent category one could use "parent.name" (since version:2.9).
+Usually the label field to retrieve from the related records is determined automatically using TCA, using this option the desired field can be specified explicitly.
+To specify the label field for recursive relations, the field names can be separated by a dot, e.g. for a category hierarchy to get the name of the parent category one could use "parent.name" (since version:2.9).
+If your label consists of multiple fields, separete them with a comma (since version:10.1).
+
+**foreignLabelFieldGlue**
+
+:Type: String
+:TS Path: plugin.tx_solr.index.queue.[indexConfig].fields.[fieldName].foreignLabelFieldGlue
+:Since: 10.1
```
saitho: Version would have to be adjusted if this doesn't make it in 10.1

Documentation/Configuration/Reference/TxSolrIndex.rst
```diff
@@ -573,7 +573,18 @@ The current record's field name to use to resolve the relation to the foreign ta
 :TS Path: plugin.tx_solr.index.queue.[indexConfig].fields.[fieldName].foreignLabelField
 :Since: 2.0
 
-Usually the label field to retrieve from the related records is determined automatically using TCA, using this option the desired field can be specified explicitly. To specify the label field for recursive relations, the field names can be separated by a dot, e.g. for a category hierarchy to get the name of the parent category one could use "parent.name" (since version:2.9).
+Usually the label field to retrieve from the related records is determined automatically using TCA, using this option the desired field can be specified explicitly.
+To specify the label field for recursive relations, the field names can be separated by a dot, e.g. for a category hierarchy to get the name of the parent category one could use "parent.name" (since version:2.9).
+If your label consists of multiple fields, separete them with a comma (since version:10.1).
```
saitho: Version would have to be adjusted if this doesn't make it in 10.1

Classes/ContentObject/Relation.php
```diff
@@ -183,27 +183,37 @@ protected function getRelatedItemsFromMMTable($localTableName, $localRecordUid,
...
+                } else {
+                    if (Util::getLanguageUid() > 0) {
+                        $record = $this->frontendOverlayService->getOverlay($foreignTableName, $record);
+                    }
+                    $itemParts[] = $record[$fieldName];
                 }
-                $relatedItems[] = $record[$foreignTableLabelField];
             }
+            $glue = ' ';
+            if (array_key_exists('foreignLabelFieldGlue', $this->configuration)) {
```
timohund: Just a proposal: Why not use the null coalescing operator here(https://www.php.net/manual/de/migration70.new-features.php)? Could be then done in one line.

saitho:
Closing. Apparently this can be achieved by using a SOLR_MULTIVALUE and UserFunc.

PR #2543, opened 2020-03-04
[TASK] Update php class docblock

jacobsenj:
This pr updates all class names mentioned in class docblocks to their acutal class name and removes all package references.


Classes/Domain/Search/ResultSet/Facets/OptionBased/Options/OptionsPackage.php
```diff
@@ -17,7 +17,7 @@
 use ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\Facets\AbstractFacetPackage;
 
 /**
- * Class HierarchyPackage
+ * Class OptionsPackage
  * @package ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\Facets\OptionBased\Options
```
dkd-kaehm: `@package *` can be removed as well.

jacobsenj: I added a commit which removes all @package references from the class docblock.

PR #2547, opened 2020-03-06
One category must have many ancestor categories

eliasfernandez:
One category must have many ancestor categories so returning only the first category, will result on not valid results when filtering by facets.

# What this pr does

Fix the multicategories problem on categoryUidToHierarchy

# How to test
Add one children category to a Document. It will only index the root category.

### DIFF
--- a/Classes/FieldProcessor/Service.php
+++ b/Classes/FieldProcessor/Service.php
@@ -91,11 +91,17 @@ class Service
                         /** @var $processor PageUidToHierarchy */
                         $processor = GeneralUtility::makeInstance(PageUidToHierarchy::class);
                         $fieldValue = $processor->process($fieldValue);
+                        if ($isSingleValueField && count($fieldValue) > 1){
+                                $isSingleValueField = false;
+                        }
                         break;
                     case 'categoryUidToHierarchy':
                         /** @var $processor CategoryUidToHierarchy */
                         $processor = GeneralUtility::makeInstance(CategoryUidToHierarchy::class);
                         $fieldValue = $processor->process($fieldValue);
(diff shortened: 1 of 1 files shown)


eliasfernandez:
Sorry to insist but i think it must be fixed. I'll try to expose the problem clearer. 

Imagine you have a document (from news by instance) and we define that kind of document to be indexed to categories with a categoryUidToHierarchy fieldProcessingInstructions.

If we have a tree like this:

~~~
parent_cat
└── middle_cat
    └── child_cat
~~~

1. If a document is only linked to *child_cat*. This document will be indexed under *parent_cat* only. 

2. Now imagine the same document but linked with *child_cat* and *middle_cat*. Then the document document will be indexed under *parent_cat*, *middle_cat* and *child_cat*.

The second behaviour is correct but the first is unexpectable.

dkd-kaehm:
@eliasfernandez 
please provide an Integration test with some test cases.
Or at least an fixture file, explaining problematic scenario. Also the sys_catecory tree and TypoScript for indexing configuration.

eliasfernandez:
I've just pushed a TYPO3 ddev instance with the solr configuration and a t3d with some categories and news. 

https://github.com/eliasfernandez/categoryUidToHierarchyBug/

I thought the problem will be on 10.0.x too but seems that only fails on 9.0.x instances.  

Thanks for checking it out @dkd-kaehm

dkd-kaehm:
@eliasfernandez Thanks for contributing to EXT:solr.
I'll close this PR, because I can't reproduce the described behaviour without this patch on EXT:solr master, 10.0.x, 11.0.x .
Feel free to reopen if it occurs on currently supported(master + two latest major releases) EXT:solr versions.

Issue #2552, opened 2020-03-26
[BUG] php - 7.4 implode deprecation error


joey-beech-it:
https://github.com/TYPO3-Solr/ext-solr/blob/ab955c4f0172d58257ee4d78f336b747cd67c8c3/Classes/Typo3PageContentExtractor.php#L65

https://www.php.net/manual/en/function.implode.php

Version | Description
-- | --
7.4.0 | Passing the glue after the pieces (i.e. not using the documented order of parameters) has been deprecated.

Should check for all instances of implode.

PR #2558, opened 2020-04-09
[BUGFIX] implode deprecation for php 7.4 (#2552)

ggothe:
# What this pr does

fixes deprecation notice emited since php 7.4

# How to test

There is no unit test covering this issue
Havent tried running integration test yet
works on my machine (tm)

Fixes: #2552


RobertLang:
This merge needs also to be made available in the branch for 10.0 as it affects this version as well.
Thanks,
Robert

Issue #2573, opened 2020-04-20
[BUG] Exception in FE using pi_results plugin


mschwemer:
Having the solr plugin installed and calling the page, leads to an exception

`Call to undefined method phpDocumentor\Reflection\DocBlock\Tags\InvalidTag::getType()`

Reason is an invalid `@var` annotation in `EXT:solr/Classes/Domain/Search/ResultSet/SearchResultSetService.php`. In line 88 the class name has a trailing `;`. 

If I remove it, the error is gone. 

**Used versions (please complete the following information):**
 - TYPO3 Version: V10 dev (master)
 - EXT:solr Version: master
 - Used Apache Solr Version: 8.2.0
 - PHP Version: 7.3.16

Using TYPO3 V10.3.0 does not produce this error.

georgringer:
@mschwemer imo this is now fixed, you can close the issue

PR #2575, opened 2020-04-20
[BUGFIX] Remove breaking semicolon from SearchResultSetService

mschwemer:
The trailing semicolon in line 90 was breaking the functionality of
phpdocumentor, which was updated by TYPO3 core recently. Removing it
makes the frontend work again.

Solves: #2573


achimfritz:
👍

Issue #2578, opened 2020-04-21
[BUG] Inconsitent index if TYPO3 Site base variants are used


sascha-egerer:
**Describe the bug**
The index is currently based on the sitehash which is build by using the TYPO3 site "base". 
TYPO3 has the possibility to define "baseVariants" which means the domain does change if some condition is met. I do use this to have a different domain for the backend than for the frontend. But this ends up in having documents in the index with 2 different sitehash values as the document is index based the "base" for the current context. This does also mean that you get different results for the same site if using different domains.

**To Reproduce**
1. Create a TYPO3 site with a baseVariant that is used if some condition does match
2. Initialize the index queue
3. Run the indexing from the CLI
4. Open TYPO3 with the domain used in the baseVariant and trigger the indexing via the indexing module

**Expected behavior**
There should be one document in solr with a unique sitehash as there is only one "site".
The siteHash should be build based on the "site identifier" and not based on the "base".

**Used versions (please complete the following information):**
 - TYPO3 Version: 9.5.15
 - EXT:solr Version: master

**Additional context**

avogt1701:
Thats correct for a baseVariants use case with different contexts (dev/prod). In our case we have multiple base variants for the same context (Production). 

I think it would be better to set the solr field "site" to the site identifier and add the application context as additional site hash field.

dmitryd:
Hopefully it will not slow done queries.

PR #2579, opened 2020-04-21
Use site identifier instead of site base to calculate sitehash

sascha-egerer:
The TYPO3 site handling has a "base" property which is the domain
of the website. But there are baseVariants that are applied if a
condition is matched. That means the "base" is not really persitend
and does not really identify the "site".
Instead of "base" the "identifier" should be used to decide which
site is currently active.

Resolves: #2578

# What this pr does

Please add a description here

# How to test


sascha-egerer:
@timohund Before i continue here i just want to ask if that is a way we can go. 
This would require a full reindex and also a force index cleanup as the extension does cleanups based on the site hash wich is different now.

dkd-kaehm:
IMHO, the base is still the right approach. I have an example:

We have hosted-solr customers, which use the same core for all their stages. Even for live and local machines.
By using the `identifier` all the machines will consume and mix-up the same index(sitehash/id/whatever), which will be sick and result to sudden side-effects.

---

@sascha-egerer You mentioned different domain name for TYPO3 Backend and indexing the records via Backend? 
What happens if indexing via CLI?

Can you explain a little bit more about the setup with different domain for Backend?
Also which conditions are used, is it possible to access FE by BE domain? ...

timohund:
@dkd-kaehm I think your arguments are arguments that support the change that @sascha-egerer  proposed. For the siteHash we should use the **final used domain**, especially to be able to have different document versions for different stages within the same core. Otherwise, a document will always be overwritten by the last writing system. Or do I miss something?

sascha-egerer:
> @dkd-kaehm I think your arguments are arguments that support the change that @sascha-egerer proposed. For the siteHash we should use the **final used domain**, especially to be able to have different document versions for different stages within the same core. Otherwise, a document will always be overwritten by the last writing system. Or do I miss something?

The problem is if the website is indexed under 2 different domains like i've described above. In the previous version this was not a problem as the Page-Indexing was always done by the domain configured via `frontendDataHelper.host` which is not possible anymore. Maybe it would also be a solution to bring that option back. 
And IMO indexing 2 stages into the same core is also not a really nice idea and should not be done.

timohund:
> And IMO indexing 2 stages into the same core is also not a really nice idea and should not be done.
 @sascha-egerer I agree, just said, it is possible ;) Regarding the frontendDataHelper. It would be nice if you could check that. Just by reading the code this should still be possible. In terms of the siteHash i would say we should have one unique siteHash by site and domain.

avogt1701:
Maybe it makes sense to add the application context to the site hash or add a optional environment variable to the site hash. So it would be possible to use the Site Identifier as "site" and use the same solr core for local / live environments.

Issue #2533, opened 2020-01-20
[BUG] Eval for solr_port in siteconfig is "int"


mschwemer:
If you use environment variables for the solr port in siteconfiguration, the field is filled with `%env(SOLR_PORT)%`. The site configuration evaluates the field to an integer. This makes it impossible to save the entire site configuration.

**To Reproduce**
1. Go to the module "Siteconfiguration"
2. Switch to the tab "Solr"
3. Add the string "`%env(SOLR_PORT)%`" to the field port
4. Try to save the record.

**Expected behavior**

I expect to be able to save the record.

**Used versions (please complete the following information):**
 - TYPO3 Version: 9.5.13
 - Browser: all
 - EXT:solr Version: 10.0.1

**Additional context**

**Solution**

Remove the evaluation to `int` in `EXT:solr/Configuration/SiteConfiguration/Overrides/sites.php` on line 55.

PR #2599, opened 2020-05-05
[BUGFIX] Remove eval int from port in site configuration

georgringer:
If using dot env, the eval int option leads to an error in the site config.

Resolves: #2533


maddy2101:
could this please be integrated into v10 as well?

dkd-kaehm:
@maddy2101 could you provide a PR against 10 relese branch?

PR #2604, opened 2020-05-06
[BUGFIX] set max TYPO3 Version to 10.4.99 in ext_emconf.php

dkd-kaehm:
set to max supported TYPO3 Version 10.4.99 in ext_emconf.php


dkd-kaehm:
@timohund 
IMHO critical before releasing 11

PR #2623, opened 2020-05-19
[BUGFIX] Re-enable Integration Tests for TYPO3 v10

bmack:
Some tests within SearchControllerTests were skipped for TYPO3 v10,
which are re-enabled now.

The main reason why this happened was/is the changed behaviour
of DependencyInjection in TYPO3 v10. This way, a Singleton (in this case
EnvironmentService of Extbase) cannot be replaced during a Functional
Call anymore - which should not be done in any case.

Some frontend-related tests failed for this reason, because it was not possible
to mock the behaviour of TYPO3_MODE=FE in Tests.

Enabling these tests showed that the tests failed as the ViewHelpers did not
receive the current request by the ControllerContext, which is now
optionally injected into all ViewHelpers that generated links.

To run these Integration Tests, a new PHPunit configuration
called "IntegrationFrontendTests.xml" is added which is executed with
the `TYPO3_MODE=FE` constant activated in the configuration.

All other tests continue to work as before.

In addition, some cleanups to avoid deprecation logs in TSFETestBootstrapper
and to execute frontend calls within another frontend calls have been
made including the handling of TYPO3_REQUEST.

### DIFF
--- a/Build/Test/IntegrationFrontendTests.xml
+++ b/Build/Test/IntegrationFrontendTests.xml
@@ -1,0 +1,29 @@
+<phpunit
+    backupGlobals="false"
+    backupStaticAttributes="false"
+    colors="true"
+    convertErrorsToExceptions="true"
+    convertWarningsToExceptions="true"
+    forceCoversAnnotation="false"
+    processIsolation="true"
+    stopOnError="true"
+    stopOnFailure="true"
+    stopOnIncomplete="false"
+    stopOnSkipped="false"
+    verbose="false">
+
+    <testsuites>
+        <testsuite name="ext-solr-integration-tests">
+            <directory>../../Tests/Integration/Controller</directory>
+            <exclude>../../Tests/Integration/Controller/Backend/</exclude>
+        </testsuite>
+    </testsuites>
+    <filter>
+        <whitelist processUncoveredFilesFromWhitelist="false">
+            <directory suffix=".php">../../Classes/</directory>
+        </whitelist>
+    </filter>
+    <php>
+        <const name="TYPO3_MODE" value="FE" />
+    </php>
+</phpunit>
--- a/Build/Test/IntegrationFrontendTests.xml
+++ b/Build/Test/IntegrationFrontendTests.xml
@@ -14,10 +14,14 @@
 
     <testsuites>
         <testsuite name="ext-solr-integration-tests">
-            <directory>../../Tests/Integration/Controller</directory>
-            <exclude>../../Tests/Integration/Controller/Backend/</exclude>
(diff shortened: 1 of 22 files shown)


Build/Test/cibuild.sh
```diff
@@ -70,4 +70,5 @@ fi
 
 echo "Run integration tests"
 INTEGRATION_BOOTSTRAP=".Build/vendor/nimut/testing-framework/res/Configuration/FunctionalTestsBootstrap.php"
-.Build/bin/phpunit --colors -c Build/Test/IntegrationTests.xml --bootstrap=$INTEGRATION_BOOTSTRAP --coverage-clover=coverage.integration.clover
\ No newline at end of file
+.Build/bin/phpunit --colors -c Build/Test/IntegrationTests.xml --bootstrap=$INTEGRATION_BOOTSTRAP --coverage-clover=coverage.integration.clover
```
dkd-kaehm, line 73: there is a check or ` &&` between commands needed. 
By failing of this command, the tests are still green https://travis-ci.org/github/TYPO3-Solr/ext-solr/jobs/688803625

Build/Test/IntegrationFrontendTests.xml
```diff
@@ -0,0 +1,29 @@
+<phpunit
```
dkd-kaehm, line 1: Could we define the constants in `setup()` or in test methods instead of separate the tests in different suites?
What happens with coverage tools like crutinizer & co. by splitting the tests in different suites?

bmack: Thought about that too, and tried that, but if we don't define this in the phpunit.xml, this is defined by Nimut Testing Framework in its Testing Bootstrap (https://github.com/Nimut/testing-framework/blob/master/src/TestingFramework/Bootstrap/AbstractBootstrap.php#L218) which happens before any PHP code is executed. Because TYPO3 is using a constant (TYPO3_MODE) internally, this can only defined once per PHP process. I had a similar issue with this in Core Testing Framework which isn't solvable in a better way (at least I don't know of) until Core gets rid of TYPO3_MODE code.

Build/Test/cibuild.sh
bmack: yes! will have a look

bmack: I fixed this, hope that's OK as is now.

dkd-kaehm:
We want to backport This Pull Request to release-11.0.x.

bmack:
@dkd-kaehm see https://github.com/TYPO3-Solr/ext-solr/pull/2633 for the backport

Issue #2624, opened 2020-05-21
[BUG] SiteNotFoundException in reports module


cjanody:
**Describe the bug**
Having a TYPO3 multi-site instance with some of them not configured results in a TYPO3\CMS\Core\Exception\SiteNotFoundException in the reports module (maybe elsewhere)

**To Reproduce**
1. Create a page tree with multiple root pages
2. Configure one of them in the sites module
3. Go to the report module
4. ... 

**Expected behavior**
Sites should appear as unconfigured

**Screenshots**
![Screenshot_20200521_084118](https://user-images.githubusercontent.com/3030277/82560259-140f1880-9b3f-11ea-895d-41c061cbf33d.png)

**Used versions (please complete the following information):**
 - TYPO3 Version: 9.5.18
 - EXT:solr Version: 11.0.1

liayn:
maybe #2634 is related

PR #2625, opened 2020-05-21
[BUGFIX] Prevent SiteNotFoundException in reports module

cjanody:
Fixes: #2624


dkd-kaehm:
rebased to current master

PR #2628, opened 2020-05-25
[BUGFIX] Fix multiple rootpages in nested sites

bmack:
SolR currently traverses all pages in the root line to the
most outer page with is_siteroot in a setup with nested sites.

The Indexer thus always indexed from the most outer page
in previous revisions, however the root page should always
be based on the first site within a single tree

# What this pr does

The PR modifies the root line object to always fetch the right
site, in addition, the indexer now fetches the right page ID and
root ID when resolving the site from the queue items.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,15 +26,13 @@ env:
     - PHP_CS_FIXER_VERSION="^2.16.1"
   matrix:
     - TYPO3_VERSION="^9.5"
-    - TYPO3_VERSION="^10.0"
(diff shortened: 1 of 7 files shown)


dkd-kaehm:
@bmack could you please rebase?

bmack:
@dkd-kaehm should we backport this change?

dkd-kaehm:
Ooh, didn't seen that.
No, I'm currently doing that.
Thanks

Issue #2626, opened 2020-05-22
[FEATURE] Logging: use logFileInfix instead of hardcoded path


frommo:
**Is your feature request related to a problem? Please describe.**
Currently Solr logs into a hardcoded logfile "typo3temp/var/logs/solr.log", which is in a different directory than the other logfiles. Since v9, TYPO3 by default logs into "typo3temp/var/log" (note the missing S). This is unexpected, and could lead to user not finding the log file at all. See also: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Feature-85236-InfixOptionToDefaultLogFileNamesForFileWriter.html

**Describe the solution you'd like**
Instead of using writerConfiguration ... logFile in ext_localconf.php, "logFileInfix" should be used (available since 9.4). This will log into the correct directory with a unique suffix for security reason, making it easier to find the log.

**Target versions**
Since logFileInfix is available in TYPO3 9.5 only, this should be included in all releases targeting this version.

dkd-kaehm:
Affected Versions:

* release-10.0.x
* release-11.0.x
* master

@frommo Could you provide a Pull-Requests for that?

PR #2629, opened 2020-05-27
[FEATURE] Change FileWriter configuration to use logFileInfix

frommo:
# What this pr does

Currently Solr logs into a hardcoded logfile "typo3temp/var/logs/solr.log", which is in a different directory than the other logfiles. Since v9, TYPO3 by default logs into "typo3temp/var/log" (note the missing S). This is unexpected, and could lead to user not finding the log file at all. See also: https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Feature-85236-InfixOptionToDefaultLogFileNamesForFileWriter.html

# How to test

Activate logging in TypoScript an do something to create log output (e.g. logging.indexing = 1). There will be a new log file created in typo3temp/var/log named "typo3_solr_f5841fa7f9.log".

Fixes: #2626


dkd-kaehm:
@frommo please rebase and force push.

Issue #2630, opened 2020-05-27
[TASK] Backport PR "Re-enable Integration Tests for TYPO3 v10" #2623 into release-11.0.x branch


dkd-kaehm:
The Backport of #2623 must be applied into release-11.0.x as soon as possible.

dkd-kaehm:
Ups, allready done in #2633.

Issue #2587, opened 2020-04-29
[TASK] Check and fix the things for skipped tests:


dkd-kaehm:
# Help wanted:

**Please indicate and point to existing issues according to desired test case.**

## All skipped tests must be double checked and fixed.

In some cases the tests must be just adjusted, in other the thing must be fixed:

- [x] `ApacheSolrForTypo3\Solr\Tests\Integration\Controller\SearchControllerTest::canRenderHierarchicalFacet`

- [x] `ApacheSolrForTypo3\Solr\Tests\Integration\Controller\SearchControllerTest::canFacetOnHierarchicalFacetItem`

- [x] `ApacheSolrForTypo3\Solr\Tests\Integration\Controller\SearchControllerTest::canFacetOnHierarchicalTextCategory`

- [x] `ApacheSolrForTypo3\Solr\Tests\Integration\Controller\SearchControllerTest::canSortByMetric`

- [x] `ApacheSolrForTypo3\Solr\Tests\Integration\Domain\Index\IndexServiceTest::canResolveBaseAsPrefix with data set "absRefPrefixIsFoo" ('foo', '/foo/en/?tx_ttnews%5Btt_news%...5a84a1')`

- [x] `ApacheSolrForTypo3\Solr\Test\Domain\Search\ResultSet\ResultSetReconstitutionProcessorTest::canApplyRenderingInstructionsOnOptions`

- [x] `ApacheSolrForTypo3\Solr\Test\Domain\Search\ResultSet\ResultSetReconstitutionProcessorTest::labelCanBeUsedAsCObject`

- [x] `ApacheSolrForTypo3\Solr\Tests\Integration\UtilTest::getConfigurationFromPageIdInitializesTsfe`

- [x] `ApacheSolrForTypo3\Solr\Tests\Integration\UtilTest::getConfigurationFromPageIdInitializesTsfeOnCacheCall`

timohund:
@dkd-kaehm Util related tests will be fixed with #2593 we should have a look on the others as well.

timohund:
@dkd-kaehm ResultSetReconstitutionProcessorTests will be fixed with #2594

timohund:
[...]

PR #2633, opened 2020-05-28
[BUGFIX:BACKPORT] Re-enable Integration Tests for TYPO3 v10

bmack:
Some tests within SearchControllerTests were skipped for TYPO3 v10,
which are re-enabled now.

The main reason why this happened was/is the changed behaviour
of DependencyInjection in TYPO3 v10. This way, a Singleton (in this case
EnvironmentService of Extbase) cannot be replaced during a Functional
Call anymore - which should not be done in any case.

Some frontend-related tests failed for this reason, because it was not possible
to mock the behaviour of TYPO3_MODE=FE in Tests.

Enabling these tests showed that the tests failed as the ViewHelpers did not
receive the current request by the ControllerContext, which is now
optionally injected into all ViewHelpers that generated links.

To run these Integration Tests, a new PHPunit configuration
called "IntegrationFrontendTests.xml" is added which is executed with
the TYPO3_MODE=FE constant activated in the configuration.

All other tests continue to work as before.

In addition, some cleanups to avoid deprecation logs in TSFETestBootstrapper
and to execute frontend calls within another frontend calls have been
made including the handling of TYPO3_REQUEST.

Fixes: #2630, #2587

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,15 +26,13 @@ env:
     - PHP_CS_FIXER_VERSION="^2.16.1"
   matrix:
     - TYPO3_VERSION="^9.5"
-    - TYPO3_VERSION="^10.0"
+    - TYPO3_VERSION="^10.4"
     - TYPO3_VERSION="dev-master"
 
 matrix:
   fast_finish: true
   allow_failures:
     - env: TYPO3_VERSION="dev-master"
-    - php: 7.2
-      env: TYPO3_VERSION="^10.0"
 
 before_install:
   - curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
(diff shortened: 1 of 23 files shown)


dkd-kaehm:
@bmack 
In the build matrix is TYPO3 ^10.0, that is strange. 
Could you please change ^10.0 to ^10.4 and remove allowed_failures for PHP 7.2

https://github.com/TYPO3-Solr/ext-solr/blob/990646796e5ec9a78810c288de4a0ff2213c2ea6/.travis.yml#L36-L37


in this PR?

bmack:
@dkd-kaehm in the backport only?

dkd-kaehm:
> @dkd-kaehm in the backport only?
Yes, I'll provide it on master. The master branch should according to the plan skip 9.5 support.

dkd-kaehm:
Wow, 
thanks @bmack!

Mabahe:
This change breaks link generation after generating the suggest url. Generated sorting and facet urls contains the type parameter from the suggest url.

Issue #2656, opened 2020-06-23
[BUG] Sorting and facet urls are broken because the link contains the `type` parameter


Mabahe:
Regression from #2633

dkd-kaehm:
@Mabahe Hello, thanks for reporting.
Is the exact position in the source code known?

Mabahe:
I guess the reason is, that the UriBuilder is reused now https://github.com/TYPO3-Solr/ext-solr/pull/2633/commits/de0bf01900064639d6d3c63831d4de8c15bee829#diff-12fa500b04ea9e7e8d08b8d0ce521464R61
Before the vh has it's own new UriBuilder instance.

Mabahe:
Add @bmack

bmack:
So we'd need a reset on the UriBuilder?

Mabahe:
Yes, is the one idea to fix it. But why you reuse the UriBuilder now?

bmack:
Otherwise it's impossible to make the Integration Tests work (or you have a better idea that I did not see?)

Mabahe:
@dkd-kaehm @timohund @achimfritz please check if this is safe and does not cause the next regression. Let me know if you prefer a revert of #2633 and solve the integration tests a different way.

markuskobligk:
I can confirm this PR works and I could not detect any side effects so far concerning link generation in other parts of the search.

PR #2659, opened 2020-06-24
[BUGFIX] Reset uriBuilder before building a new uri

Mabahe:
Fixes a regression from #2633. Since then the uriBuilder is reused
and needs a reset for every new uri.

Resolves: #2656


dkd-kaehm:
Aha, cool. There is the Backport.
Thanks @Mabahe.

Issue #2663, opened 2020-07-01
[BUG] Wrong pageUid assignment in getHeaders method of class PageIndexerRequest


jacobsenj:
Within method getHeaders() in class PageIndexerRequest the page uid is currently assigned as

```php
$pageId = $this->indexQueueItem->getRecordPageId();
```

which assigns the page parent uid to the variable. Instead the variable should use the already present page uid in the index queue record like so:

```php
$pageId = $this->indexQueueItem->getRecordUid();
```

Reference: https://github.com/TYPO3-Solr/ext-solr/commit/18d9e2b2bdf4ce21e28ff7622b3c01619f832ade#commitcomment-40293781

PR #2664, opened 2020-07-01
[BUGFIX] Fix assignment for page uid variable

jacobsenj:
The variable is currently assigned to the page parent uid, but for page records the index queue item record uid represents the page uid already.

Fixes: #2663


dkd-kaehm:
@jacobsenj Thanks for that fix. 
We should add unit test for this case.

Issue #2661, opened 2020-06-30
[BUG] RenderType is missing in flexform definition 


dgorges:
If a form only is selected in the backend, the option section in the flexform displays an unknown type select.
Unknown type: select, render type: select

The fix is just to add in the form.xml in Line 36: 
`<renderType>selectSingle</renderType>` in the config section.

After that the template can be set correctly

dkd-kaehm:
Backports merged as well. 
Can be closed.

PR #2669, opened 2020-07-06
[BUGFIX] Fix missing renderType attribute in flexform for search plugin

jacobsenj:
In the search plugin selection "form only" the renderType attribute is missing
in the flexform configuration which prevents rendering the custom template
select in the backend. This fixes it by adding the renderType attribute.

Fixes: #2661


dkd-kaehm:
Thanks @jacobsenj 
We need backports into 11 and 10.

Issue #2677, opened 2020-07-15
[BUG] Wrong typoscript path in autosuggest documentation


sneopu:
The path of "suggest" is wrong in Autosuggest.rst
It should not be inside "search"

dkd-kaehm:
Hello @sneopu,

Thanks for reporting. 
Currently I do not agree with you suggestion.
Why Do you think it is currently wrong?

See in `Configuration/TypoScript/Examples/Suggest/setup.txt` Liines 26 - 32:

https://github.com/TYPO3-Solr/ext-solr/blob/520d2efadc00496d1205ce2dd984bd8408919da0/Configuration/TypoScript/Examples/Suggest/setup.txt#L26-L32

sneopu:
It is also wrong there ...
In https://github.com/TYPO3-Solr/ext-solr/blob/520d2efadc00496d1205ce2dd984bd8408919da0/Configuration/TypoScript/Solr/setup.txt#L231 it is outside "search"
Also in TypoScriptConfiguration Class the path is without "search"    https://github.com/TYPO3-Solr/ext-solr/blob/3b52790e6458e43daab6964a13e157ee24327f38/Classes/System/Configuration/TypoScriptConfiguration.php#L1923

dkd-kaehm:
OOh,
would propose to move them all inside of `search`, because there is the right place for "suggestions".

Mabahe:
Search is search and suggest is suggest IMO. Please do not change this and simple please fix the examples. 
In `ext-solr/Configuration/TypoScript/Solr/setup.txt´ it is right! The wrong example was introduced with the fluid templating in 
https://github.com/TYPO3-Solr/ext-solr/commit/109b203aa8c5b70f497a84af2ac7aa8ced0f8191#diff-3894f9575cc36404a19ad30eb10a6e15R4

dkd-kaehm:
> Search is search and suggest is suggest IMO. Please do not change this and simple please fix the examples.
> In `ext-solr/Configuration/TypoScript/Solr/setup.txt´ it is right! The wrong example was introduced with the fluid templating in
> [109b203#diff-3894f9575cc36404a19ad30eb10a6e15R4](https://github.com/TYPO3-Solr/ext-solr/commit/109b203aa8c5b70f497a84af2ac7aa8ced0f8191#diff-3894f9575cc36404a19ad30eb10a6e15R4)

[...]

PR #2678, opened 2020-07-15
Fix path of suggest in typoscript settings

sneopu:
Fix the path of "suggest"

Fixes: #2677


dkd-kaehm:
Since @Mabahe does not agree with my suggestion to move `suggestion` into `search` path, at least the examples must be adjusted.

@sneopu could you please adjust the examples?

dkd-kaehm (approved):

PR #2681, opened 2020-07-22
[TASK] Drop TYPO3 9.5 LTS support for future release 11.1.x

dkd-kaehm:
Due to switching to new release strategy,
there will be no major EXT:Solr version 12 release up to TYPO3 version 12. 

Since now we'll strive for adjusting the releases in parallel to TYPO3 LTS releases.
So the EXT:Solr version 11.0 was the last one which supports two TYPO3 LTS versions.

Future "major" version of EXT:Solr will be the same as TYPO3 11 LTS version.


dkd-kaehm:
TYPO3 v 10.4.5 is broken, restart the jobs after 10.4.6 is released

Issue #2634, opened 2020-06-02
[BUG] Restricted pages are not being indexed in Typo3 10


robertschramm:
**Describe the bug**
The Solr page indexer is not able to index restricted pages (fe_user group set on page). Instead of a 200:found HTTP return and a json formatted result set we get a 403:forbidden and no json return. This is true with the latest TYPO3 version together with the latest Solr extension version.

**To Reproduce**
1. Install a new project with TYPO3 10 via composer mode
2. Choose to have a start page pre configured
3. Install Solr via composer
4. Perform table updates in maintenance module
5. Create a frontend user group via list module
6. Create a new page under your start page which has the access restricted to either your new frontend user group or to "Show at any login"
7. Configure the Solr connection via the site:configuration module, add the Typoscript includes as shown in https://docs.typo3.org/p/apache-solr-for-typo3/solr/11.0/en-us/GettingStarted/ConfigureExtension.html
8. Make sure that the connection to your Solr server is working via the Solr extension menue in TYPO3 backend
9. Now perform your first indexing via Solr extension menue. The restricted page you have created is not being indexed. An error is displayed for your restricted page in the section "Indexing errors"

**Expected behavior**
The restricted page should be indexed successfully.

**Screenshots**
An error (like in screenshot) is thrown by the page indexer.
<img width="1423" alt="New_TYPO3_site__TYPO3_CMS_10_4_3_" src="https://user-images.githubusercontent.com/66314043/83524391-d8cbfc80-a4e3-11ea-8de4-961ec193a87f.png">

The result of the request is a 403:forbidden page in TYPO3 10:
<img width="982" alt="Postman_typo3_10" src="https://user-images.githubusercontent.com/66314043/83526221-64468d00-a4e6-11ea-952e-3e7492a61269.png">

The result of the same reproduction steps in TYPO3 9 is:
<img width="978" alt="Postman_typo3_9" src="https://user-images.githubusercontent.com/66314043/83526246-6b6d9b00-a4e6-11ea-9eb2-d8f7168a90b7.png">

**Used versions:**
 - TYPO3 Version: [10.4.3]
 - EXT:solr Version: [11.0.1]
 - Used Apache Solr Version: [8.5.1]
 - PHP Version: [7.2.1]

**Additional context**
[...]

PR #2695, opened 2020-08-12
[BUGFIX] Restricted pages are not being indexed in Typo3 10 

dkd-kaehm:
Fixes: #2634

### DIFF
--- a/Classes/IndexQueue/FrontendHelper/PageIndexer.php
+++ b/Classes/IndexQueue/FrontendHelper/PageIndexer.php
@@ -1,4 +1,5 @@
 <?php
+
 namespace ApacheSolrForTypo3\Solr\IndexQueue\FrontendHelper;
 
 /***************************************************************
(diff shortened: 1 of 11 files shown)


dkd-kaehm:
@georgringer 
Could you please tests this PR within TYPO3 9.5 and 10.4?

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -0,0 +1,158 @@
...
+    public function process(ServerRequestInterface $request, RequestHandlerInterface $handler): ResponseInterface
+    {
+        if (Util::getIsTYPO3VersionBelow10()
+            || !$request->hasHeader(PageIndexerRequest::SOLR_INDEX_HEADER)
+        ) {
+            return $handler->handle($request);
+        }
+
+        // disable TSFE cache for TYPO3 v10
+        $request = $request->withAttribute('noCache', true);
```
georgringer, line 86: this won't work if no_cache is disabled via `LocalConfiguration.php`
I propose disabling the cache with a middleware/hook when indexing, something like https://github.com/studiomitte/powermail_freebie/blob/master/Classes/Hooks/CacheHook.php

Classes/IndexQueue/FrontendHelper/PageIndexer.php
```diff
@@ -69,33 +74,29 @@ class PageIndexer extends AbstractFrontendHelper implements SingletonInterface
     protected $responseData = [];
 
     /**
-     * @var \TYPO3\CMS\Core\Log\Logger
+     * @var Logger
```
georgringer, line 77: it makes reviewing really hard when you mix such CGL issues with bugfixing

Classes/Middleware/FrontendUserAuthenticator.php
```diff
@@ -0,0 +1,158 @@
...
+        return $request;
+    }
+
+    /**
+     * Gets the access rootline as defined by the request.
+     *
+     * @param PageIndexerRequestHandler $handler
+     * @return Rootline The access rootline to use for indexing.
+     */
+    protected function getAccessRootline(PageIndexerRequestHandler $handler): ?Rootline
```
georgringer, line 146:
```suggestion
    protected function getAccessRootline(PageIndexerRequestHandler $handler): Rootline
```

shouldn't that always return a Rootline object

Classes/Middleware/FrontendUserAuthenticator.php
dkd-kaehm: I'll create a new issue for that. We should clean up the old stuff anyway, there are multiple things, which try to disable caches.

georgringer:
@dkd-kaehm do you want to merge that?

dkd-kaehm:
Yes, incl. Backport onto 11.

dkd-kaehm:
@georgringer 
Do you have some pages with restrictions to the content elements? Does this patch work with them?

Issue #2307, opened 2019-04-29
[FEATURE] use assoc keys for tx_solr[filter] facet parameters instead of numeric index


dkd-kaehm:
**Troubles with numeric index:**
* depended on the users sequence of using facets, the numbers are always different, therefore:
  * it is impossible to make TypoScript conditions like `[globalVar = tx_solr|filter|0 = type:pages]`
  * speaking url configurations are complicated
  * different URLs parameters-set point to the same content.
  * etc. 

**Solution:**
* use `tx_solr[filter][facetName:facetValue]=1` construct for option based facet-filters
* about all ranges and other facet types the exact solution must be discussed here

**Target versions**
All versions prior 2019.05

**Notice**: 

Following snippet does a part without breaking something:
```
\ApacheSolrForTypo3\Solr\Domain\Search\SearchRequest::addFacetValue() {
        if ($this->getHasFacetValue($facetName, $facetValue)) {
            return $this;
        }

        $facetValues = $this->getActiveFacets();
        $facetValues[$facetName . ':' . $facetValue] = $facetName . ':' . $facetValue;
        $this->setActiveFacets($facetValues);

        $this->stateChanged = true;
        return $this;
}
```

PR #2705, opened 2020-08-27
[FEATURE] use assoc keys for tx_solr[filter] facet parameters instead of numeric index

3l73:
Introduce a new style how the facet array represented inside of the url with a combination of key and value in order to be able to keep a specific order.

# What this pr does

This pull request allows to change the url parameters from index based to associative keys for facets.
Using associative keys, the value of a facet will be replaced with 1. A value of 1 means, that the facet is active. A value of 0 means, the value is inactive.

Additionaly a new sort option for url parameters is available. The sort of parameters is mandatory for associative keys.

By default Solr behaves as before.

**Note:** This change affects the representation of facets inside of the URL strings.

# How to test

1. Include all possibile facet configuration in your SysTemplate: 

```
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/DateRange/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/Hierarchy/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/NumericRange/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/Options/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/OptionsFiltered/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/OptionsPrefixGrouped/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/OptionsSinglemode/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/OptionsToggle/setup.txt'
@import 'EXT:solr/Configuration/TypoScript/Examples/Facets/QueryGroup/setup.txt'
```
2. Index all pages and search for `*`
3. Select facets in following order
  1. Type
  2. Age
  3. PID Range

As a result the order of filter inside of the URL have to be:
  1. Age
  2. PID Range
  3. Type

* Now click on page number two. The order of parameters should not change.
* Remove facet PID Range.
As a result the order of filter inside of the URL have to be:
  1. Age
  2. Type

* Run all unit test. You should not see any errors.
* Run all integration test. You should not see any errors.

Fixes: #2307

### DIFF
--- a/Classes/Domain/Search/ResultSet/Facets/UrlFacetDataBag.php
+++ b/Classes/Domain/Search/ResultSet/Facets/UrlFacetDataBag.php
@@ -1,0 +1,406 @@
+<?php
+
+namespace ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\Facets;
+
+/**
+ * Copyright notice
+ *
+ * (c) 2020 Lars Tode <lars.tode@dkd.de>
+ * All rights reserved
+ *
+ * This script is part of the TYPO3 project. The TYPO3 project is
+ * free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * The GNU General Public License can be found at
+ * http://www.gnu.org/copyleft/gpl.html.
+ *
+ * This script is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * This copyright notice MUST APPEAR in all copies of the script!
+ */
+
+use ApacheSolrForTypo3\Solr\System\Util\ArrayAccessor;
+use ApacheSolrForTypo3\Solr\Utility\ParameterSortingUtility;
+
+/**
+ * Data bag for facets inside of an url
+ *
+ * @author Lars Tode <lars.tode@dkd.de>
+ * @api
+ */
+class UrlFacetDataBag implements \Countable
+{
+    /**
+     * Parameters array has a numeric index
+     */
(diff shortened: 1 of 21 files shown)


dkd-kaehm (changes requested):
Please refactor this snipped.
Please rebase to current master.

Classes/Domain/Search/ResultSet/Facets/UrlFacetDataBag.php
```diff
@@ -0,0 +1,406 @@
...
+use ApacheSolrForTypo3\Solr\System\Util\ArrayAccessor;
+use ApacheSolrForTypo3\Solr\Utility\ParameterSortingUtility;
+
+/**
+ * Data bag for facets inside of an url
+ *
+ * @author Lars Tode <lars.tode@dkd.de>
+ * @api
+ */
+class UrlFacetDataBag implements \Countable
```
dkd-kaehm, line 37: Would propose to name it "FacetContainer" instead.

dkd-kaehm:
@3l73 please squash and use message and Documentation/Configuration/Reference/TxSolrSearch.rst change from #2944

```
faceting.urlParameterStyle
~~~~~~~~~~~~~~~~~~~~~~~~~~

:Type: Option/String (index or assoc)
:TS Path: plugin.tx_solr.search.faceting.urlParameterStyle
:Since: 11.1
:Default: index

Allows to change the url style of facets. This can be legacy index or more modern associative style.

Index style: tx_solr[filter][0]=type:pages
Associative style: tx_solr[filter][type:pages]=1

faceting.urlParameterSort
~~~~~~~~~~~~~~~~~~~~~~~~~~

:Type: Boolean
:TS Path: plugin.tx_solr.search.faceting.urlParameterSort
:Since: 11.1
:Default: 0

Allows to enable sorting of url parameters, so the single state of facets is associated with same url, no matter in which order the facets were selected.
```

dkd-kaehm (approved):

PR #2709, opened 2020-08-31
[TASK] integrate githubs  probot / no-response

dkd-kaehm:
* will close the issues automatically if no response from creator is taken in 28 days
   after adding more-information-needed label


dkd-dobberkau:
I am fine with this. Thanks!

Issue #2257, opened 2019-02-14
[FEATURE] Support route enhancers of TYPO3


timohund:
**Is your feature request related to a problem? Please describe.**

It would be nice to support route enhancers of TYPO3 9 LTS in EXT:solr in order to support speaking urls. We need to check for which use cases this makes sence.

**Target versions**

10

dkd-kaehm:
Related: #2307

jproy418:
+1, Looking foward for this feature. Thanks for doing that!

dkd-kaehm:
@jproy418 
please checkout `upcoming-feature/Introduce_Route-Enhancer` branch and test it.
Make comments in #2755

PR #2715, opened 2020-09-04
[FEATURE] Introduce middleware and route enhancer for URLs

3l73:
# What this pr does

This pull request introduce route enhancer and aspects in order to create beaturiful URLs

# How to test

## Facet as URL path segment

1. Create a new site containing the Solr plugin for search results
2. Extend the site configuration with the route enhancer  (see example below)
   1. Define what facet should be used ath path segment
3. Index all required documents

- As a result you should see the combined facet values as one url sgement.
- If nof facet is selected you should not get an error

## Mask filter inside of the URL

1. Follow the first two steps of test case `Facet as URL path segment`
2. Enable the mask option inside of the router configuration
3. Configure facets you want to mask (see restrictions in the example below)
4. Index all required documents

- Configured facets are concat together and does not have the prefix `tx_solr[filter][]`
- Not configured facets are concat together and still have the prefix `tx_solr[filter][]`
- This changes effect on
 1. Facet links
 2. Page links

Example for the router configuration:
```yaml
routeEnhancers:
  Products:
    type: CombinedFacetEnhancer
    limitToPages:
      - 10
    extensionKey: tx_solr
    solr:
      # The plugin namespace used
      # @see https://docs.typo3.org/p/apache-solr-for-typo3/solr/master/en-us/Configuration/Reference/TxSolrView.html#pluginnamespace
      pluginNamespace: 'tx_solr'
      multiValueSeparator: '+'
      ignoreUrlKeys: ['no_cache', 'cHash', 'id', 'MP']
      replaceCharacters:
        ' ': '_'
      query:
        # You can tell Solr to mask query facets. This feature require the map below
        #
        # For example you have following filter:
        #   tx_solr[filter][0]=color:red&tx_solr[filter][1]=color:green&tx_solr[filter][2]=color:yellow
        # Mask will:
        # 1. implode all values into a single string and sort it -> green,red,yellow
        # 2. replace tx_solr[filter][]=color: with color=
        #
        # As a result the query will modified into:
        # color=green,red,yellow
        #
        mask: true

        # To reduce the amount of parameters you can force Solr to concat the values.
        # For example you have following filter:
        #   tx_solr[filter][0]=color:red&tx_solr[filter][1]=color:green&tx_solr[filter][2]=color:yellow
        # Concat will:
        # 1. collect all filters of the same type
        # 2. will sort all filter values alpha numeric
        # 3. join the values together
        #
        # As a result the query will modified into:
        # tx_solr[filter][0]=color=green,red,yellow
        #
        # Note: If you actived the mask option, the concat feature turn on automatically
        #
        concat: true

        # valueSeparator: ','

        # In order to simplify a filter argument, you have to define a corresponding map value
        # There is no automatically reduction of filter parameters at the moment available.
        # The key is the name of your facet, the value what use instead.
        #
        # Important:
        # There are some restrictions for the values. The use of TYPO3 core parameters is prohibited.
        # This contains at the moment following strings: no_cache, cHash, id, MP, type
        map:
          color: color
          taste: taste
          product: product
    routePath: '/{type}'
    # Note: All arguments inside of namespace tx_solr. See -> extensionKey
    # Example: Argument 'type' define as 'filter-type' will convert into 'tx_solr/filter-type'
    _arguments:
      type: filter-type
    requirements:
      type: '.*'
``` 

Fixes: #2257


3l73:
Closed in favor for https://github.com/TYPO3-Solr/ext-solr/pull/2755

PR #2718, opened 2020-09-07
Prevent duplicate urls for page 0

DanielSiepmann:
Reset page argument if set to 0.
As 0 is the default, there is no need to have two different states to represent this.
By resetting, the array will not return the info and links to page 0 will not include the page argument.
This prevents two separate urls to the same source.

# What this pr does

Reset page internally when being set to zero.

# How to test

Create a search result with pagination. Open another page then the first.
Check the generated url for the first page, it should no longer contain the page argument.
Before this change it would contain the argument with value 0.

Fixes: #2719

### DIFF
--- a/Classes/Domain/Search/SearchRequest.php
+++ b/Classes/Domain/Search/SearchRequest.php
@@ -421,7 +421,11 @@ class SearchRequest
     {
         $this->stateChanged = true;
         $path = $this->prefixWithNamespace('page');
-        $this->argumentsAccessor->set($path, $page);
+        if ($page === 0) {
+            $this->argumentsAccessor->reset($path);
+        } else {
+            $this->argumentsAccessor->set($path, $page);
+        }
         return $this;
     }
 
--- a/Classes/Domain/Search/SearchRequest.php
+++ b/Classes/Domain/Search/SearchRequest.php
@@ -421,10 +421,10 @@ class SearchRequest
     {
         $this->stateChanged = true;
         $path = $this->prefixWithNamespace('page');
+        $this->argumentsAccessor->set($path, $page);
+        // use initial url by switching back to page 0
         if ($page === 0) {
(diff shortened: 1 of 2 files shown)


dkd-kaehm:
@DanielSiepmann thanks for PR, there is the test case implemented for this behavior. We should look in git history to know, why we did so previously.

DanielSiepmann:
The test was introduced with 520d2efadc00496d1205ce2dd984bd8408919da0, I guess it just covered the existing functionality without further thoughts. I've adjusted the test to match the new behaviour. Feel free to dig deeper and provide feedback or adjust.

Classes/Domain/Search/SearchRequest.php
```diff
@@ -421,7 +421,11 @@ public function setPage($page)
     {
         $this->stateChanged = true;
         $path = $this->prefixWithNamespace('page');
-        $this->argumentsAccessor->set($path, $page);
+        if ($page === 0) {
+            $this->argumentsAccessor->reset($path);
+        } else {
+            $this->argumentsAccessor->set($path, $page);
+        }
```
dkd-kaehm, line 428:
```suggestion
        $this->argumentsAccessor->set($path, $page);
        // use initial url by switching back to page 0
        if ($page === 0) {
            $this->argumentsAccessor->reset($path);
        }
```

dkd-kaehm: Please apply if you think, the suggestion makes sense.

dkd-kaehm (approved):
Code looking good for me. 
Please apply my suggestion, if it makes sense.

dkd-kaehm:
@DanielSiepmann could you please backport it onto 11, if you need it 11? 
Please use squashed commit from master for that. 
I'll merge it now.

Issue #2719, opened 2020-09-07
[BUG] Do not generate two different urls for initial page


DanielSiepmann:
**Describe the bug**
When adding pagination, there are two urls for the first page. One can either omit the page argument, or set it to 0.
The 2nd one is the default for pagination links, leading to two different urls.

**To Reproduce**
1. Go to a search result with pagination
2. Click on page 2.
3. Click on page 1.
4. You'll see a page argument in the url set to 0.

**Expected behavior**
I would expect the same urls as I had initially, without a page argument set to 0.

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.6
 - Browser: Firefox
 - EXT:solr Version: 11.0.2
 - Used Apache Solr Version: 8.5.1
 - PHP Version: 7.3.21-1+ubuntu20.04.1+deb.sury.org+1 
 - MySQL Version: 8.0.21

PR #2726, opened 2020-09-14
[BUGFIX] Prevent duplicate urls for page 0 (#2718)

DanielSiepmann:
Reset page argument if set to 0.
As 0 is the default, there is no need to have two different states to represent this.
By resetting, the array will not return the info and links to page 0 will not include the page argument.
This prevents two separate urls to the same source.

# What this pr does

Reset page internally when being set to zero.

# How to test

Create a search result with pagination. Open another page then the first.
Check the generated url for the first page, it should no longer contain the page argument.
Before this change it would contain the argument with value 0.

Fixes: #2719

### DIFF
--- a/Classes/Domain/Search/SearchRequest.php
+++ b/Classes/Domain/Search/SearchRequest.php
@@ -422,6 +422,10 @@ class SearchRequest
         $this->stateChanged = true;
         $path = $this->prefixWithNamespace('page');
         $this->argumentsAccessor->set($path, $page);
+        // use initial url by switching back to page 0
+        if ($page === 0) {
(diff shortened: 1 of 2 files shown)


DanielSiepmann:
Backport of #2718 from master to v11.

DanielSiepmann:
I don't see a relation between failing test and this PR.
Maybe it is related to changes within TYPO3 10.4.8 release?

dkd-kaehm:
> I don't see a relation between failing test and this PR.
> Maybe it is related to changes within TYPO3 10.4.8 release?

yep, #2727

Issue #2523, opened 2019-12-29
[TASK] Restore note on enabling index in EXT:solr 10 documentation


saitho:
I just set up Solr for TYPO3 for the first time. I'm using site configuration for configuring Solr.
I had the issue that the page indexer wouldn't work no matter what I did.

Turned out that I had to enable `config.index_enable` in TypoScript setup, which was not in the Getting Started documentation. I just noticed that it was still present in 9.0 documentation.

If anyone else can confirm that this setting is still needed, we should restore that part in the documentation for 10.

Tuurlijk:
Just wasted a good couple of hours on this too. Since this issue has been open for so long . . .  seems we need to fix it ourselves. Will you do a pr? or should I make one?

dkd-kaehm:
Hi, 
anyone can create PRs. Please note if you do, you need PRs/Backports from release-10.x to master.

dkd-kaehm:
@Tuurlijk 
Thanks for your PRs.
We need it in release-11.x branch as well.

PR #2736, opened 2020-09-25
[DOCS] Note that config.index_enable is still needed - release 10

Tuurlijk:
# What this pr does

Makes it clear that `config.index_enable = 1` is still required to index anything.

# How to test

Setup a clean install and try to index any content without `config.index_enable = 1`

Fixes: #2523


dkd-kaehm (approved):
:+1:

Issue #2743, opened 2020-10-07
[Bug] In a recursive SOLR_RELATION only one element from tt_content gets indexed into the solr core


haraldwitt:
I have a custom table "offers" which contains a field "product_uid" pointing to the product table.
I have the other custom table "products" with the field "content_elements" pointing to tt_content. 
"content_elements" is an inline element in the same manner tx_news it does.
Now I want to index the content elements as SOLR_CONTENT with an indexing configuration for "offers". Why that? The reason is, that "offers" contains an endtime-field which is very important for the search results!
My index configuration looks as follows:
```
plugin.tx_solr.index.queue {
	offers = 1
	offers {
		table = offers
		fields {
			...
			content = SOLR_CONTENT
			content.cObject = COA
			content.cObject {
				20 = SOLR_RELATION
				20 {
					localField = product_uid
					foreignLabelField = content_elements.bodytext
					singleValueGlue = | |
				}
			}
			...
		}
	}
}
```
Using this, only the bodytext of the LAST related record of tt_content will appear in the content-field of the solr core_de.

But when indexing "products", the bodytext of ALL related records will be found in the solr core_de.
The configuration for products:
```
plugin.tx_solr.index.queue {
	products = 1
	products {
		table = products
		fields {
			...
			content = SOLR_CONTENT
			content {
				cObject = COA
				cObject {
					10 = SOLR_RELATION
					10 {
						localField = content_elements
						foreignLabelField = bodytext
						singleValueGlue = | |
					}
				}
			}
			...
		}
	}
}
```
But this is not suittable for me because of the missing endtime-field in products.

haraldwitt:
Ok, I found, this is duplicate to issue #2147
https://github.com/TYPO3-Solr/ext-solr/pull/2147

haraldwitt:
Ok, I got it!

Class:	ApacheSolrForTypo3\Solr\ContentObject\Relation
Line 	325++
Method:	resolveRelatedValue

There you con find the following:
            // recursion
            $relatedItemsFromForeignTable = $this->getRelatedItemsFromForeignTable(
                $foreignTableName,
                $relatedRecord['uid'],
                $foreignTableTca['columns'][$foreignTableLabelField],
                $parentContentObject
            );
            $value = array_pop($relatedItemsFromForeignTable);

Obviously the array_pop() makes the problem. 
I replaced array_pop()-line with the following:
[...]

PR #2745, opened 2020-10-08
[BUGFIX] Recursive SOLR_RELATION only one element from tt_content gets indexed into the solr core

haraldwitt:
[Bugfix] In a recursive SOLR_RELATION only one element from tt_content gets indexed into the solr core #2743

# What this pr does

Please add a description here

# How to test

Fixes: #2743


dkd-kaehm (changes requested):

dkd-kaehm:
Currently the fix breaks other functionality.
See the build integration test error: https://travis-ci.org/github/TYPO3-Solr/ext-solr/jobs/733968309#L885

We should check if we really need the `$value = serialize($relatedItemsFromForeignTable);` in else condition.

Issue #2746, opened 2020-10-12
[BUG] Deprecated second parameter for BackendUserAuthentication->modAccess is used


NamelessCoder:
**Describe the bug**
Deprecated second parameter for BackendUserAuthentication->modAccess is used in IfHasAccessToModuleViewHelper.

**To Reproduce**
1. Be on at least TYPO3 9.5
2. Enable reporting of deprecation errors (trigger_error + E_USER_DEPRECATED)
3. Visit the Solr "Info" backend module
4. An error is triggered

**Expected behavior**
No error should be triggered; the return value from modAccess() can be cast to boolean which results in identical behavior.

**Used versions (please complete the following information):**
 - TYPO3 Version: 9.5.21
 - Browser: Chrome (but irrelevant)
 - EXT:solr Version: 11.0.2
 - Used Apache Solr Version: irrelevant, triggers on any version
 - PHP Version: irrelevant, triggers on any version (if configured to not suppress E_USER_DEPRECATED)
 - MySQL Version: irrelevant

3l73:
A fix for the master branch is ready to review.

Backports will be available after review and test.

PR #2749, opened 2020-10-16
[BUGFIX] Removes secondary parameter #2746

3l73:
Calls of method BackendUserAuthentication::modAccess with two arguments deprecated since TYPO3 9.5, only one parameter is available in TYPO3 10.4

Resolves: #2746

# What this pr does

This pull request removes the second parameter in call of method BackendUserAuthentication::modAccess.
This parameter is deprecated since TYPO3 9.5 and dropped in version 10.4.

https://github.com/TYPO3/TYPO3.CMS/blob/bf49d5eab37c660fce310905e18f8880ac378cae/typo3/sysext/core/Classes/Authentication/BackendUserAuthentication.php#L447

# How to test

1. Be on at least TYPO3 9.5
2. Enable reporting of deprecation errors (trigger_error + E_USER_DEPRECATED)
3. Visit the Solr "Info" backend module
4. An error is triggered

_Expected behavior_
No error should be triggered; the return value from modAccess() can be cast to boolean which results in identical behavior.

Fixes: #2746

### DIFF
--- a/Classes/ViewHelpers/Backend/Security/IfHasAccessToModuleViewHelper.php
+++ b/Classes/ViewHelpers/Backend/Security/IfHasAccessToModuleViewHelper.php
@@ -61,18 +61,32 @@ class IfHasAccessToModuleViewHelper extends AbstractConditionViewHelper
         $this->registerArgument('signature', 'string', 'The full signature of module. Simply mainmodulename_submodulename in most cases.');
     }
 
-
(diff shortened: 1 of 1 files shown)


dkd-kaehm (changes requested):

dkd-kaehm (approved):
:+1:

PR #2754, opened 2020-10-22
[DOCS] Fix TS lastSearches path

FearFreddy:
# What this pr does

Just a correction of a wrong typoscript path in TxSolrSearch.rst which stole some minutes of my day.


dkd-kaehm:
Thanks for the fix.
We need this in older versions as well.

dkd-kaehm:
Backported to 10 and 11

Issue #2756, opened 2020-10-23
[BUG] TypeError in ApacheSolrForTypo3\Solr\System\Url\UrlHelper::getScheme()


DrWh0286:
**Describe the bug**
During page indexing if no valid url is present in
`\ApacheSolrForTypo3\Solr\Domain\Index\PageIndexer\Helper\UriBuilder\AbstractUriStrategy::getPageIndexingUriFromPageItemAndLanguageId()`
line 101 a log entry should be written. In case of missing url scheme the following error is thrown:

`Return value of ApacheSolrForTypo3\Solr\System\Url\UrlHelper::getScheme() must be of the type string, null returned`

I don't know the reason, why there is no valid url, but I think the type error can be solved by ensuring correct return values in ApacheSolrForTypo3\Solr\System\Url\UrlHelper methods.

**To Reproduce**
Having an item with invalid url, which has no url scheme. This causes, that the scheme is not set in array `\ApacheSolrForTypo3\Solr\System\Url\UrlHelper::$urlParts`. So there is not a string returned from `getScheme()`, but null.

**Expected behavior**
The method `ApacheSolrForTypo3\Solr\System\Url\UrlHelper::getScheme()` should return an empty string.

**Used versions (please complete the following information):**
 - TYPO3 Version: 9.5.22
 - Browser: n/a
 - EXT:solr Version: 11.0.3
 - Used Apache Solr Version: 8.5.2
 - PHP Version: 7.3.19-1~deb10u1
 - MySQL Version: 5.7.32

3l73:
@maritwho we are planing to refactor class UrlHelper in order to implement the PSR-7 standard. It will base on class `\TYPO3\CMS\Core\Http\Uri`. The PSR-7 interface itself did not define return types.

-> #2776 

Did you figure out which URL did lead to this issue?

dkd-kaehm:
Fixed in Issue #2776 -> PR #2790

PR #2757, opened 2020-10-23
[BUGFIX] Fix type error in UrlHelper

DrWh0286:
If some url part is not set the getters have to return
an emtpy string.

Fixes: #2756


dkd-kaehm:
Fixed in Issue #2776 -> PR #2790

dkd-kaehm:
We must Port the `Tests/Unit/System/Url/UrlHelperTest.php` and `Classes/Report/SiteHandlingStatus.php` to master.

PR #2758, opened 2020-10-23
[BUGFIX] Fix type error in UrlHelper

maritwho:
If some url part is not set the getters have to return
an emtpy string.

Fixes: #2756


dkd-kaehm:
Thanks for PR.
Theoretically it should be as in your PR, but in which context are the methods in use?
What about the parts, which are using these methods and expect to get the valide uri parts or fail?

dkd-kaehm:
Please provide exact context, why this PR is needed, the type incompatibility is sometimes a part of design/pattern, even they are bad.

Issue #2761, opened 2020-10-28
[BUG] Pass feUser as missing fifth parameter to TypoScriptFrontendController


Konafets:
**Describe the bug**
With TYPO3 v10 the `TypoScriptFrontendController::_construct()` expects the fifth parameter to be a cHash or an object of `FrontendUserAuthentication`. Otherwise the property `TypoScriptFrontendController->fe_user` is not initialized and as an result the indexing will fail when indexing a language with an ID > 0 

**To Reproduce**
1. Have a multi language setup
2. Add the news records to the queue 
3. Click on 'Index now'  and wait.

**Expected behavior**
I expect the indexing without any errors

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.9
 - Browser: Chrome
 - EXT:solr Version: 11.0
 - Used Apache Solr Version: 8.5
 - PHP Version: 7.4
 - MySQL Version: 5.7

dkd-kaehm:
We need Backport on to 11.

PR #2762, opened 2020-10-28
[BUGFIX] Pass `FrontendUserAuthentication` to `TypoScriptFrontendController`

Konafets:
# What this pr does

The indexing fails when the `TypoScriptFrontendController` is constructed w/o a frontend user in `TypoScriptFrontendController::initUserGroups()` on line 930 when the visibility `includeHiddenContent` assigned to `$this->fe_user->showHiddenRecords;` This PR passes the `$fe_user` to the `TypoScriptFrontendController::_construct()`.

Fixes: #2761

### DIFF
--- a/Classes/FrontendEnvironment/Tsfe.php
+++ b/Classes/FrontendEnvironment/Tsfe.php
@@ -81,11 +81,12 @@ class Tsfe implements SingletonInterface
 
 
         if (!isset($this->tsfeCache[$cacheId])) {
-
+            $feUser = GeneralUtility::makeInstance(FrontendUserAuthentication::class);
+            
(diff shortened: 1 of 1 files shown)


djcprinse:
I can confirm this patch resolves the issues we face where the Solr indexer is stuck at indexing on the error `Call to undefined method stdClass::fetchGroupData()` at the [TyposcriptFrontendController](https://github.com/TYPO3/TYPO3.CMS/blob/c27d578fb7cef48a60a086cd933dca42506d2e79/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L932)

dkd-kaehm (approved):

PR #2763, opened 2020-10-28
[FEATURE] RFC: Enable dataProcessing on Results

cirk2:
# What this pr does

Typo3 7.4 Introduced Data Processing for the `FLUIDTEMPLATE` cObject. I have found it to be quite useful but it is not globally available in extensions.
This PR Enables the feature on the Results Action and should be easy to be expanded to the other actions in the `SearchController`.
Location of the `dataProcessing` block in the setup is just my first idea, and retrieval of the value probably should be packed into a getter.

# How to test

`Configuration/TypoScript/Examples/DataProcessing/setup.txt` has an example for TypoScript setup. A DataProcessor has to be Provided for testing, either one of the ones Supplied by Typo3 or a custom one.


Classes/Controller/SearchController.php
```diff
@@ -36,6 +36,16 @@ class SearchController extends AbstractBaseController
...
     protected $view;
 
+    /**
+     * @var ContentDataProcessor
+     */
+    protected $contentDataProcessor;
+
+    public function __construct()
+    {
+        $this->contentDataProcessor = GeneralUtility::makeInstance(ContentDataProcessor::class);
```
dkd-kaehm, line 46: Please use Injector style, to be able to mock this stuff in tests.

Classes/Controller/SearchController.php
```diff
@@ -113,6 +123,11 @@ public function resultsAction()
             ];
 
             $values = $this->emitActionSignal(__CLASS__, __FUNCTION__, [$values]);
```
dkd-kaehm, line 125: Is this signal unusable for the things you want?
If there is not all what you want, pack this in array.
Would propose to extend all the signals and pack the stuff in the docs and examples.

dkd-kaehm (commented):
Could you please explain some exact use case for this stuff?

Classes/Controller/SearchController.php
cirk2: Main difference is that Data Processing is configured via typoscript, allowing to plug different configs in places.
Instead of Signals which are global.

dkd-kaehm: No, i mean that we could plug this stuff via this signal[s], via something like "Facade/Gateway" similar design pattern.

cirk2: The `ContentDataProcessor` only needs the `ContentObjectRenderer` which we can get pretty much anywhere injection works. And the configuration from Typoscript of course.
So yeah we could plug it via the signals.

dkd-kaehm: FYI: signal/slot is deprecated
https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ApiOverview/Hooks/EventDispatcher/Index.html#eventdispatcher

PR #2778, opened 2020-11-06
[TASK] move to travis stages

dkd-kaehm:
# What this Pull-Request does:

An experiment for introduction of [Travis build stages](https://docs.travis-ci.com/user/build-stages/). 
This should reduce the build and fail time.
The building and testing of Docker Image functionality takes about 4 minutes and can be checked once, because they are the same. Also the build&test of Docker image can be omitted by PHP/TYPO3 Version combinations.
As Idea the Travis build stages and workspaces should be used to save 3-4 minutes for each PHP/TYPO3 Version combination.
Moreover the Apache-Solr Server bootstraping by reusing the build image, can save additional 3-4 minutes.

This is currently as draft to define all the needed stages and workspaces as first step.

In second step we will push the artefacts in workspaces to be able to share them between the stages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-dist: bionic
+dist: focal
 
 language: php
 
@@ -6,8 +6,6 @@ services:
   - mysql
 
 php:
-  - 7.2
-  - 7.3
   - 7.4
 
 jdk:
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
Closing in favour of #2781, due to travis-ci is not triggering.

Issue #2759, opened 2020-10-27
[BUG] Use correct variable in FAQ code example


julianhofmann:
Code example is setting `solr_port_read: '%env(SOLR_HOST)%'`. This should obviously be `solr_port_read: '%env(SOLR_PORT)%'`

PR #2780, opened 2020-11-06
[DOCS:PORT:MASTER] Use *_PORT variable for setting the port

dkd-kaehm:
Fixes: #2759


dkd-kaehm:
No needs to wait for Build, can be merged immediately.

Issue #2635, opened 2020-06-03
[Task] Map managed resources to core-name instead of schema.


web-commerce-og:
**Describe the bug**
When creating new cores via the suggested command in the docs, multiple cores use a common configset. All of the cores using the same config-set **and** also the same language share their synonyms and stopwords. This might lead to unexpected results when searching.

**To Reproduce**
1. Create at least two cores with this command, using the samt config-set and language
```
curl 'http://host:port/path-to-solr/admin/cores?action=CREATE&name=core1_de&configSet=ext_solr_8_0_0&schema=german/schema.xml&dataDir=../../data/german'
curl 'http://host:port/path-to-solr/admin/cores?action=CREATE&name=core2_de&configSet=ext_solr_8_0_0&schema=german/schema.xml&dataDir=../../data/german'
```
2. create/modify synonyms e.g. via the backend module and reload the cores
3. see the synonyms of site A in site B and vice versa e.g. in the Backend Module.

**Expected behavior**
It would be useful, if the path to the synonym and stopword files could include the core name or something similiar, so that every core could use its own lists. If I want to use the same lists, one could use a core for multiple sites.

**Used versions:**
 - TYPO3 Version: 9.5.18
 - Browser: Firefox
 - EXT:solr Version: 10.0.1
 - Used Apache Solr Version: 8.2.0
 - PHP Version: 7.3
 - MySQL Version: 5.5

dkd-kaehm:
Related: #579
@dkd-friedrich 
Use this issue by ideas about "[TASK] Enable core specific synonyms and stop words" #579.

dkd-kaehm:
May be possible via https://lucene.apache.org/solr/guide/8_5/configuring-solrconfig-xml.html#substituting-properties-in-solr-config-files

PR #2794, opened 2020-11-19
 [Task] Map managed resources to core-name instead of schema

dkd-kaehm:
Changes the mappings of managed resources from schema to core name

Fixes:#2635


dkd-kaehm (commented):
Following suggestion should be checked.

Issue #2797, opened 2020-11-20
[BUG] garbage collector with translations


achimfritz:
**Describe the bug**

delete a translated page, (or set it to hidden), deletes solr-documents in all languages and delete index queue item

**To Reproduce**
1. Solr-Configuration with two languages and two cores
2. page with translation
3. page-document indexed in both cores
4. set translation to hidden
5. index queue item is deleted
6. page-document is deleted in both cores

**Expected behavior**

* should not remove IndexQueue Item
* should not remove Pages of other languages

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4
 - EXT:solr Version: dev-master

dkd-kaehm:
Backports in 10 and 11 must be created.

dkd-kaehm:
PR incl. backports onto 10 and 11 is merged.

PR #2798, opened 2020-11-20
[BUGFIX] garbage collector on translations #2797

achimfritz:
* should not remove IndexQueue Item
* should not remove Pages of other languages

Resolves: #2797

Fixes: #2797


dkd-kaehm (approved):
:+1:

Issue #2432, opened 2019-09-28
[FEATURE] Recursive page exclude


christophlehmann:
**Is your feature request related to a problem? Please describe.**

 I need the possibility to exclude subpage trees. Currently its possible with `hidden=1` + `extendToSubpages=1`, but i won't hide them. Think about a subpage tree with demo pages, system error pages, login, logout, etc.

**Describe the solution you'd like**

I think of a new pagefield `no_search_recursive` or maybe better extending `no_search` to have values like

0 = index
1 = no index
2 = no index recursive

**Describe alternatives you've considered**

I would not abuse the combination of no_search + extendToSubpages because from core side of view they do not have a direct relation.

**Target versions**

9.X

christophlehmann:
@timohund Are you up to the change with extending the no_search field with an recursive option for 9.x Branch? It would also fix #2433

timohund:
@christophlehmann i would propose to introduce that with the next major version (10 => current dev-master) or later. Are you able to prepare a pull request including tests for that?

christophlehmann:
@timohund I need the patch for this issue in 9.x

The new logic in this issue and #2433 ist 90% the same. 

I would provide a patch for #2433 with target version 9.x + master. The changed due to this issue  would go into master and i will use a local patch in my project. For tests i may need a little help of you, depending how much time my changes need. Deal? :-)

dkd-kaehm:
Todo: 
* Port #2804
* Add tests
* Add docs

PR #2804, opened 2020-12-02
[FEATURE] Exclude sub entries of page/storage recursively

dkd-kaehm:
* adds `no_search_sub_entries` field  to page and storage folder
* updates the sub entries of page/storage marked as `no_search_sub_entries`

Todo: Add documentation incl. screenshots

Resolves: #2432

### DIFF
--- a/Classes/Access/Rootline.php
+++ b/Classes/Access/Rootline.php
@@ -140,6 +140,7 @@ class Rootline
         $pageId,
         $mountPointParameter = ''
     ) {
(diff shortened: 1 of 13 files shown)


dkd-friedrich (approved):
LGTM

As soon as we implement this feature in the current versions, we should check the handling of differing no_search_sub_entries settings in a page tree and mount points. A notice about excluded sub pages could also be helpful.

PR #2806, opened 2020-12-04
[BUGFIX] Use rawurldecode on facets to handle (replace #1657)

3l73:
# What this pr does

The test case might be a bit weak and this PR does not regard any *en*coding of those values, since we're not using them in our projects, so I can't attest to it. However, this fixes my issues, so I'm looking for feedback…

# How to test

Replace: #1657
Fixes: #1656


dkd-kaehm (approved):
:+1:

dkd-kaehm (approved):

PR #2808, opened 2020-12-04
[TASK] Wrap code with anonymus function

3l73:
# What this pr does

To avoid leaking variables the code wraps with anonymous functions.
This is part of TYPO3 best practice.

Fixes: #1493


dkd-kaehm (approved):
:+1:

Issue #2810, opened 2020-12-14
[BUG] Solr breaks chinese characters


Konafets:
**Describe the bug**
Our site supports (traditional and simplified) chinese. While indexing some pages errored in the Solr index queue module but there was no error in the logs. Here is what happened. 

Solr fetches the page and extracts the content from it. The method `preg_replace()` breaks some unicode chars and I can see the famous � sign in the content _after_ the content was processed. Later, when the content is converted to JSON, it breaks silently by returning an empty string because. Guzzle returns an empty response and Solr marks the page as an error. 

To fix this I added the `u` modifier to switch the Unicode support on for `preg_replace()`. From the [PHP manual](https://www.php.net/manual/de/reference.pcre.pattern.modifiers.php):

> u (PCRE_UTF8)
> This modifier turns on additional functionality of PCRE that is incompatible with Perl. Pattern and subject strings are treated as UTF-8. An invalid subject will cause the 
> preg_* function to match nothing; an invalid pattern will trigger an error of level E_WARNING. Five and six octet UTF-8 sequences are regarded as invalid since PHP 
> 5.3.4 (resp. PCRE 7.3 2007-08-28); formerly those have been regarded as valid UTF-8.

**To Reproduce**
1. Have a site with this string "媒体和投资者 新闻 财务报告 公司股票信息 概览"
2. Index the page
3. Go to the Index queue page module and find an error

**Expected behavior**
A page with chinese characters gets indexed.

**Used versions (please complete the following information):**
 - TYPO3 Version: [e.g. 10.4.10]
 - Browser: doesn't matter
 - EXT:solr Version: [e.g. 11.0.0]
 - PHP Version: [e.g. 7.4.0]

dkd-kaehm:
@Konafets 
thanks for reporting and fixing this issue.

PR #2811, opened 2020-12-14
Don't break chinese unicode characters

Konafets:
# What this pr does

Our site supports (traditional and simplified) chinese. While indexing some pages errored in the Solr index queue module but there was no error in the logs. Here is what happened. 

Solr fetches the page and extracts the content from it. The method `preg_replace()` breaks some unicode chars and I can see the famous � sign in the content _after_ the content was processed. Later, when the content is converted to JSON, it breaks silently by returning an empty string because. Guzzle returns an empty response and Solr marks the page as an error. 

To fix this I added the `u` modifier to switch the Unicode support on for `preg_replace()`. From the [PHP manual](https://www.php.net/manual/de/reference.pcre.pattern.modifiers.php):

> u (PCRE_UTF8)
> This modifier turns on additional functionality of PCRE that is incompatible with Perl. Pattern and subject strings are treated as UTF-8. An invalid subject will cause the 
> preg_* function to match nothing; an invalid pattern will trigger an error of level E_WARNING. Five and six octet UTF-8 sequences are regarded as invalid since PHP 
> 5.3.4 (resp. PCRE 7.3 2007-08-28); formerly those have been regarded as valid UTF-8.

Fixes: #2810

### DIFF
--- a/Classes/Typo3PageContentExtractor.php
+++ b/Classes/Typo3PageContentExtractor.php
@@ -134,7 +134,7 @@ class Typo3PageContentExtractor extends HtmlContentExtractor
         // clean content
         $content = self::cleanContent($content);
         $content = trim($content);
-        $content = preg_replace('!\s+!', ' ', $content); // reduce multiple spaces to one space
+        $content = preg_replace('!\s+!u', ' ', $content); // reduce multiple spaces to one space
 
         return $content;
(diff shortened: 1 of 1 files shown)


Konafets:
Closing in favor of #2812

PR #2812, opened 2020-12-14
[BUGFIX] Enable unicode when fetching pages

Konafets:
# What this pr does

Our site supports (traditional and simplified) chinese. While indexing some pages errored in the Solr index queue module but there was no error in the logs. Here is what happened. 

Solr fetches the page and extracts the content from it. The method `preg_replace()` breaks some unicode chars and I can see the famous � sign in the content _after_ the content was processed. Later, when the content is converted to JSON. The empty content is than sent back to Guzzle. Solr considers a response with empty content as error and therefore marks the page as an error. 

To fix this I added the `u` modifier to switch the Unicode support on for `preg_replace()`. From the [PHP manual](https://www.php.net/manual/de/reference.pcre.pattern.modifiers.php):

> u (PCRE_UTF8)
> This modifier turns on additional functionality of PCRE that is incompatible with Perl. Pattern and subject strings are treated as UTF-8. An invalid subject will cause the 
> preg_* function to match nothing; an invalid pattern will trigger an error of level E_WARNING. Five and six octet UTF-8 sequences are regarded as invalid since PHP 
> 5.3.4 (resp. PCRE 7.3 2007-08-28); formerly those have been regarded as valid UTF-8.

Fixes: #2810

### DIFF
--- a/Classes/Typo3PageContentExtractor.php
+++ b/Classes/Typo3PageContentExtractor.php
@@ -134,7 +134,7 @@ class Typo3PageContentExtractor extends HtmlContentExtractor
         // clean content
         $content = self::cleanContent($content);
         $content = trim($content);
-        $content = preg_replace('!\s+!', ' ', $content); // reduce multiple spaces to one space
+        $content = preg_replace('!\s+!u', ' ', $content); // reduce multiple spaces to one space
 
         return $content;
     }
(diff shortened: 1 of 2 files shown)


dkd-kaehm (approved):
Could you please provide Backport PRs with 
"[BUGFIX:BACKPORT:10] ..." and 
"[BUGFIX:BACKPORT:11] ..." in
Commit and PR titles?

dkd-kaehm:
The back ports are created.

dkd-kaehm (approved):

Issue #2813, opened 2020-12-18
[TASK] Change configuration files to TYPO3 file extensions


3l73:
**What should be done in the scope of this task?**

Currently the file extension `txt` is in use for TypoScript configuration and the fie extension `typoscript` for TSConfig files.

Following Changes should be done:

1. Change the file extension for TypoScript from `txt` into `typoscript`
2. Change the file extension for TSconfig from `typoscript` into `tsconfig`

The old files should be in place and include the new file names.

As a result, editor settings can applied to configuration files without effecting other files.

3l73:
@dkd-kaehm Please merge in case all tests are green.

PR #2815, opened 2020-12-18
[TASK:BACKPORT:11] Change configuration files to TYPO3 file extensions

3l73:
# What this pr does

This Pull-Request change the file extensions from TypoScript and TSconfig files.
Previously TypoScript files use the file extension `txt` which is replaced with `typoscript`.
The TSconfig file used `typoscript` which is replaced with `tsconfig`. According to coding guidelines the TSconfig file was moved from `Configuration/TSconfig/ContentElementWizard.typoscript` to `Configuration/TSconfig/Page/Mod/Wizards/NewContentElement.tsconfig`.

Hints within code and documentations are updated according to these changes.

All changes them self should no effect current installations, since the old files are still in place and include the new ones. 

*Note:* There are no changes to the TypoScript or TSconfig itself.

## Documentation

- TSconfig Coding Guideline: https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/CodingGuidelines/CglTsConfig.html

*Note:*  The term `TSconfig` differs threw the documentation in TYPO3, so the name of the folder `TSconfig` does not change in this Pull-Request, since most core extensions use the same writing.

# How to test

1. Open the TypoScript object browser inside of the backend
2. You should see the constants and setup, even if you include the old files.

Since there are not Changes to the TypoScript itself only the includes needs to be testet.

Fixes: #2813


3l73:
This backport contains one merge conflict:
1. File `ext_localconf.php` missing the changes from the current master. The original structure is in use, only the include of the PageTS was modified.

Issue #2576, opened 2020-04-21
[BUG] Environment variables not handled correctly in site management


dkd-friedrich:
**Describe the bug**
If environment variables are used to set EXT:solr specific settings in your site configuration, this values are not correctly displayed and handled in the site management.

**To Reproduce**

1. Define an environment variable, e.g. `SOLR_SCHEME_READ`
2. Set environment variable in your site configuration (`<project-root>/config/sites/<identifier>/config.yaml`), e.g. `solr_scheme_read: '%env(SOLR_SCHEME_READ)%'`
3. Open tab "Solr" in your site configuration
4. Environment variables are not correctly handled/displayed, in this example "Scheme" shows `[ INVALID VALUE ("%env(SOLR_SCHEME_READ)%") ] [%env(SOLR_SCHEME_READ)%]`. Due to the validation, e.g. for the port setting, saving the whole site configuration is blocked.

**Expected behavior**
Values from environment variables should be displayed, but not be editable. 

**Screenshots**

@mschwemmer already provided some explaining screenshots in #2538, showing the set environment variables and the form in the site management, thx.

![79352920-b7667f80-7f3a-11ea-9ca6-75df303d1232](https://user-images.githubusercontent.com/9529685/79829748-35e07880-83a4-11ea-849c-73ee7fa4b19a.png)

![79353155-f7c5fd80-7f3a-11ea-83df-0f1ccdccc64f](https://user-images.githubusercontent.com/9529685/79829753-38db6900-83a4-11ea-96c0-ab7ef7178137.png)

**Used versions:**
 - TYPO3 Version: 9.5.0+
 - EXT:solr Version: 10+

PR #2817, opened 2020-12-22
[BUGFIX] ENV vars not handled correctly in site management module

dkd-kaehm:
* Use input and value picker for solr_scheme_read/write instead of select,
   to be able to insert/see the ENV variable.

Fixes: #2576

This Pull-Request allows to put/see the ENV variable in site configuration via Sites module:

TYPO3-Solr_sites_module


dkd-friedrich (approved):
LGTM

pduersteler:
Is there something one can help with to get this through? It would be nice to have this as we're currently work-arounding for the `solr_scheme_read` configuration.

dkd-kaehm:
> Is there something one can help with to get this through? It would be nice to have this as we're currently work-arounding for the `solr_scheme_read` configuration.

We must rebase and get backports onto 11 and 10 branches.

PR #2820, opened 2020-12-22
[BUGFIX:BACKPORT:10] garbage collector on translations #2797

dkd-kaehm:
should not remove IndexQueue Item
should not remove Pages of other languages

Resolves: #2797


3l73 (approved):
Looks good to me.

Issue #2824, opened 2021-01-08
[BUG] Single / in search term leads to an exception


3l73:
**Describe the bug**
Searching for a term containing a single Slash, e.g. `Users/` leads to a TYPO3 core exception.

The exception:
```
PHP Warning: preg_match_all(): Compilation failed: missing closing parenthesis at offset 25 in EXT:solr/Classes/Domain/Search/Score/ScoreCalculationService.php line 126
```

**To Reproduce**

To reproduce use the project https://github.com/TYPO3-Solr/solr-ddev-site

1. Open the backend
2. Select page `EXT:Solr test cases`
3. Add a text element with content `/Users/user1/Pictures/picture1.png`
4. Save page and index
5. Open frontend and search for `Users/`

**Expected behavior**
Page `EXT:Solr test cases` should be listed within the search results

**Screenshots**

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.10
 - Browser: Chromium
 - EXT:solr master (Hash c099f5b73cf8b6189ae2ac75bb5e11820da40c1a)
 - Used Apache Solr Version: 8.5.2
 - PHP Version: 7.4
 - MySQL Version: 10.2

PR #2825, opened 2021-01-08
[BUGFIX] Quote field within score calculation

3l73:
# What this pr does

Quote the value of variable $field for regular expressions.
The quote avoid exceptions in case the value contains one single slash.

# How to test

To reproduce use the project https://github.com/TYPO3-Solr/solr-ddev-site

1. Open the backend
2. Select page `EXT:Solr test cases`
3. Add a text element with content `/Users/user1/Pictures/picture1.png`
4. Save page and index
5. Open frontend and search for `Users/`

**Expected behavior**
Page `EXT:Solr test cases` should be listed within the search results

Fixes: #2824


dkd-kaehm (approved):
:+1:

PR #2833, opened 2021-01-14
[TASK] Disable cache time information for ajax request

georgringer:
If `$GLOBALS['TYPO3_CONF_VARS']['FE']['debug']` is enabled (e.g. on Staging or local environments), the TypoScriptFrontendController will add some information to the response 
```html
<!-- Cached page generated 14-01-21 08:19. Expires 15-01-21 08:19 -->
```

to avoid this, the setting `config.debug=0` can be used which turns this off, no matter which global debug config is set

Resolves: #2834


dkd-kaehm (approved):
:+1:

dkd-kaehm:
@georgringer Thanks for fixing that.
I assume we need it in 10 and 11 branches back ported.

georgringer:
@dkd-kaehm yes this can be easily backported

PR #2836, opened 2021-01-18
[BUGFIX] Check if $recordUid is non-numeric before substitution

smichaelsen:
It is assumed here that if the $status is 'new', then the $recordUid must be a 'NEW...' string, which is normally the case.
However the $id is passed by reference and might be changed by other extensions. gridelements for example changes the $id.
https://gitlab.com/coderscare/gridelements/-/issues/156
So the check must be made more robust here.


dkd-kaehm:
@smichaelsen Thanks for the pull requests.
We want a testcase for this behavior.
We must fake a "third party"-Hook changing the $id reference value like on gridelements.
May be we want to make comment on method to prevent the type hinting for those methods, to prevent fatals by type mismatch.

3l73 (approved):
Looks good to me

3l73 (approved):
Looks good to me.

PR #2857, opened 2021-02-12
[BUGFIX:BP:10] Check if $recordUid is non-numeric before substitution

dkd-kaehm:
It is assumed here that if the $status is 'new', then the $recordUid must be a 'NEW...' string, which is normally the case.
However the $id is passed by reference and might be changed by other extensions. gridelements for example changes the $id.
https://gitlab.com/coderscare/gridelements/-/issues/156
So the check must be made more robust here.

Fixes: #2836


3l73 (approved):
Looks good to me.

Issue #2830, opened 2021-01-13
[BUG] Cached Translate Viewhelper (+Solution +Explanation)


dev-rke:
**Describe the bug**
When a search result page is cached, the translate viewhelper throws an exception in vsprintf().

**To Reproduce**
1. Create a page with plugin, set up a standard solr + standard configuration
2. Ensure page can be cached and fluid will be cached
3. Search for a non-existant search word in your search index, twice (!), so fluid is cached on second run + cache hash is generated
4. Exception is thrown:

```
#1476107295 TYPO3\CMS\Core\Error\Exception

PHP Warning: vsprintf(): Too few arguments in /var/www/web/typo3conf/ext/solr/Classes/ViewHelpers/TranslateViewHelper.php line 61

at TYPO3\CMS\Core\Error\ErrorHandler->handleError(2, 'vsprintf(): Too few arguments', '/var/www/web/typo3conf/ext/solr/Classes/ViewHelpers/TranslateViewHelper.php', 61, array('id' => 'results_searched_for', 'extensionName' => 'Solr', 'arguments' => array('wtf'), 'languageKey' => null, 'alternativeLanguageKeys' => null, 'result' => 'Searched for %s.'))
```

**Expected behavior**
No exception is thrown.

**Solution and Explanation**
Change line 61 and remove the arguments key from the $arguments array, as it is resolved in cached state already.
```diff
-61 $result = vsprintf($result, $arguments['arguments']);
+61 $result = vsprintf($result, $arguments);
```

Furthermore here comes some special PHP behaviour into context.
When the first vsprint argument contains a replacement marker, the second argument is evaluated.
If the first argument contains no replacement marker, the second argument is ignored and won't raise an error.
So no error will be thrown, when `$result` does not contain any replacement marker.
For a better understanding, please see the following simplified example on an interactive PHP shell (php -a):
```
php > echo vsprintf('Hello %s', $uninitialized);

Warning: vsprintf(): Too few arguments in php shell code on line 1

Call Stack:
  261.1194     394080   1. {main}() php shell code:0
  261.1194     394080   2. vsprintf($format = 'Hello %s', $args = NULL) php shell code:1

php > echo vsprintf('Hello world', $uninitialized);
Hello world
```

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.1, i remember this bug is years old, but is annoying in the logs
 - Browser: Firefox, Chrome
 - EXT:solr Version: 10

PR #2859, opened 2021-02-12
[BUGFIX:BP:10] Exception on Cached state of TranslateViewHelper

dkd-kaehm:
fix arguments for vsprintf in TranslateViewHelper when it's cached

Fixes: #2830


3l73 (approved):
LGTM

Issue #2847, opened 2021-01-28
Solr indexes workspace records in TYPO3 10.4


bfzgaier:
Before workspace records were place on pid -1. With 10.4 this behaviour changed.

Now workspace records are placed on the corresponding page. Solr still uses `pid != 1` to filter workspace records

https://github.com/TYPO3-Solr/ext-solr/blob/b6ff0f2314bccf20658688baa04ea3c33e7a5ba4/Classes/IndexQueue/Initializer/AbstractInitializer.php#L304

On possible solution is, to instead check for the workspace id of a record.
```php
$conditions['versioningWS'] = 't3ver_wsid = 0'; 
```
The same approach is used in `TYPO3\CMS\Core\Database\Query\Restriction\WorkspaceRestriction`

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Important-89555-Workspace-relatedDatabaseRecordsContainTheProperPageID.html

dkd-kaehm:
@bfzgaier 
Thanks for reporting. We'll fix it in release-11.0.x and master.

PR #2864, opened 2021-02-19
[BUGFIX] Change filter for workspace

3l73:
# What this pr does

Since TYPO3 10.4 the workspace now is not identify by pid value -1,
instead field 't3ver_wsid' is in use to filter workspace records.

This bugfix changes the filter to this new field and takes care about TYPO3 9.5 compatibility.

# How to test

This test uses project https://github.com/TYPO3-Solr/solr-ddev-site

1. Install extension workspaces: `ddev composer require typo3/cms-workspaces:^10.4`
2. Activate extension: `ddev exec vendor/bin/typo3cms extension:activate workspaces`
3. Create a new workspace `Solr test` within the backend
  - Note: When save an error occured, but the workspace is created
4. Switch to workspace `Solr test`
5. Add a new site `Workspace-site`
6. Activate the page
  - Note: When save an error occured, but the page is activated
7. Go to the Solr Index Queue and initialise the Queue

You should not find the page within the index queue.

Fixes: #2847

### DIFF
--- a/Classes/IndexQueue/Initializer/AbstractInitializer.php
+++ b/Classes/IndexQueue/Initializer/AbstractInitializer.php
@@ -301,7 +301,8 @@ abstract class AbstractInitializer implements IndexQueueInitializer
 
         if (!empty($GLOBALS['TCA'][$this->type]['ctrl']['versioningWS'])) {
             // versioning is enabled for this table: exclude draft workspace records
-            $conditions['versioningWS'] = 'pid != -1';
+            /* @see \TYPO3\CMS\Core\Database\Query\Restriction\WorkspaceRestriction::buildExpression */
+            $conditions['versioningWS'] = 't3ver_wsid = 0';
         }
 
         if (count($conditions)) {


Classes/IndexQueue/Initializer/AbstractInitializer.php
```diff
@@ -30,6 +30,7 @@
 use ApacheSolrForTypo3\Solr\Domain\Index\Queue\QueueItemRepository;
 use ApacheSolrForTypo3\Solr\Domain\Site\Site;
 use ApacheSolrForTypo3\Solr\System\Logging\SolrLogManager;
+use ApacheSolrForTypo3\Solr\Util;
```
dkd-kaehm, line 33: For master we could remove the usage.
We need it in backport against release-11.0.x branch.

3l73: The use statement was removed.

dkd-kaehm (approved):

Issue #2865, opened 2021-02-19
[BUG] Variants with custom variant field not expanded


dkd-friedrich:
**Describe the bug**

Currently the `VariantsProcessor` expects the contents of the variant field to be in lowercase and explicitly applies a `mb_strtolower`. As Solr returns the contents in the `expanded` section unmodified, the variants may not be detected correctly and can not be accessed. This bug doesn't occur with the default variantId field, since the contents of this field are always entirely in lowercase.

**To Reproduce**

Activate the variants and use an custom stringS field

```
plugin.tx_solr.search.variants = 1
plugin.tx_solr.search.variants.variantField = p214_stringS
```

Of course, a field must be selected that provides matching Solr documents to test the variants. I.e. there must be documents with different but also identical contents.

**Expected behavior**

 `variants` in `ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\Result\SearchResult` should be set correctly
 
 **Used versions:**
 - TYPO3 Version: 9.5
 - EXT:solr Version: 10-11
 - Used Apache Solr Version: 8.5.1

PR #2869, opened 2021-02-22
[BUGFIX:BACKPORT:10] Fix handling of case sensitive variant ids

dkd-friedrich:
# What this pr does

This PR backports the fix of the handling of case sensitive variant ids for
EXT:solr 10.x

The VariantsProcessor expects the content of the variant field to be
in lowercase and explicitly applies an mb_strtolower. Since Solr returns
the contents in the expanded field unchanged and the contents might be
case sensitive, the variants may not be recognized correctly and cannot
be accessed.

This issue is fixed by removing the unnecessary mb_strtolower.

# How to test

Configure a custom variant field with case sensitive contents and check if the 
variants are detected correctly and are accessible.

Fixes: #2865


dkd-kaehm (approved):
:+1:

Issue #2870, opened 2021-02-23
[FEATURE] Store number of existing variants


dkd-friedrich:
**Is your feature request related to a problem? Please describe.**
Since the number of returned variants is limited by `variants.limit`, it's currently not possible to determine how many variants exists.

**Describe the solution you'd like**
Solr returns the number of variants in `expanded->#variantId#->numFound`, we should store this information in the SearchResult

**Target versions**
EXT:solr 9+

PR #2871, opened 2021-02-23
[FEATURE] Store number of existing variants

dkd-friedrich:
# What this pr does

To allow to access the number of existing variants, the SearchResult
is extended. Without this modification the number of existing variants
can is unknown, as the number of returned variants may be limited by
variants.limit.

# How to test

Configure variants and check the set `variantsNumFound` in SearchResult

Resolves: #2870

### DIFF
--- a/Classes/Domain/Search/ResultSet/Result/SearchResult.php
+++ b/Classes/Domain/Search/ResultSet/Result/SearchResult.php
@@ -35,6 +35,25 @@ use ApacheSolrForTypo3\Solr\System\Solr\Document\Document;
  */
 class SearchResult extends Document
 {
+    /**
+     * The variant field value
+     *
+     * Value of Solr collapse field, which is defined via
+     * TypoScript variable "variants.variantField"
+     *
+     * @var string
+     */
+    protected $variantFieldValue = '';
+
+    /**
+     * Number of variants found
+     *
+     * May differ from documents in variants as
+     * returned variants are limited by expand.rows
+     *
+     * @var int
+     */
+    protected $variantsNumFound = 0;
 
     /**
      * @var SearchResult[]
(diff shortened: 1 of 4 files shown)


Documentation/Configuration/Reference/TxSolrSearch.rst
```diff
@@ -841,7 +841,7 @@ The following example shows how to keep all options of all facets by keeping the
 plugin.tx_solr.search.faceting.keepAllFacetsOnSelection = 1
 plugin.tx_solr.search.faceting.countAllFacetsForSelection = 1
 plugin.tx_solr.search.faceting.minimumCount = 0
-```
+```````````````````````````````````````````````
```
dkd-kaehm, line 844: Not sure we want this change
Thats the code-block with three-backtick wrappers.

Classes/Domain/Search/ResultSet/Result/SearchResult.php
```diff
@@ -35,6 +35,24 @@
  */
 class SearchResult extends Document
 {
+    /**
+     * The variant field value
+     *
+     * Value of "variants.variantField"
```
dkd-kaehm, line 41:
```suggestion
     * Value of "variants.variantField" of Solrs response.
```

We have the same TypoScript setting https://docs.typo3.org/p/apache-solr-for-typo3/solr/master/en-us/Configuration/Reference/TxSolrSearch.html#variants-variantfield. 
Would propose at least a proper comment, to avoid double Dutch on the contexts.
Better is the speaking name.

dkd-kaehm (commented):
I'll test/analyse the stuff together with EXT:solrfluidgrouping addon. 
This addon should use the same technology... 
Hopefully that does not break anything.

Classes/Domain/Search/ResultSet/Result/SearchResult.php
dkd-friedrich: This comment refers to the TS setting, not to the Solr response.

`variantFieldValue ` will be filled with the content of the field configured in TS setting `variants.variantField`

Documentation/Configuration/Reference/TxSolrSearch.rst
dkd-friedrich: I think we can/should skip this, the IDE did that automatically.

Classes/Domain/Search/ResultSet/Result/SearchResult.php
dkd-friedrich: ok, I see, this comment mistakable, I will change this to:

`Value of Solr collapse field, which is defined via TypoScript variable "variants.variantField"`

Documentation/Configuration/Reference/TxSolrSearch.rst
dkd-friedrich: Done

Classes/Domain/Search/ResultSet/Result/SearchResult.php
dkd-friedrich: Done

dkd-kaehm (approved):
:+1:

Issue #2878, opened 2021-03-01
[BUG] Variant field value not set if no variants found


dkd-friedrich:
**Describe the bug**
If variants are enabled and a returned Solr document has no variants, the variant field value is not set. Currently `variantFieldValue` is only set if `variantId ` is set in `expanded`, but if there is only one variant in the Solr index there will be no data in the `expanded ` array.

**To Reproduce**
1. Activate variants 
2. Set the variant field to simulate
2.1 one document with at least one variant
2.1 one document without further variants
3. Check the variant field value of the document without variants

**Expected behavior**
Variant field value should be filled even if a Solr document has no further variants 

**Used versions (please complete the following information):**
 - TYPO3 Version: 9.5 - 10.4
 - EXT:solr Version: release-10, release-11, master

PR #2881, opened 2021-03-01
[BUGFIX:BACKPORT:10] Fix missing variant field value

dkd-friedrich:
If variants are enabled and a returned Solr document has no variants, the variant field value is not set. Currently variantFieldValue is only set if variantId is set in expanded, but if there is only one variant in the Solr index there will be no data in the expanded array.

This comit fixes this issue by ensuring that the variant field value is set even if there is no data in the expanded array.

Backport for EXT:solr 10.x

# How to test

1. Activate variants
2. Set the variant field to simulate
2.1. one document with at least one variant
2.2. one document without further variants
3. Check the variant field value of the document without variants, it's now set

Fixes: #2878

### DIFF
--- a/Classes/Domain/Variants/VariantsProcessor.php
+++ b/Classes/Domain/Variants/VariantsProcessor.php
@@ -91,13 +91,13 @@ class VariantsProcessor implements SearchResultSetProcessor
                 continue;
             }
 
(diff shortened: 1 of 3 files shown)


dkd-kaehm (approved):
:+1:

Issue #1922, opened 2018-03-30
Make setting focus on search input with autosuggest optional


Mabahe:
In order to be able to not having the focus on the input field when suggest is activated introduce an extra css class.

PR #2895, opened 2021-03-15
Remove initial focus() for accessibility

mck:
The Inital focus() should be removed since the user should always start at the beginning of the page.
In the current state the user just gets pulled in the search field with no idea where he is.

# What this pr does

Remove inital focus() of search input field

# How to test

On Page load the focus should not jump into the search input field.


dkd-kaehm:
Please see how to make focus optional: #1922 and #1923

PR #2896, opened 2021-03-15
Update de.locallang.xlf

mck:
Submit is an action so the right wording would be "suchen"

# What this pr does

Change the translation for the "submit" action label

### DIFF
--- a/Resources/Private/Language/de.locallang.xlf
+++ b/Resources/Private/Language/de.locallang.xlf
@@ -85,7 +85,7 @@
 			</trans-unit>
 			<trans-unit id="submit" xml:space="preserve" approved="yes">
 				<source>Search</source>
-				<target>Suche</target>
+				<target>suchen</target>
 			</trans-unit>
 			<trans-unit id="sorting_sortBy" xml:space="preserve" approved="yes">
 				<source>Sort by</source>


dkd-kaehm (commented):
Thanks for your contributions.
I can't currently confirm the change and want at least the confirmation from somebody else.

You can currently use the TYPO3s way to override the translation labels you want to change.
Please refer to the docs or see following example: 

https://github.com/TYPO3-Solr/ext-solrfluidexample/blob/master/ext_localconf.php#L6

dkd-kaehm:
@dkd-friedrich 
Please confirm or close the change.
By confirmation we need Backports onto release-10.0.x and release-11.0.x.
Please consider the compatibility to #2875

dkd-friedrich:
Sounds reasonable to me, but I would suggest to start with capital letter 
`suchen` => `Suchen`

Our other labels for buttons and links always start with a capital letter and we should keep it that way here as well

Resources/Private/Language/de.locallang.xlf
```diff
@@ -85,7 +85,7 @@
 			</trans-unit>
 			<trans-unit id="submit" xml:space="preserve" approved="yes">
 				<source>Search</source>
-				<target>Suche</target>
+				<target>suchen</target>
```
dkd-friedrich, line 88: I would suggest to start with a capital letter, similiar to the other labels

dkd-friedrich (changes requested):

dkd-kaehm:
Is part of #2875

PR #2917, opened 2021-04-23
[BUGFIX] remove escaping on suggestion prefix

dkd-kaehm:
Since solarium escapes the special uri characters with RFC1738,
there are no needs to do the same within the EXT:Solr.

see:
https://github.com/solariumphp/solarium/blob/97a9df0c2d9ba35db3c747b30c0809702296f18b/src/Component/RequestBuilder/RequestParamsTrait.php#L153

This is the first step of removing the escaping/encoding from EXT:Solr API.
Other components of EXT:solr will be fixed as well in different pull requests.

### DIFF
--- a/Classes/Controller/SuggestController.php
+++ b/Classes/Controller/SuggestController.php
@@ -2,28 +2,18 @@
 
 namespace ApacheSolrForTypo3\Solr\Controller;
 
-/***************************************************************
- *  Copyright notice
+/*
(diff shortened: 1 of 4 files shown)


dkd-kaehm:
BAckports aginst 10 and 11 must be done after this PR is approved.

3l73 (approved):
LGTM

PR #2921, opened 2021-04-28
Release 8.1.x

cwplus:
this is not a bug , but a question,

I searched the docs, but couldn't find a way or an example to add a new parser.

In the class \ ApacheSolrForTypo3 \ Solr \ Domain \ Search \ ResultSet \ Result \ Parser \ ResultParserRegistry we find a function registerParser (), but we can't find a way to call this class and the function outside of solr.

I am thinking, if I can extend this class (with XClass), to add my custom parser 

Can you help me plz ?


3l73:
@cwplus I guess you don't want to offer a pull request to your question :-)

May be you can create a new Issue with your question?

* https://github.com/TYPO3-Solr/ext-solr/issues/new?assignees=&labels=&template=task.md&title=%5BTASK%5D 

Meanwhile I'll check it to give you a suitable answer.

PR #2898, opened 2021-03-16
[BUGFIX] use pages configuration as default #issue-2742

achimfritz:
Site->getPages() should use page indexConfigurationName if no configuration is found for rootPage

Resolves: #2742

# What this pr does

use "pages" indexConfigurationName even if no Configuration found for rootPage

# How to test

s. Integration Test

Fixes: #2742

PR #2925, opened 2021-05-07
 [BUGFIX] Record indexing doesn't work anymore if page queue is disabled

dkd-kaehm:
The initializer MUST ignore only those pages, that match the `additionalWhereClause` directly, and NOT the matched page incl. whole sub-tree of that page.
The Record-Monitoring stack ignores the state of parents-tree,
which match the `additionalWhereClause` and adds the pages to the index queue after first change anyway.
This change fixes this issue and makes it possible to index the pages and records
stored inside of inaccessible sub-tree.

Fixes: #2241
Closes: #2898
Replaces: #2949


Classes/Domain/Site/Site.php
```diff
@@ -150,33 +142,24 @@ public function getDefaultLanguage()
...
-     * all page types! Deleted pages are not included.
-     *
-     * @param int|string $rootPageId Page ID from where to start collection sub pages
-     * @param int $maxDepth Maximum depth to descend into the site tree
-     * @return array Array of pages (IDs) in this site
+     * @inheritDoc
      */
-    public function getPages($rootPageId = 'SITE_ROOT', $maxDepth = 999)
+    public function getPages(
+        ?int $pageId = null,
```
dkd-friedrich, line 148: I think this change could be breaking, as getPages() is a public method and so far accepted int and string values. 

String values, especially the keyword SITE_ROOT could be used in third-party extensions and would stop working with these changes, what do you think?

dkd-kaehm: That PR is for master aka 11.1.x, so the BC should be fine for that.
The origianal PR #2898 should go onto release-11.0.x instead of that PR.

dkd-friedrich: Ok, fine for me as there are breaking changes anyway

But we should mention this in the release notes, please as add discused

PR #2939, opened 2021-05-26
[BUGFIX] Delete documents for valid connections only

christophlehmann:
This prevents the Exception `Call to a member function getWriteService() on null` when solr is disabled (solr_enabled_read: false)

# What this pr does

It prevents an exception when solr is disabled (solr_enabled_read: false)

# How to test

Set `solr_enabled_read: false` and save a translated page

Fixes: #2940


dkd-kaehm (approved):
@christophlehmann 
Please make Backport into releas-11.0.x

PR #2941, opened 2021-05-26
[BUGFIX:BP:11] Delete documents for valid connections only

christophlehmann:
Relates: #2939


dkd-kaehm (approved):
@christophlehmann 
Thanks for fixing that issue,
This comes in 11.0.5 release, which will be released shortly.

PR #2946, opened 2021-05-28
[TASK] Bump Chart.js to v2.9.4

dkd-kaehm:
Update Chart.js to newer v2.9.4, see https://github.com/advisories/GHSA-h68q-55jf-x68w for more details.


dkd-kaehm:
Works with v2.9.4.
No needs for vaiting on a build status.
The Ports in relese-10.0.x and relese-11.0.x will be done directly in branches and without PRs.

Issue #2725, opened 2020-09-14
[TASK] Migrate cHash handling


DanielSiepmann:
cHash handling was disabled with some code as it wasn't compatible, see #2211.
The mentioned "issue" or missing feature inside of TYPO3 is fixed since v10 LTS: https://review.typo3.org/c/Packages/TYPO3.CMS/+/63062.

Therefore this workaround should be removed, and a proper example configuration should be provided instead.

Right now we apply the following patch to EXT:solr:

```patch
diff --git a/Classes/Domain/Search/Uri/SearchUriBuilder.php b/Classes/Domain/Search/Uri/SearchUriBuilder.php
index 81eb4321..7fc5246a 100644
--- a/Classes/Domain/Search/Uri/SearchUriBuilder.php
+++ b/Classes/Domain/Search/Uri/SearchUriBuilder.php
@@ -310,36 +310,22 @@ class SearchUriBuilder
      */
     protected function buildLinkWithInMemoryCache($pageUid, array $arguments)
     {
-        $values = [];
-        $structure = $arguments;
-        $this->getSubstitution($structure, $values);
-        $hash = md5($pageUid . json_encode($structure));
+        $hash = md5($pageUid . json_encode($arguments));
         if (isset(self::$preCompiledLinks[$hash])) {
             self::$hitCount++;
             $uriCacheTemplate = self::$preCompiledLinks[$hash];
         } else {
             self::$missCount++;
-            $this->uriBuilder->setTargetPageUid($pageUid);
-            $uriCacheTemplate = $this->uriBuilder->setArguments($structure)->setUseCacheHash(false)->build();
-
-            // even if we call build with disabled cHash in TYPO3 9 a cHash will be generated when site management is active
-            // to prevent wrong cHashes we remove the cHash here from the cached uri template.
-            // @todo: This can be removed when https://forge.typo3.org/issues/87120 is resolved and we can ship a proper configuration
-            $urlHelper = GeneralUtility::makeInstance(UrlHelper::class, $uriCacheTemplate);
-            $urlHelper->removeQueryParameter('cHash');
-            $uriCacheTemplate = $urlHelper->getUrl();
+            $uriCacheTemplate = $this->uriBuilder
+                ->setTargetPageUid($pageUid)
+                ->setArguments($arguments)
+                ->setUseCacheHash(true)
+                ->build();
 
             self::$preCompiledLinks[$hash] = $uriCacheTemplate;
         }
 
[...]

PR #2972, opened 2021-06-28
[TASK] Adapt cHash handling for TYPO3 10

dkd-friedrich:
# What this pr does

cHash handling was adapted/disabled to fix generated links in TYPO3 9,
as this adaptions lead to invalid links, e.g. facet links, in TYPO3 10
and Site configuration support is available since EXT:solr 10 there is
no need for this adaptions.

This commit removes the implemented removal of the cHash parameter,
now the UriBuilder is used to build the link template.

# How to test

Index some contents and check the generated links in the frontend:
- facet
- sorting
- pagination

Resolves: #2725

### DIFF
--- a/Classes/Domain/Search/Uri/SearchUriBuilder.php
+++ b/Classes/Domain/Search/Uri/SearchUriBuilder.php
@@ -75,7 +75,7 @@ class SearchUriBuilder
      * @param $facetValue
      * @return string
      */
-    public function getAddFacetValueUri(SearchRequest $previousSearchRequest, $facetName, $facetValue)
+    public function getAddFacetValueUri(SearchRequest $previousSearchRequest, $facetName, $facetValue): string
     {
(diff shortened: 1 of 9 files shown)


dkd-kaehm (changes requested):
We have a troubles with range facets, since the links are builded by JS/Browser and we can't
build all the links for all possible ranges-combinations.
I would propose to draw-up the tactics for dynamic exclusion of tx_solr params plugin-namespace...

dkd-friedrich:
As already mentioned in the related issue, rafael and I have taken a closer look at the problem and have now chosen a slightly different path.

Since no cHash calculation is possible for the links of the range facets generated via JavaScript and a more extensive caching was possible in the old variant, we decided to exclude the tx_solr parameters. Configuration options for namespaces and parameters is added and a default cHash configuration provided.

Issue #3009, opened 2021-08-31
[BUG] Fatal error when installing 11.1.1 [Non-COMPOSER]


jdinse:
When installing the solr extension 11.1.1 within the Extension Manageer, I got a fatal error:

`Fatal error: Declaration of TYPO3\SymfonyPsrEventDispatcherAdapter\EventDispatcherAdapter::dispatch(object $event, ?string $eventName = NULL): object must be compatible with Symfony\Contracts\EventDispatcher\EventDispatcherInterface::dispatch($event) in /Users/jdinse/www/typo3_src-10.4.20/vendor/typo3/symfony-psr-event-dispatcher-adapter/src/EventDispatcherAdapter.php on line 20`

Neither the frontend nor the backend or the install tool is reachable anymore.

with 11.0.5, the installation is fine.

**Used versions:**
 - TYPO3 Version: 10.4.20
 - Browser: Chrome  92.0.4515.159
 - EXT:solr Version: 11.1.1
 - PHP Version: 7.4.0

dkd-kaehm:
@jdinse 
Thanks for reporting.
Please provide the full stack trace of this error.

dkd-kaehm:
@jdinse 
Currently installed on non_composer with introduction package.
I can't reproduce this issue. 

---

Will be closed after 28 days of silence.

dkd-kaehm:
@jdinse 
Please join TYPO3s Slack ext-solr channel to get help with that trouble.
I'll close that issue.

rafu1987:
I have the same issue, can't use the extension for now unfortunately.

dkd-kaehm:
@jdinse @rafu1987 
Please try attached EXT:solr 11.1.2-RC0
That is the state of pull-requst #3029 
[solr_11.1.2.zip](https://github.com/TYPO3-Solr/ext-solr/files/7145714/solr_11.1.2.zip)

ulrike-cosmoblonde:
I tested version 11.1.2-RC0 and can confirm that the problem described in #3034 is solved with that version. Thank you!

PR #3029, opened 2021-09-10
[BUGFIX] Fatal error when installing 11.1.1 [Non-COMPOSER]

dkd-kaehm:
This change removes the libraries, which are required by solarium and already provided by TYPO3 typo3_src-X.Y.Z.
It prevents the TYPO3 installations overloading/installing to incompatible versions of libraries required by TYPO3 core.

Fixes: #3009


dkd-kaehm:
@bmack 
Is it the the way to go?

dkd-kaehm (commented):
Seems to work now.

ulrike-cosmoblonde:
I tested version 11.1.2-RC0 and can confirm that the problem described in #3034 is solved with that version. Thank you!

Issue #3031, opened 2021-09-22
[TASK] Ensure valid cache ids


dkd-friedrich:
**What should be done in the scope of this task?**

Solr extensions are hooking into the DataHandler observing record changes and 
partly these hooks are using the TwoLevelCache and the DataHandler record ids 
to cache already processed elements.

As the NEW ids processed by the DataHandler can be set by extensions and may be invalid
for the usage with the cache frontend, we should sanitize the ids to avoid exceptions.

PR #3032, opened 2021-09-22
[TASK] Ensure valid cache ids

dkd-friedrich:
# What this pr does

Ensures that the cache id used in the TwoLevelCache is compatible with the  FrontendInterface::PATTERN_ENTRYIDENTIFIER.

# How to test

Run the integrations test to be sure the TwoLevelCache still works like expected

Resolves: #3031


dkd-kaehm (approved):
:+1:

PR #3038, opened 2021-09-27
[BUGFIX] Respect TCA setting of 'tstamp' field

jschlier:
# What this pr does
When updating a tracked, localisable item in the Solr queue the TCA setting for the "tstamp" / last updated on field ist not evaluated.
This leads to an exception with records that use a non-default name for this database field.

This PR changes this behaviour to use the TCA settings for the `$itemType`.

# How to test
Add (or change) a custom record that is configured for Solr to use a non-default name for this field. E.g:
```php
$GLOBALS['TCA'][$table]['ctrl']['tstamp'] = 'updated_on';
```
Then you can test this exception and the fix before and after applying this small change.

Fixes: TYPO3-Solr#3037

### DIFF
--- a/Classes/Domain/Index/Queue/QueueItemRepository.php
+++ b/Classes/Domain/Index/Queue/QueueItemRepository.php
@@ -289,7 +289,7 @@ class QueueItemRepository extends AbstractRepository
             $queryBuilder = GeneralUtility::makeInstance(ConnectionPool::class)->getQueryBuilderForTable($itemType);
             $queryBuilder->getRestrictions()->removeAll();
             $localizedChangedTime = $queryBuilder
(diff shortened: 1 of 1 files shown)


dkd-friedrich (changes requested):
Thanks for the adjustment, sounds reasonable to me in general.

In the PR, however, 2 different variables were used by mistake, `$timeStampField` and `$tStampField`. Please check.

jschlier:
Oh. My bad - I quickly fixed this typo.

dkd-kaehm:
We want the ports into release-11.1.x and release-11.0.x

dkd-friedrich (approved):
Thx! LGTM

dkd-friedrich:
> We want the ports into release-11.1.x and release-11.0.x

I've prepared the backports

- #3039
- #3040

PR #3068, opened 2021-10-19
[BUGFIX] changed from always picking the 0 array value to pick always…

Fanor51:
Hello, 

we got the issues that we have TYPO3 pages were the default language ist disabled in the site config. With this settings we cant use the TYPO3 BE Module "Core Optimization" because there always was showing up an exception (Call to a member function getAdminService() on null). 

With this little fix we prevent the hard access on the default language solr settings (uid 0) and use instead the first in the connection array. 

In our example "$this->selectedSolrCoreConnection" got only the key [8]=> ...  (language uid) and we don´t have 0.

What is your opinion about that?

Best Regard
Fabio

### DIFF
--- a/Classes/Controller/Backend/Search/AbstractModuleController.php
+++ b/Classes/Controller/Backend/Search/AbstractModuleController.php
@@ -384,7 +384,7 @@ abstract class AbstractModuleController extends ActionController
         if (empty($solrCoreConnections)) {
             return;
         }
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
Thanks for Fixing that Issue.
We want back ports into 
* release-11.5.x
* release-11.2.x 
* ~~release-11.1.x~~
* ~~release-11.0.x~~

PR #3070, opened 2021-10-22
[BUGFIX] Hierarchy facet items containing dashes cannot be unset

tmaroschik:
If you try to apply the hierarchy facet to a path of UUIDs instead of integers the facet can be selected, but not unselected anymore. This happens because the HierarchyFacetParser assumes that every value containing a dash is one that already contains the depth param. Which is not the case when dealing with uuids like `/228d03bf-de44-4fcd-91fd-6ab05175fefc/901d43ca-28cf-4dbd-983d-ac96a0930530/a3f94d07-30fe-4ad1-a0cf-b5082db3aa94/`.

This fix introduces a check that matches the value against the depth param pattern.


simouel:
I would like to see this patch merged as well
:)

dkd-kaehm:
@tmaroschik, @simouel  
We want the unit or integration test case for that.
The description with steps to reproduce is missing.
Could you add a test case?

Issue #2235, opened 2019-02-01
Garbage collector ignores records translations


dogawaf:
Hi

Scenario:
- a record is created in default language (gb)
- it is now in the index queue, and after indexing, in solr's core_gb
- the record is localized in the other language (fr)
- it is updated in the index queue, and after indexing, in solr's core_gb and core_fr

All is fine.

Now : 
- I hide the localized record
- the garbage collector does not remove the record from core_fr
- I hide the record in default language
- the garbage collector removes the record from both core_gb and core_fr

IMHO, the garbage collector should manage properly operations on overlayed records.

internezzo-prod:
It seems the issue is that in  https://github.com/TYPO3-Solr/ext-solr/blob/master/Classes/GarbageCollector.php#L251 collectGarbage is called using the localized of the record.
But since the records are indexed using the uid of the default-language-record, nothing is removed...

PR #3078, opened 2021-10-24
[BUGFIX] Remove hidden translated record in index

christophlehmann:
# What this pr does

Takes care about removing translated record in index. It also remove the superflous `@return` since it's not used anywhere

# How to test

Hide a translation of a record. The translation is not in index anymore

Fixes: #2235


dkd-kaehm:
@christophlehmann, thanks. 
Closed, because already merged in #3149 and it is a part of main:
https://github.com/TYPO3-Solr/ext-solr/commit/1b76421159bd8c666ec6f1242be41ef11d09ee5b

Issue #3094, opened 2021-11-05
Installation script failed because of missing mirror


premar:
**Describe the bug**
Install script cant be executed because mirror is missing.

**To Reproduce**
./ext-solr/Resources/Private/Install/install-solr.sh

```
Starting installation of Apache Solr with the following settings:
Install Mode: Development                                    
Solr Version: 8.9.0                                    
Installation Path: /opt/solr                           
Port: 8983                                               
Checking requirements.
ERROR couldn't ping Apache download mirror, try again using wget
ERROR Also couldn't reach the Apache download mirror using wget. Please check your internet connection.
Please install all missing requirements or fix any other errors listed above and try again.
```

**Expected behavior**
Installation of Solr

**Additional context**
https://github.com/TYPO3-Solr/ext-solr/blob/b8a800265fade592444b2eea6094c084004b864f/Resources/Private/Install/install-solr.sh#L15

There is no A-Record for mirror.dkd.de anymore?

dkd-dobberkau:
Apache Software Foundation has ceased to distribute its releases thru the mirrors. I will check if this way to install apache solr is still updated.

juergen-venne:
So what is the recommended alternative way to install solr in this context?

dkd-kaehm:
@premar Please test the script
from #3100 pull request.

The PR was inspired by original Apache Solr Dockerfile. https://github.com/docker-solr/docker-solr/blob/b786d89646823e5250368cb25fa2f55fdc50ec7e/8.9/Dockerfile#L29-L31

premar:
@dkd-kaehm looks good to me. Could download and install solr with the new script. Therefore the problem is resolved for me, thank you for the adjustments.

PR #3100, opened 2021-11-18
[TASK] Replace mirrors for Apache Solr binaries on install-solr.sh

dkd-kaehm:
The installation script is fixed and marked as not recommended variant of setting up Apache Solr server.

Fixes: #3094


dkd-kaehm:
We want a back ports for following branches:
* release-11.5.x
* release-11.1.x
* release-11.0.x

Issue #3099, opened 2021-11-18
[BUG] Error thrown when using routeEnhancer and empty facet


Nicola-Widmer:
**Describe the bug**
An error gets thrown when sending an empty facet while using an routeEnhancer

**To Reproduce**
1. Configure RouteEnhancer
2. Add facet as input-field in search form
3. Send empty form
4. Error gets thrown

**Expected behavior**
Empty facet gets ignored

**Used versions (please complete the following information):**
 - TYPO3 Version: [e.g. 10.4.21]
 - Browser: All
 - EXT:solr Version: [e.g. 11.1.2]

**Additional context**
Already fixed it. I'll create an pr after this issue

PR #3101, opened 2021-11-18
[Bugfix] routeenhancer with empty filters

Nicola-Widmer:
Skips empty filters so no error gets thrown when using search & facets

# What this pr does

When using routeEnhancers and filters as inputs, there is a possibility that an filter is empty  which results in an url like
`?tx_solr[q]=&tx_solr[filter][0]=`. This would throw an error

This pr adds an check if the filter is completely empty and skips it

# How to test
Use an routeEnhancer and add an facet to the search form. Leave the fields empty and send

Fixes: #3099


dkd-kaehm:
Thanks for fixing that.
We want a back port for following branches:
* release-11.5.x
* release-11.1.x

Nicola-Widmer:
I've prepared following backports:

- #3105 
- #3106

dkd-kaehm (approved):
:+1:
Thanks for fixing.
The PR will be merged.

Issue #3108, opened 2021-11-22
[BUG] Solr response 400 ("Bad request") due to illegal characters?


kitzberger:
**Describe the bug**

A 400 ("Bad request") response from solr doesn't seem to be handled as an error but leads to the indexing queue being stuck at where it's at.

**To Reproduce**

We've got the following config for `tx_news_domain_model_news` records:

```
plugin.tx_solr.index.queue.tx_news_domain_model_news.fields {
      ....
      teaser = teaser

      content = SOLR_CONTENT
      content {
        cObject = COA
        cObject {
          10 = TEXT
          10 {
            field = bodytext
            noTrimWrap = || |
          }

          20 = SOLR_RELATION
          20 {
            localField = content_elements
            foreignLabelField = bodytext
            singleValueGlue = | |
          }
        }
      }
      ...
}
```

The error can only be seen after adding an `else` into `EXT:solr/Classes/IndexQueue/Index.php->indexItem()`:

```php
            if ($response->getHttpStatus() == 200) {
                $itemIndexed = true;
            } else {
                var_dump($item);
                var_dump($itemDocument);
                var_dump($response);
                throw new Exception('Solr Response: ' . $response->getHttpStatusMessage(), 1637318787);
            }
```

Here's the error message:
```
Error adding field 'content'='lorem ipsum' msg=Multiple values encountered for non multiValued copy field teaser: lorem ipsum"
```

I can only assume that it's gotta be related to that part of `general_schema_fields.xml`:

```xml
	<copyField source="content" dest="teaser" maxChars="300" />
```

**Expected behavior**

The indexing failure should be properly logged as an error and the item thus skipped on subsequent runs.

**Used versions:**
 - TYPO3 Version: 9 and 10
 - EXT:solr Version: 11.0.x
 - Used Apache Solr Version: 8.2.0
 - PHP Version: 7.4.0

dkd-friedrich:
After a brief check this issue is not related to #2903.

`IndexService` will mark failed items if an exception is thrown, but not if the item just isn't indexed. An item not being indexed is valid, since other conditions like the start time are considered and may prevent indexing for the time of the index run. 

If an item is not indexed due to it's start time a retry is required. But if indexing fails due to a Solr issue, the item should be marked as failed to prevent that further indexing is blocked. Indexing also might be blocked if there are many items which can not be indexed right now due to their start time, but this should be discussed separately.

dkd-friedrich:
[...]

PR #3109, opened 2021-11-22
Handle solr responses != 200

kitzberger:
# What this pr does

This make sure that a non 200 response from solr is handled as a proper error. This patch is considered as a workaround rather than the actual solution. I leave that to the pro's ;-)

# How to test

Make solr return a status 400 and nothing should happen to the index queue item. Then apply this patch and the item should be handled as an error and subsequent runs should be processing the next items in line.

Related: #3108

### DIFF
--- a/Classes/IndexQueue/Indexer.php
+++ b/Classes/IndexQueue/Indexer.php
@@ -191,6 +191,8 @@ class Indexer extends AbstractIndexer
             $response = $this->solr->getWriteService()->addDocuments($documents);
             if ($response->getHttpStatus() == 200) {
                 $itemIndexed = true;
+            } else {
+                throw new Exception('Solr Response: ' . $response->getHttpStatusMessage(), 1637318787);
             }
         } catch (HttpException $e) {
             $response = new ResponseAdapter($e->getBody(), $httpStatus = 500, $e->getStatusMessage());


dkd-kaehm:
@kitzberger 
Thanks for reporting and advice.
It seems the exceptions are caught in `IndexService`

https://github.com/TYPO3-Solr/ext-solr/blob/03080d4d55eeb9d50b15348f445d23e57e34e461/Classes/Domain/Index/IndexService.php#L127-L138

The newly added exception will not interrupt the indexing process, and the error message will be delegated to the index queue item. That is fine to not to interrupt the process, but that all is at least strange, because
the `Solarium//HttpException` exception, which is caught in next lines after your change, should be thrown on 400+ status codes. See description of `Solarium//HttpException`  :
 
https://github.com/solariumphp/solarium/blob/5b2cc069b30bd990e9130e7a3b8c54ce4b1fe04a/src/Exception/HttpException.php#L12-L27

See example:

https://github.com/solariumphp/solarium/blob/090c447cdab59051191f936cbaec905752bcc7b1/src/Core/Query/Result/Result.php#L58-L75

---

IMHO we should inspect the try-catch block and why is HttpException not thrown on 400+.

dkd-friedrich:
I am closing this PR as the problem has been resolved in the meantime, see https://github.com/TYPO3-Solr/ext-solr/pull/3539

Issue #3096, opened 2021-11-14
[BUG] Autosuggest for german Umlaute not working in Ext:solr 11


ulrike-cosmoblonde:
Hi,

I have been using the autosuggest feature in several projects on German content without any Umlaut issues with the solr extension v10.
When using the schema.xml from the solr extension version 10 in my solr instance, then autosuggest works for Umlaute.
But with the latest schema.xml shipped with the solr extension version 11, Umlaute are no longer working in autosuggest.
So entering "Künst" does not produce any suggest results, but entering "Kunst" does show e.g. Kunstler, kunstlich, etc.

When comparing the configuration for the field type textSpell (which is the recommended field type for autosuggest) in both schemas, the difference and hence the problem is the additional  filter "solr.ASCIIFoldingFilterFactory" in the index and query analyzer. 

```
<!-- Setup simple analysis for spell checking -->
<fieldType name="textSpell" class="solr.TextField" positionIncrementGap="100" omitNorms="true">
<analyzer type="index">
	<tokenizer class="solr.StandardTokenizerFactory"/>
	<filter class="solr.LowerCaseFilterFactory"/>

	<filter class="solr.DictionaryCompoundWordTokenFilterFactory"
		dictionary="german/german-common-nouns.txt"
		minWordSize="5"
		minSubwordSize="4"
		maxSubwordSize="15"
		onlyLongestMatch="false"
	/>

	<!-- no synonyms here because we do not want to add them as spell suggestion -->
	<filter class="solr.ManagedStopFilterFactory" managed="${solr.core.name}"/>
	<filter class="solr.ASCIIFoldingFilterFactory"/>

	<filter class="solr.RemoveDuplicatesTokenFilterFactory"/>
</analyzer>
<analyzer type="query">
	<tokenizer class="solr.StandardTokenizerFactory" />

	<filter class="solr.LowerCaseFilterFactory"/>

	<filter class="solr.ManagedSynonymGraphFilterFactory" managed="${solr.core.name}" />
	<filter class="solr.ManagedStopFilterFactory" managed="${solr.core.name}"/>
	<filter class="solr.ASCIIFoldingFilterFactory"/>

	<filter class="solr.RemoveDuplicatesTokenFilterFactory"/>
</analyzer>
</fieldType>
```

When removing this filter in both analyzer and re-indexing all data, then Umlaute work again for autosuggest.

[...]

PR #3118, opened 2021-12-08
[BUGFIX:BP:11.1] Fix autosuggest with non-ascii terms

dkd-friedrich:
# What this pr does
The introduced ASCII folding filters or language depending normalization filters lead to issues with the auto suggest function due to the differing stemming behaviour.

To fix this issue the original token is preserved if possible, this e.g. allows suggestions for search terms with and without accents. As this extension might lead to unwanted duplicates a new field textSpellExact is introduced, which considers non-ascii characters as given.

# How to test

As soon as branch is checked out and configset is updated, you can test the  default auto suggest behaviour, e.g. with german cores and documents containing "künstler" you should get suggestions for "kunst" and "künst".

By switching the spell field via `plugin.tx_solr.suggest.suggestField = spellExact` you should get more exact suggestions, as they used to be in EXT:solr prior 11.0.

Fixes: #3096

### DIFF
--- a/Documentation/Appendix/DynamicFieldTypes.rst
+++ b/Documentation/Appendix/DynamicFieldTypes.rst
@@ -73,6 +73,8 @@ Extension               Type                                 Multivalue  Comment
 \*_textExactM           textExact                            Yes
 \*_textSpellS           textSpell                            No
 \*_textSpellM           textSpell                            Yes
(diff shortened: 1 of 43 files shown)


dkd-kaehm (approved):
LTGM
:+1:

Issue #3126, opened 2021-12-15
[BUG] SearchUriBuilder always adds `tx_solr.filter`


rengaw83:
**Describe the bug**

Since https://github.com/TYPO3-Solr/ext-solr/commit/45a45f4af53a92959b46cf1f7849a852c76713a4 the `&tx_solr[filter]=` parameter is added to each URI from the SearchUriBuilder, even if it was not set beforehand.

The following line forces the creation of `tx_solr.filter` in arguments, since the arguments array is handled as reference.
```php
$this->sortFilterParametersIfNecessary($previousSearchRequest, $arguments['tx_solr']['filter']);
```

**To Reproduce**
Steps to reproduce the behavior for facet links:
(bug exists for all links created by SearchUriBuilder)
1. Configure facets for tx_solr
2. Change the pluginNameSpace to `sitesearch`
4. See `&tx_solr[filter]=` in the generated facet URI's

**Expected behavior**
1. The SearchUriBuilder should check the configured namespace before he calls `sortFilterParametersIfNecessary`
2. The SearchUriBuilder should check if `$arguments['tx_solr']['filter']` is already set or not before `sortFilterParametersIfNecessary` call was performed.
Currently, if `namespace.filter` does not exists already, it will be created since `$filterArguments` is handled as reference.

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.21
 - EXT:solr Version: 11.1.2
 - PHP Version: 7.4

dkd-kaehm:
@rengaw83 
Thanks for fixing that issue.
Please re-push to all pull requests, to trigger the builds.

rengaw83:
Hi @dkd-kaehm,
have created PR for 11.5 branch in #3137 (ty for fast version correction in title ;)).
PR for 11.1 was repushed in #3128.
PR for main was rebased and repushed in #3127. But build generally don't seem to work so far?
Let me know if there is still something i can do.

dkd-kaehm:
@rengaw83 
Thanks for the fixes. 
I'll merge all the PRs after successful build on 11.1.

PR #3127, opened 2021-12-15
[Bugfix] Prevent unwanted filter parameters from being generated (main)

rengaw83:
The following line forces the creation of `tx_solr.filter` in arguments, since the arguments array is handled as reference and do not observe `pluginNamespace` configuration:
```php
$this->sortFilterParametersIfNecessary($previousSearchRequest, $arguments['tx_solr']['filter']);
```

This PR fixes the folowing issues:
1. `tx_solr.filter` are only sorted if present in arguments array
2. `tx_solr.filter` are not forced to be created anymore
3. namespace configuration is observed 

Fixes: #3126 for main.

Fixes: #3126 for main.
See #3128 for v11.1 (TYPO3 10).

### DIFF
--- a/Classes/Domain/Search/Uri/SearchUriBuilder.php
+++ b/Classes/Domain/Search/Uri/SearchUriBuilder.php
@@ -118,7 +118,7 @@ class SearchUriBuilder
 
         $arguments = $persistentAndFacetArguments + $additionalArguments;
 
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
Thanks for the fix.

Issue #3153, opened 2022-01-12
[FEATURE] Improve data update handling


dkd-friedrich:
**Is your feature request related to a problem? Please describe.**
To ensure the Solr index is up-to-date an extensive monitoring is done, in huge sites this may slow down the TYPO3 backend, as many records and pages have to up checked and updated

**Describe the solution you'd like**
Decoupling of the monitoring from the work in the TYPO3 backend and enabling separate execution via scheduler task. In the first step, 3 options are forseen:
- direct execution (status quo)
- downstream execution via scheduler
- deactivation of monitoring

Next step could be the implementation of a further option that could enable direct execution via a detached CLI process.

**Target versions**
11.2, 11.5

PR #3154, opened 2022-01-12
[FEATURE] Improve data update handling

dkd-friedrich:
# What this pr does

Decouples the monitoring and data update handling and introduces 3 options of data monitoring

- immediate (status quo)
- delayed
- no monitoring

# How to test

Configure the different `monitoringType`s in the extension manager and perform data updates, in the end the index queue and data updates should be performed as usual, but the execution moment differs.

By default (`immediate`) all updates will be performed immediately, like in all prior versions of the Solr extension. In the delayed mode the events will be queued and processed by the Scheduler `Event Queue Worker` task. If this option is active the queue and index updates will be delayed and take place as soon as the task is executed.

To implement an own monitoring the monitoring can also be disabled, if you choose the no monitoring option, no updates will be processed.

Resolves: #3153

### DIFF
--- a/Classes/GarbageCollector.php
+++ b/Classes/GarbageCollector.php
@@ -24,13 +24,15 @@ namespace ApacheSolrForTypo3\Solr;
  *  This copyright notice MUST APPEAR in all copies of the script!
  ***************************************************************/
 
-use ApacheSolrForTypo3\Solr\Domain\Index\Queue\GarbageRemover\StrategyFactory;
-use ApacheSolrForTypo3\Solr\IndexQueue\Queue;
+use Psr\EventDispatcher\EventDispatcherInterface;
 use ApacheSolrForTypo3\Solr\System\TCA\TCAService;
-use TYPO3\CMS\Backend\Utility\BackendUtility;
 use TYPO3\CMS\Core\DataHandling\DataHandler;
 use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use ApacheSolrForTypo3\Solr\Domain\Index\Queue\UpdateHandler\GarbageHandler;
+use ApacheSolrForTypo3\Solr\Domain\Index\Queue\UpdateHandler\Events\RecordDeletedEvent;
+use ApacheSolrForTypo3\Solr\Domain\Index\Queue\UpdateHandler\Events\PageMovedEvent;
+use ApacheSolrForTypo3\Solr\Domain\Index\Queue\UpdateHandler\Events\RecordGarbageCheckEvent;
 
 /**
  * Garbage Collector, removes related documents from the index when a record is
@@ -41,7 +43,7 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
  * @author Ingo Renner <ingo@typo3.org>
  * @author Timo Schmidt <timo.schmidt@dkd.de>
  */
-class GarbageCollector extends AbstractDataHandlerListener implements SingletonInterface
+class GarbageCollector implements SingletonInterface
 {
     /**
      * @var array
@@ -54,13 +56,20 @@ class GarbageCollector extends AbstractDataHandlerListener implements SingletonI
     protected $tcaService;
 
     /**
-     * GarbageCollector constructor.
-     * @param TCAService|null $TCAService
+     * @var EventDispatcherInterface
      */
-    public function __construct(TCAService $TCAService = null)
+    protected $eventDispatcher;
+
+    /**
+     * GarbageCollector constructor.
+     *
+     * @param TCAService|null $TCAService
+     * @param EventDispatcherInterface|null $eventDispatcher
+     */
+    public function __construct(TCAService $TCAService = null, EventDispatcherInterface $eventDispatcher = null)
     {
-        parent::__construct();
         $this->tcaService = $TCAService ?? GeneralUtility::makeInstance(TCAService::class);
+        $this->eventDispatcher = $eventDispatcher ?? GeneralUtility::makeInstance(EventDispatcherInterface::class);
     }
 
     /**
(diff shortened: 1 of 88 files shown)


dkd-kaehm:
:+1:
Please copy the rows https://github.com/TYPO3-Solr/ext-solr/blob/release-11.1.x/.github/workflows/ci-matrix.json#L12-L15
to release-11.2.x

dkd-friedrich:
> 👍 Please copy the rows https://github.com/TYPO3-Solr/ext-solr/blob/release-11.1.x/.github/workflows/ci-matrix.json#L12-L15 to release-11.2.x

Right, I've already done that while preparing the 11.2 branch:
=> https://github.com/TYPO3-Solr/ext-solr/commit/a22f5dda90f33a9eb9992fd84fc63f2f84b99625

We also have to update the configsets for 11.2, but we can do that with the discussed update to Solr 8.11

dkd-kaehm:
> > +1 Please copy the rows https://github.com/TYPO3-Solr/ext-solr/blob/release-11.1.x/.github/workflows/ci-matrix.json#L12-L15 to release-11.2.x
> 
> Right, I've already done that while preparing the 11.2 branch: => [a22f5dd](https://github.com/TYPO3-Solr/ext-solr/commit/a22f5dda90f33a9eb9992fd84fc63f2f84b99625)
> 
> We also have to update the configsets for 11.2, but we can do that with the discussed update to Solr 8.11

We want this changes in master branch as well. 
And https://github.com/TYPO3-Solr/ext-solr/blob/main/.github/workflows/ci.yml#L3-L9

Classes/IndexQueue/RecordMonitor.php
```diff
@@ -173,21 +64,16 @@ protected function getUpdateSubPagesRecursiveTriggerConfiguration()
...
-        DataHandler $tceMain
-    ) {
+        $uid
+    ): void {
         if ($command === 'delete' && $table === 'tt_content' && $GLOBALS['BE_USER']->workspace == 0) {
-            // skip workspaces: index only LIVE workspace
-            $pid = $this->getValidatedPid($tceMain, $table, $uid);
-            $this->indexQueue->updateItem('pages', $pid, time());
+            $this->eventDispatcher->dispatch(
+                new ContentElementDeletedEvent((int)$uid)
```
dkd-kaehm, line 75: Do we not lose the PID there?
Or we collect the PID via soft deleted records?

Classes/IndexQueue/RecordMonitor.php
```diff
@@ -300,36 +125,25 @@ protected function applyRecordChangesToQueue($table, $uid, $pid)
...
+        $table = (string)$table;
         if ($this->skipMonitoringOfTable($table)) {
             return;
         }
 
         if ($status === 'new' && !MathUtility::canBeInterpretedAsInteger($recordUid)) {
             $recordUid = $tceMain->substNEWwithIDs[$recordUid];
         }
-        if ($this->isDraftRecord($table, $recordUid)) {
+        if (Util::isDraftRecord($table, (int)$recordUid)) {
```
dkd-kaehm, line 139: We should avoid static methods and Util classes.
See https://phpunit.readthedocs.io/en/9.5/test-doubles.html

Classes/System/Solr/Node.php
```diff
@@ -93,7 +93,7 @@ public static function fromArray(array $configuration): Node
      *
      * @param array  $configuration
      * @param string $name
-     * @throws |UnexpectedValueException
+     * @throws |\UnexpectedValueException
```
dkd-kaehm, line 96: Using use import is better.
Why is pipe here?

dkd-kaehm (approved):
LGTM
can be merged after PageEvent instantiation.

Classes/IndexQueue/RecordMonitor.php
dkd-friedrich: Right, there is no need to validate and store the pid for tt_content records here, as we know that tt_content uses soft deletes and we can do that processing downstream.

The used `DataHandler` method `getPID` that is used to determine and validate the pid, ignores the `DeletedRestriction`.

Classes/System/Solr/Node.php
dkd-friedrich: Looks like an error to me, I couldn't find a reason for it in the history. But you're right, I should have removed the pipe directly, I've done that now.

I've also added an import for `UnexpectedValueException`

Classes/GarbageCollector.php
```diff
@@ -164,15 +116,9 @@ protected function deleteSubEntriesWhenRecursiveTriggerIsRecognized($table, $uid
...
-            // ...still needed?
-
-            // must be removed from index since the pid changes and
-            // is part of the Solr document ID
-            $this->collectGarbage($table, $uid);
-
-            // now re-index with new properties
-            $this->getIndexQueue()->updateItem($table, $uid);
+            $this->eventDispatcher->dispatch(
+                new PageMovedEvent((int)$uid, (string)$table)
```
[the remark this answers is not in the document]
dkd-friedrich: Nice catch, I've removed the unneeded table parameter

Classes/IndexQueue/RecordMonitor.php
dkd-friedrich: I agree in principle, we should reconsider the concept of the Util class, especially with regard to the [TYPO3 coding guidelines](https://docs.typo3.org/m/typo3/reference-coreapi/11.5/en-us/CodingGuidelines/CglPhp/PhpArchitecture/ModelingCrossCuttingConcerns/StaticMethods.html) and notes on static methods and utitily classes. 

The old class method `isDraftRecord`  was only a wrapper for Util, but since the calls in our unit tests are unproblematic at this point and it is also considered bad practice to test against mocks, the reintroduction of the wrapper method wouldn't really help us in my view.

dkd-kaehm (approved):
:+1:

Issue #3158, opened 2022-01-14
[BUG:11.5] Fulfill: Breaking: #92502 - Make Extbase handle PSR-7 responses only.


dkd-kaehm:
**Describe the bug**
All controllers must be refactored to fulfill the TYPO3 11 BC. See https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Breaking-92502-MakeExtbaseHandlePSR7ResponsesOnly.html  

Todo:
- [x] #3138

- [x] Fix skipped `\ApacheSolrForTypo3\Solr\Tests\Integration\Controller\SearchControllerTest::frontendWillRenderErrorMessageForSolrNotAvailableAction()`
- [x] Fix skipped `\ApacheSolrForTypo3\Solr\Tests\Integration\Controller\SearchControllerTest::frontendWillForwardToErrorActionWhenSolrEndpointIsNotAvailable()`

Relates: #3138

dkd-kaehm:
Done in https://github.com/TYPO3-Solr/ext-solr/commit/d40bcd67afc6949cb3017ae4638f991871cfb14b

PR #3159, opened 2022-01-14
[BUGFIX] #3158: "search unavailable" is no longer displayed

dmitryd:
# What this pr does

This PR fixes "search unavailable" missing on Frontend.

Note: one method is removed because it is overriden and it cannot return
values according to method signature. Therefore simply return the response
directly.

Related to: #3158


dkd-kaehm:
Done in https://github.com/TYPO3-Solr/ext-solr/commit/d40bcd67afc6949cb3017ae4638f991871cfb14b

Issue #3161, opened 2022-01-17
[BUG] Errors are not always evaluated correctly


Smove:
**Describe the bug**
In the commit a2988d2ff699b69ae571f1934e37a8c17255fb7f, care was taken to catch all \Throwables. However, the method to generate the error log was not adjusted.

**Expected behavior**
The error should be logged correctly

**Screenshots**
![image](https://user-images.githubusercontent.com/3686891/149762332-85c3ec8b-14da-4765-8f3c-c6a1001a0046.png)

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.5
 - EXT:solr Version: 11.5-dev
 - PHP Version: 7.4

dkd-kaehm:
done in #3162

PR #3162, opened 2022-01-17
Allow to generate indexing error log from throwable

Smove:
Since we catch \Throwable instead of \Exception, we should change the typehint of the error log generating method, too. This will prevent errors while catching \Error instances.

# What this pr does

This pr will change the method signature of ApacheSolrForTypo3\Solr\Domain\Index\IndexService::generateIndexingErrorLog 

Fixes: #3161


dkd-kaehm:
@Smove 
Thanks for the fix.
Please rebase to release-11.5.x, to be able to merge this PR.

PR #3165, opened 2022-01-20
[BUGFIX] Prevent some "undefined array key" warnings with php 8

rengaw83:
In TYPO11 with PHP 8 there are some Errors like *PHP Warning: Undefined array key*.
This PR adds some additional checks and prevents this errors.

The additional options in IndexingConfigurationSelectorField are required to prevent warnings in TYPO3 core Backend FormElements and FieldWizards.

### DIFF
--- a/Classes/FieldProcessor/PageUidToHierarchy.php
+++ b/Classes/FieldProcessor/PageUidToHierarchy.php
@@ -68,10 +68,15 @@ class PageUidToHierarchy extends AbstractHierarchyProcessor implements FieldProc
         $results = [];
 
         foreach ($values as $value) {
-            list($rootPageUid, $mountPoint) = GeneralUtility::trimExplode(',',
-                $value, true, 2);
-            $results[] = $this->getSolrRootlineForPageId($rootPageUid,
-                $mountPoint);
+            $results[] = $this->getSolrRootlineForPageId(
+                /** @scrutinizer ignore-type */
+                ...GeneralUtility::trimExplode(
+                    ',',
+                    $value,
+                    true,
+                    2
+                )
+            );
         }
 
         return $results;
--- a/Build/Test/IntegrationFrontendTests.xml
+++ b/Build/Test/IntegrationFrontendTests.xml
@@ -7,8 +7,8 @@
     convertWarningsToExceptions="true"
     forceCoversAnnotation="false"
     processIsolation="true"
-    stopOnError="true"
-    stopOnFailure="true"
+    stopOnError="false"
+    stopOnFailure="false"
     stopOnIncomplete="false"
     stopOnSkipped="false"
     verbose="false">
--- a/Classes/Backend/IndexingConfigurationSelectorField.php
+++ b/Classes/Backend/IndexingConfigurationSelectorField.php
@@ -180,6 +180,7 @@ class IndexingConfigurationSelectorField
     {
         $parameterArray = [
             'fieldChangeFunc' => [],
+            'itemFormElID' => $this->formElementName,
             'itemFormElName' => $this->formElementName,
             'itemFormElValue' => $selectedValues,
             'fieldConf' => ['config' => ['items' => $items]],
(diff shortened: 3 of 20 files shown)


dkd-kaehm:
@rengaw83 Thanks for the fix, I'll look at that soonish.

dkd-kaehm:
I rebased your PR against current state and disabled stop running tests on failures, to let the build to show all PHP8 troubles within Search/FE context.

dkd-kaehm:
@rengaw83 
would you like to fix the rest of PHP 8 complains 
See https://github.com/TYPO3-Solr/ext-solr/runs/4899118383?check_suite_focus=true#step:12:104

rengaw83:
yes, i'll look at the rest of php warnings next week.

rengaw83:
It isn't that easy at the moment @dkd-kaehm.  
Have created local test database and test solr, but couldn't get the integration tests to work on my machine. Only unit tests are working.  
I use the github actions now, but the test execution stops at the first warning. So i has to fix the errors step by step. Hope it is okay.

rengaw83:
@dkd-kaehm have fixed all php8 issues.

Currently the `SearchControllerTest::canSeeTheParsedQueryWhenABackendUserIsLoggedIn` test fails, because the string `Parsed Query:` is not in the html content.
With this change https://github.com/TYPO3-Solr/ext-solr/pull/3165/files#diff-3cd6c649c12efcf70c18bbccf56c351152a54f258fdaddc9d3952d929be33fe0R49-R57 i have added some checks if the parsed query exists. If not, no output will be generated.

Just to understand and to be sure: The view helper wants to output the parsed query when it's empty?  
Should i change the condition back into following, so the parsed query prints out even it is not present in debug responce?
```
if ($backendUserIsLoggedIn === true && $resultSet && $resultSet->getUsedSearch() !== null) {
    $content = '<br><strong>Parsed Query:</strong><br>';
    if ($resultSet->getUsedSearch()->getDebugResponse()
	      && !empty($resultSet->getUsedSearch()->getDebugResponse()->parsedquery)) {
	      $content .= htmlspecialchars($resultSet->getUsedSearch()->getDebugResponse()->parsedquery);
	  }
}
```

dkd-kaehm:
> @dkd-kaehm have fixed all php8 issues.
> 
> Currently the `SearchControllerTest::canSeeTheParsedQueryWhenABackendUserIsLoggedIn` test fails, because the string `Parsed Query:` is not in the html content. With this change https://github.com/TYPO3-Solr/ext-solr/pull/3165/files#diff-3cd6c649c12efcf70c18bbccf56c351152a54f258fdaddc9d3952d929be33fe0R49-R57 i have added some checks if the parsed query exists. If not, no output will be generated.
> 
> Just to understand and to be sure: The view helper wants to output the parsed query when it's empty? Should i change the condition back into following, so the parsed query prints out even it is not present in debug responce?
> 
> ```
> if (true === $backendUserIsLoggedIn && $resultSet && null !== $resultSet->getUsedSearch()) {
>     $content = '<br><strong>Parsed Query:</strong><br>';
>     if (null !== $resultSet->getUsedSearch()->getDebugResponse()
>         && !empty($resultSet->getUsedSearch()->getDebugResponse()->parsedquery)) {
>         $content .= htmlspecialchars($resultSet->getUsedSearch()->getDebugResponse()->parsedquery);
>     }
> }
> ```

IMHO the container should be visible if BE user logged in, no matter if it is empty or not. I think the current state should be restored and the question transferred to an issue if you think, there is something wrong.

rengaw83:
ty for fast responce @dkd-kaehm, have restored the output. all issues are fixed now, all tests are green.

dkd-kaehm:
I'll pick up the that patch in #3151 by rebasing and close manually after merging #3151.
@rengaw83 Thanks for the fixes.

Classes/FieldProcessor/PageUidToHierarchy.php
```diff
@@ -68,10 +68,14 @@ public function process(array $values)
...
-            $results[] = $this->getSolrRootlineForPageId($rootPageUid,
-                $mountPoint);
+            $results[] = $this->getSolrRootlineForPageId(
+                ...GeneralUtility::trimExplode(
+                    ',',
+                    $value,
+                    true,
+                    2
+                )
+            );
```
dkd-kaehm, line 78: Is that change really wanted?
See: https://scrutinizer-ci.com/g/TYPO3-Solr/ext-solr/inspections/5efe43ac-cf9b-4387-90e7-704f8507fb30/issues/files/Classes/FieldProcessor/PageUidToHierarchy.php?status=new&selectedLabels%5B0%5D=9&orderField=path&order=asc&fileId=Classes/FieldProcessor/PageUidToHierarchy.php&honorSelectedPaths=0

I would like to avoid the new scrutinizer issues if possible.

dkd-kaehm (changes requested):

Classes/FieldProcessor/PageUidToHierarchy.php
rengaw83: GeneralUtility::trimExplode creates an array with one or two keys.  
The method getSolrRootlineForPageId expects one or two arguments. 
The `...` can be used when calling functions to unpack an array or Traversable variable or literal into the argument list.  
See [php manual](https://www.php.net/manual/en/functions.arguments.php#example-162).

The old code produces errors when trimExplode returns an array with only one key.
```php
list($rootPageUid, $mountPoint) = GeneralUtility::trimExplode(',', $value, true, 2);
$results[] = $this->getSolrRootlineForPageId($rootPageUid, $mountPoint);
```

When the `...` argument list is a problem, i can change the code to this:
```php
$args = GeneralUtility::trimExplode(',', $value, true, 2);
$results[] = $this->getSolrRootlineForPageId($args[0] ?? 0, $args[1] ?? '');
```

Let me know how we want to proceed.

Issue #3173, opened 2022-01-25
[BUG] PHP Warning: Undefined array key "TSFE" in RootPageResolver.php line 161


DrWh0286:
**Describe the bug**
In case of saving records with pid=0 with php 8 a warning appears:

PHP Warning: Undefined array key "TSFE" in /var/www/html/public/typo3conf/ext/solr/Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php line 161

**To Reproduce**
1. Go to page with uid=0
2. Open a record for editing (e.g. be_groups, be_users)
3. Save the record

**Expected behavior**
Having no php warning after saving a record.

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.5
 - Browser: Chrome
 - EXT:solr Version: task/2976_TYPO3.11_compatibility (0e8702a1a6e685b5df17b2ab489db67b41f3d47a)
 - Used Apache Solr Version: 8.5.2 (typo3solr/ext-solr:11.0.3 image with ddev)
 - PHP Version: 8.0
 - MySQL Version: MariaDB 10.3

dkd-kaehm:
Fixed in: 6800394c02414aa6e22b47bd3c7d4fbf3921e533

PR #3174, opened 2022-01-25
[FIX] Undefined array key "TSFE" in class RootPageResolver

DrWh0286:
Fixes: #3173

# What this pr does

Avoiding call on $GLOBALS['TSFE'] if not set

# How to test
Running TYPO3 11.5 with php 8

1. Open page with uid=0 in list mode
2. Open record for editing (e.g. be_groups, be_users)
3. Save the record
4. No PHP warning should appear

Fixes: #3173

### DIFF
--- a/Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php
+++ b/Classes/Domain/Index/Queue/RecordMonitor/Helper/RootPageResolver.php
@@ -158,7 +158,7 @@ class RootPageResolver implements SingletonInterface
     {
         /** @var Rootline $rootLine */
         $rootLine = GeneralUtility::makeInstance(Rootline::class);
-        $rootPageId = intval($pageId) ?: intval($GLOBALS['TSFE']->id);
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
Thanks for fixing that, please rebase to release-11.5.x.

DrWh0286:
@dkd-kaehm I did so. Now it seems, that some checks are crashed 🤔 . Can you maybe have a short look please?

dkd-kaehm:
IMHO the `$GLOBALS['TSFE']` will be newer set within the given context.
The fallback to `$GLOBALS['TSFE']` could be removed, but all usages of `getRootPageId()` method checked, that the `$pageId` is set propely.

DrWh0286:
Yes I agree. Especially because in this case with php version < 8 the method returns 0 anyway. So it seems to have no impact.

PS: To be honest I needed some working state of the extension for the project starting next week 😉 . So I already use my commit as patch. That's the reason for this pragmatic fix.

dkd-kaehm:
@DrWh0286 
Thanks for contribution, 
I'll close that PR in favour of #3206. In #3206 we'll try to remove `$GLOBALS['TSFE']` usages in whole class.

PR #3185, opened 2022-02-02
[BUGFIX] Change backend index queue amount to 10

georgringer:
In a246cb8e34d80da6227445c0c800620f16264a00 the amount changed to 1 but should be 10.
In the long run it should be configurable


dkd-kaehm:
IMHO we do not want 10 Items, because this button is for debugging purposes only. 
I would propose to mark this button with the hint.

PR #3186, opened 2022-02-03
[BUGFIX] Fix notice in TranslateViewHelper

georgringer:
If using PHP 8.1, the following exception is thrown

```
PHP Runtime Deprecation Notice: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in
```

Cast the value to a string to avoid the exception


dkd-kaehm:
The main branch will be merged with release-11.5.x after release.
The fix is in 3b91901e613d90c74aec01b0aac7e13f26a608d1

PR #3188, opened 2022-02-03
[BUGFIX] Fix notice in TranslateViewHelper

georgringer:
If using PHP 8.1, the following exception is thrown

```
PHP Runtime Deprecation Notice: strpos(): Passing null to parameter #1 ($haystack) of type string is deprecated in
```

Cast the value to a string to avoid the exception


dkd-kaehm (approved):
:+1: 
thanks

PR #3193, opened 2022-02-07
[P:11.5:FEATURE] Improve data update handling

dkd-friedrich:
# What this pr does

Ports the improved data update handling included in 11.2 to upcoming 11.5.

Decouples the monitoring and data update handling and introduces 3 options of data monitoring

- immediate (status quo)
- delayed
- no monitoring

# How to test

Configure the different `monitoringType`s in the extension manager and perform data updates, in the end the index queue and data updates should be performed as usual, but the execution moment differs.

By default (`immediate`) all updates will be performed immediately, like in all prior versions of the Solr extension. In the delayed mode the events will be queued and processed by the Scheduler `Event Queue Worker` task. If this option is active the queue and index updates will be delayed and take place as soon as the task is executed.

To implement an own monitoring the monitoring can also be disabled, if you choose the no monitoring option, no updates will be processed.

Resolves: #3153


dkd-kaehm (approved):
LGTM

Issue #2916, opened 2021-04-15
[BUG] solr_path_write is not used


EmanuelOster:
**Describe the bug**
The plugin allows the configuration of a separate write connection for solr and even offers part of it through the UI.
However, the `solr_path_write` configuration is completely ignored and instead the `solr_path_read` configuration is used instead.
The same happens if I try to configure the `solr_core_write` option for a given language. `solr_core_read` is used instead.

**To Reproduce**
1. Go to 'Sites', open a site configuration and switch to the Solr-Tab
2. Configure a different path for the write configuration and save
3. Go to Apache Solr->Info
4. `Following Apache Solr servers have been contacted:` is displayed with the write configuration for host and Port, but read configuration for path and core.

**Expected behavior**
Only the write configuration is used for write connections.

**Screenshots**
![image](https://user-images.githubusercontent.com/23719567/114865246-4201e900-9df2-11eb-9fc2-9d67178df226.png)
![image](https://user-images.githubusercontent.com/23719567/114865298-534af580-9df2-11eb-929e-915ac0ba14b1.png)

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.12
 - EXT:solr Version: 11.0.3
 - Used Apache Solr Version: 8.8.1

**Additional context**
I took a look at the source code and found the relevant code in the SiteRepository:
```php
'write' => [
                        'scheme' => SiteUtility::getConnectionProperty($typo3Site, 'scheme', $languageUid, 'write', 'http'),
                        'host' => SiteUtility::getConnectionProperty($typo3Site, 'host', $languageUid, 'write', 'localhost'),
                        'port' => (int)SiteUtility::getConnectionProperty($typo3Site, 'port', $languageUid, 'write', 8983),
                        // @todo: transform core to path
                        'path' =>
                            SiteUtility::getConnectionProperty($typo3Site, 'path', $languageUid, 'read', '/solr/') .
                            SiteUtility::getConnectionProperty($typo3Site, 'core', $languageUid, 'read', 'core_en') . '/' ,
                        'username' => SiteUtility::getConnectionProperty($typo3Site, 'username', $languageUid, 'write', ''),
                        'password' => SiteUtility::getConnectionProperty($typo3Site, 'password', $languageUid, 'write', ''),
                        'timeout' => SiteUtility::getConnectionProperty($typo3Site, 'timeout', $languageUid, 'write', 0)
                    ],
```
[...]

PR #3200, opened 2022-02-14
[BUGFIX] Fix write connection

dkd-friedrich:
# What this pr does

Path of write connection is read from read scope, so different paths
couldn't be used, this is fixed by inserting the right scope.

# How to test

Configure read and write connections and use different path settings.

Fixes: #2916


dkd-kaehm (approved):
:+1:

PR #3214, opened 2022-03-07
[BUGFIX:11.5]  Access restricted pages can not be indexed on TYPO3 11.5

dkd-kaehm:
# What this pr does

Fixes indexing of access restricted pages.


dkd-kaehm:
@dmitryd 
Please check this out, because in Slack asked.

dmitryd:
@dkd-kaehm 

Unfortunately still not working for me :(

We have CEs on the page like this:

1. Anonymous
2. Login box (Hide at login)
3. Protected (user group 4)
4. Protected (user group 4)
5. Protected (user group 4)

Page itself is public.

So in `getIndexableContent()` it is like anonymous user despite FE user and group set.

dmitryd:
Tested, last commit fixes the issue for me. Thanks a lot for your work! 🙇 

If anybody else has the problem and wants to test it, you will need to clear all caches.

PR #3215, opened 2022-03-08
FIX: Argument 1 passed to UrlHelper::setPort() must be of the type int, string given

kitzberger:
# What this pr does

Fixes exception that occures when trying to override port for page indexer:

Bildschirmfoto vom 2022-03-08 14-20-52

# How to test

```
plugin.tx_solr.index.queue {
    pages = 1
    pages {
        indexer {
            frontendDataHelper {
                scheme = http
                host = localhost
                port = 80
            }
        }
    }
}
```


dkd-kaehm (commented):
:+1:

PR #3216, opened 2022-03-08
FIX: Argument 1 passed to AbstractSolrTask::setRootPageId() must be of the type int, string given

kitzberger:
# What this pr does

Fixes exception when setting up the scheduler task for force reindexing page.

Bildschirmfoto vom 2022-03-08 14-40-28


dkd-kaehm:
Thanks for the fix.

Issue #3225, opened 2022-03-14
[BUG:11.5] Changes on access restricted pages(fe_group is set) are not recognized


dkd-kaehm:
**Describe the bug**
Changes on access restricted pages(fe_group is set) are not recognized.

**To Reproduce**
1. Clear index queue
2. Change some access restricted(solr-ddev-site PID 82) page
3. Go to index queue module and see the empty queue

**Expected behavior**
The changed acces restricted page MUST be recognized and noted in index queue 

**Screenshots**
-

**Used versions (please complete the following information):**
 - TYPO3 Version: [e.g. 11.5]
 - EXT:solr Version: [ release-11.5.x @ [fbf20c4](https://github.com/TYPO3-Solr/ext-solr/commit/fbf20c41d0a5cad42c78448f42c298ef7570b4d4) ]
 - PHP Version: no matter but 7.4

**Additional context**
The error is located in lines below. 

https://github.com/TYPO3-Solr/ext-solr/blob/fbf20c41d0a5cad42c78448f42c298ef7570b4d4/Classes/FrontendEnvironment/Tsfe.php#L150-L154

Those lines do not work anymore since TYPO3 11.

Possible solutions:
* https://review.typo3.org/c/Packages/TYPO3.CMS/+/71235
* Something like #3214 or similar should be implemented

PR #3228, opened 2022-03-14
[BUGFIX] Fix feuser initialisation in BE context

dkd-friedrich:
# What this pr does

Observation of database records that might be indexed require an own initialization of TSFE and frontend users, as the fe_user initialization is incomplete access restricted pages can't be observed correctly.

This commit fixes this issue by simulating the required frontend groups.

# How to test

Configure page indexing, clear index queue and change an access protected page, page must be added to the index queue.

Resolves: #3225

### DIFF
--- a/Classes/FrontendEnvironment/Tsfe.php
+++ b/Classes/FrontendEnvironment/Tsfe.php
@@ -142,15 +142,19 @@ class Tsfe implements SingletonInterface
                 )
             );
 
+            /** @var FrontendUserAuthentication $feUser */
             $feUser = GeneralUtility::makeInstance(FrontendUserAuthentication::class);
             // for certain situations we need to trick TSFE into granting us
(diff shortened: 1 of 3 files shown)


Classes/FrontendEnvironment/Tsfe.php
```diff
@@ -151,6 +152,8 @@ protected function initializeTsfe(int $pageId, int $language = 0, ?int $rootPage
             if (!empty($pageRecord['fe_group'])) {
                 $userGroups = array_unique(array_merge($userGroups, explode(',', $pageRecord['fe_group'])));
```
dkd-kaehm, line 153:
```suggestion
                $userGroups = explode(',', $pageRecord['fe_group']);
```
IMHO the -1 group isn't relevant if other groups are set.
the [0] group has each user.

dkd-kaehm (approved):

PR #3258, opened 2022-05-23
[TASK] Add allowed composer plugins

simonschaufi:
# What this pr does

Add allowed composer plugins

# How to test

Run `composer install` under composer  2.3.4 which should not ask any questions about allowed composer plugins.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,7 +58,11 @@
   },
   "config": {
     "vendor-dir": ".Build/vendor",
-    "bin-dir": ".Build/bin"
+    "bin-dir": ".Build/bin",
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
@simonschaufi 
Thanks for contributing. The change is already done in release-11.5.x.

https://github.com/TYPO3-Solr/ext-solr/blob/ec66f49e50895331efe5a01561d92d83adc47e02/composer.json#L67

Please do not create currently any pull-requests against main branch(but do against release-11.5.x), because we will merge release-11.5.x branch into main soonish.

Issue #3274, opened 2022-07-04
[BUG] Memory leak/deadlock on missing PAGE object


saitho:
**Describe the bug**
This is a special case that regularly occurs during our local development.
We regularly switch between different websites, resulting in a missing PAGE object until Composer and Database are synced.
When this happens and Solr indexer runs via cronjob, the task will use up all available RAM.

Looking into this, it happens when a language in site configuration does not have a root page translation; and a 404 handler is defined. So it's some sort of edge case here, but it still should not happen.

**To Reproduce**
1. Clear page object in TypoScript setup: `page >`
2. Add a new language in site configuration but no root page translation in backend
3. Add a 404 error handler
```yaml
errorHandling:
  -
    errorCode: '404'
    errorHandler: Page
    errorContentSource: 't3://page?uid=84'
```
4. Watch your RAM usage and run the indexer command (or queue initialize command from `Ext:solrconsole`).
5. The php process' RAM usage increases until it reaches 100%.
This may make a system unusable until the process finishes or is killed automatically.

**Expected behavior**
The expected behaviour is that missing `page` object in frontend does not increase the RAM usage unnecessarily.

**Used versions:**
 - TYPO3 Version: 11.5.8
 - EXT:solr Version: 11.5.0-rc-2
 - Used Apache Solr Version: 8.11.1
 - PHP Version: 7.4.29
 - MySQL Version: MariaDB v10.4.24

dkd-kaehm:
@saitho 
I deactivated currently EXT:solr on solr-ddev-site. The frontend gets time-outs:

Please see if TYPO3 Core has some issues with deadlock in 404 Handlings, if none create a issue for core. 
```
Error handler could not fetch error page "https://solr-ddev-site.ddev.site/ru/404", reason: cURL error 28: Operation timed out after 30001 milliseconds with 0 bytes received (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://solr-ddev-site.ddev.site/ru/404
```
IMHO: That shouldn't be fixed in that way. The check should be done before trying initializing of TSFE and setting it to null in cache probably on this place:

https://github.com/TYPO3-Solr/ext-solr/blob/3ea33b4f8387f77d4a34670d0aa7bf02bdf57054/Classes/FrontendEnvironment/Tsfe.php#L315-L325

dkd-kaehm:
[...]

PR #3275, opened 2022-07-04
[BUGFIX] Avoid 404 handler deadlock in special case

saitho:
# What this pr does

The error occurs in a special case when:
- there is no "page" object in TypoScript
- a 404 handler is configured in site configuration
- there is a language without translated root page

In this case, the 404 handler will be stuck in a deadlock, increasing
the RAM usage until it uses all of the available RAM.

This deadlock happens, when the tsfeCache is null. Therefore we ensure
TSFE is set in cache when internal server errors occurred.

# How to test

See #3274. With this fix, the error should not happen anymore.

Fixes: #3274

### DIFF
--- a/Classes/FrontendEnvironment/Tsfe.php
+++ b/Classes/FrontendEnvironment/Tsfe.php
@@ -30,6 +30,7 @@ use TYPO3\CMS\Core\Context\UserAspect;
 use TYPO3\CMS\Core\Context\VisibilityAspect;
 use TYPO3\CMS\Core\Core\SystemEnvironmentBuilder;
 use TYPO3\CMS\Core\Domain\Repository\PageRepository;
+use TYPO3\CMS\Core\Error\Http\InternalServerErrorException;
 use TYPO3\CMS\Core\Exception\SiteNotFoundException;
 use TYPO3\CMS\Core\Http\ServerRequest;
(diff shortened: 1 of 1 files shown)


dkd-kaehm (changes requested):
Hey @saitho,
could you please check if following change instead of your patch will fix your troubles?

https://github.com/TYPO3-Solr/ext-solr/blob/3ea33b4f8387f77d4a34670d0aa7bf02bdf57054/Classes/FrontendEnvironment/Tsfe.php#L130

to

`if (!array_key_exists($cacheIdentifier, $this->tsfeCache)) { `

It seems to be wrong condition for checking the cache hits on nullable values....

saitho:
> Hey @saitho, could you please check if following change instead of your patch will fix your troubles?

@dkd-kaehm Nope, did not work :(

dkd-kaehm:
For instance, the fix was done in CORE See: https://github.com/typo3/typo3/commit/fc51ccbf2b.
We want to test if that fix working on EXT:solr.

dkd-kaehm:
@saitho could you please check if the error is gone with TYPO3 https://github.com/TYPO3/typo3/commit/fc51ccbf2b + ?

saitho:
Can't confirm this issue with the latest TYPO3 version.

Issue #2655, opened 2020-06-17
[BUG] Bad suggest performance with TYPO3 10


timohund:
**Describe the bug**

The performance of the autosuggest on the TYPO3 10 system feels currently very bad (10sec on the dev system). We should analyze that also in combination with TYPO3 9 and check why it is much slower then before.

**To Reproduce**
1. Configure the autosuggest with the default template (Search (Example) - Suggest/autocomplete)

**Expected behavior**
Autosuggest should respond in milliseconds

markuskobligk:
Cannot reproduce this behaviour, the autosuggest is very fast for me on TYPO3 10.4.5 and EXT:solr 11.0.2, even though the Solr server runs on an external machine.

dkd-kaehm:
I think, this is on OsXs ddev environment only.

markuskobligk:
Ok, I can't test this, my computer ain't that fruity :)

dkd-kaehm:
Thanks for confirming Markus, i'll close this issue.

PR #3294, opened 2022-07-24
[BUGFIX] slow response when allowedSites = __all

### LINKED ISSUE #3292: [BUG] very slow response when allowedSites = __all
**Describe the bug**
Using the setting `search.query.allowedSites = __all` in a multi-site setup with many sites makes the search response and the auto-suggest very, very slow. I have 1 TYPO3 with 33 sites, and response times are between 5 and 15 seconds when using '__all'. If I set the value to all 33 domains, the response time < 1s.

I have debugged the issue and found that generating the list of domains is what takes up way too much time, I measured 5 to 12 seconds for these 33 sites. Additional problem is that this is done for every request, including every AJAX request for the auto-suggest feature.

Looking into the code (SiteHashService, SiteRepository), I found that an enormous amount of work is done of which in the end only the  list of domains is used. What is even worse is that the result gets cached, but in a runtime cache which is empty at the next request, and because this method is only called once for each request, this caching only generates more overhead.

The core SiteFinder class can produce the same list of domains in less than 1 ms. Unless I missed something crucial that's being done in the complex current process I would propose to replace the method in SiteHashService (and further clean up that class):

```php
<?php

        ....

	protected function getDomainListOfAllSites(): string
	{
		$siteFinder = GeneralUtility::makeInstance(TYPO3\CMS\Core\Site\SiteFinder::class);
		$sites = $siteFinder->getAllSites();
		$domains = [];
		foreach ($sites as $site) {
			$domains[] = $site->getBase()->getHost();
		}

		return implode(',', $domains);
	}
```

**To Reproduce**
1. Create a setup with many sites on different domains or subdomains
2. set  `plugin.tx_solr.search.query.allowedSites = __all`
3. check how long it takes to for the search page or the auto-suggest to respond
4. now try the same with all domains explicitly listed in allowedSites
You'll find a remarkable speed improvement with step 4. 

**Expected behavior**
Using '__all' should not slow down the search or auto-suggest.

 - TYPO3 Version: [11.5.13]
 - EXT:solr Version: [11.5.0-rc-2]
 - Used Apache Solr Version: [8.11.1]
 - PHP Version: [7.4.30]
 - MariaDB Version: [10.3]

**Possibly related to:**
 #2655

netcoop:
# What this pr does

Fixes very slow response when search.query.allowedSites = __all, esp. on multi-site installations.

# How to test

Add a couple of extra sites with their own domain or subdomain and measure the time it takes for the method getDomainListOfAllSites to complete. Without the patch, this could easily run up to 0.3 or 0.5 seconds for each domain. With this fix, the time is negligible (and so it should, because this method is process uncached).

Fixes: #3292

### DIFF
--- a/Classes/Domain/Site/SiteHashService.php
+++ b/Classes/Domain/Site/SiteHashService.php
@@ -19,6 +19,7 @@ namespace ApacheSolrForTypo3\Solr\Domain\Site;
 
 use Doctrine\DBAL\Driver\Exception as DBALDriverException;
 use Throwable;
+use TYPO3\CMS\Core\Site\SiteFinder;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 
 /**
@@ -77,20 +78,19 @@ class SiteHashService
     }
 
     /**
-     * Returns a comma separated list of all domains from all sites.
+     * Returns a comma separated list with domains of all sites.
      *
      * @return string
-     * @throws DBALDriverException
      * @throws Throwable
      */
     protected function getDomainListOfAllSites(): string
     {
-        $sites = $this->getAvailableSites();
+        $siteFinder = GeneralUtility::makeInstance(SiteFinder::class);
+        $sites = $siteFinder->getAllSites();
         $domains = [];
         foreach ($sites as $site) {
-            $domains[] = $site->getDomain();
+            $domains[] = $site->getBase()->getHost();
         }
-
         return implode(',', $domains);
     }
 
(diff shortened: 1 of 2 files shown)


dkd-kaehm (changes requested):
@netcoop 
Thanks for fixing that and sorry for late response.
IMHO the single usage of `getDomainListOfAllSites()` method is on search context, where TypoScript and other stuff from all sites is not required.
It requires just the domain of each site to be able to generate the site-hash for filter query like 
`siteHash:abcde OR siteHash:edcba OR ...` ignoring the configuration state of solr on each site.
Therefore i would suggest to adjust the test case to the new scenario suggested by you.

Besides that, please add following in your commit message:
```
Fixes: #3292
Relates: #2655
`

Issue #3302, opened 2022-08-03
[BUG] Exception: SuggestQuery Argument (string), null given


josefglatz:
In TYPO3 v11.5 following errors occurs:

```php
Core: Exception handler (WEB): Uncaught TYPO3 Exception: trim(): Argument #1 ($string) must be of type string, null given
TypeError thrown in file .../apache-solr-for-typo3/solr/Classes/Domain/Search/Query/SuggestQuery.php in line 61.
Requested URL: https://www.domain.tld/suche/suggest.json?tx_solr[callback]=jQuery36000680741456213087_1659453754080&tx_solr%5BqueryString%5D=%20&_=1659453754081

Core: Error handler (FE): PHP Warning: Undefined array key 2 in .../apache-solr-for-typo3/solr/Classes/Domain/Search/Query/SuggestQuery.php line 61
```

**Expected behavior**
No matter what search is done: the PHP class must be type safe in that situation. Of course, some of the logic can be handled within Javascript. Mayber there is also something to change.

**Used versions (please complete the following information):**
 - TYPO3 Version: [e.g. 11.5.lateststable]
 - EXT:solr Version: [e.g. 11.5.0-rc-2]
 - Used Apache Solr Version: [e.g. 8.11.1]
 - PHP Version: [e.g. 8.1.8]
 - MySQL Version: [e.g. 8.0.0]

PR #3303, opened 2022-08-03
Fix empty suggest query triggering a PHP error

garfieldius:
# What this pr does

Sending a suggest query containing only spaces causes PHP 8+ to throw a
TypeError exception because the result of the regular expression will
not have a third entry in its matches array.

Use null-coalescing operator to ensure a string is given to the trim
function in the SuggestQuery constructor.

# How to test

Enter only spaces into the search field (which has the autocomplete/suggest function properly configured)  of a TYPO3 installation that uses PHP 8.0 or newer.

Additionally a test was added: \ApacheSolrForTypo3\Solr\Tests\Unit\Domain\Search\Query\SuggestQueryTest->testSuggestQueryDoesNotErrorOnEmptyKeywords

Fixes: #3302

### DIFF
--- a/Tests/Unit/Domain/Search/Query/SuggestQueryTest.php
+++ b/Tests/Unit/Domain/Search/Query/SuggestQueryTest.php
@@ -19,7 +19,6 @@ use ApacheSolrForTypo3\Solr\Domain\Search\Query\QueryBuilder;
 use ApacheSolrForTypo3\Solr\Domain\Search\Query\SuggestQuery;
 use ApacheSolrForTypo3\Solr\System\Configuration\TypoScriptConfiguration;
 use ApacheSolrForTypo3\Solr\Tests\Unit\UnitTest;
-use TYPO3\CMS\Core\Utility\DebugUtility;
 
 /**
  * Tests the ApacheSolrForTypo3\Solr\SuggestQuery class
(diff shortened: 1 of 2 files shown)


Tests/Unit/Domain/Search/Query/SuggestQueryTest.php
```diff
@@ -19,6 +19,7 @@
 use ApacheSolrForTypo3\Solr\Domain\Search\Query\SuggestQuery;
 use ApacheSolrForTypo3\Solr\System\Configuration\TypoScriptConfiguration;
 use ApacheSolrForTypo3\Solr\Tests\Unit\UnitTest;
+use TYPO3\CMS\Core\Utility\DebugUtility;
```
josefglatz, line 22: @garfieldius intended? looks like an überbleibsel

garfieldius: Simmt. Hab's entfernt.

simonschaufi (approved):

dkd-friedrich (approved):

dkd-friedrich:
Thanks for fixing this, will backport this to 11.5.

simonschaufi:
@dkd-friedrich could you please tag a new version?

dkd-friedrich:
> @dkd-friedrich could you please tag a new version?

@simonschaufi I'll look at a few more minor issues today and think that I can then provide a new RC afterwards. 

A new RC was already planned, as our patches have been adopted into the core in the meantime. I'm currently not sure about a stable release for TYPO3 11.5, but an intermediate step can't hurt.

simonschaufi:
an RC release would also be fine as long as I get the latest updates ;)

PR #3319, opened 2022-08-11
[FEATURE] add logging for failed http requests

### LINKED ISSUE #3318: [FEATURE] Improve logging when the request fails

When there is a problem with the server response (like a 401/403/404), it is impossible to know what happened. The only exception in logs and queue module is not very helpful to determine the cause of the problem:

```
Failed indexing Index Queue item 2
"code":1319116885,"message":"Failed to execute Page Indexer Request. See log for details. Request ID: 62f514f64701a"

ApacheSolrForTypo3\\Solr\\IndexQueue\\PageIndexerRequest->getUrlAndDecodeResponse('https://swt11.d...', Object(ApacheSolrForTypo3\\Solr\\IndexQueue\\PageIndexerResponse))
ApacheSolrForTypo3\\Solr\\IndexQueue\\PageIndexerRequest->send('https://swt11.d...')
ApacheSolrForTypo3\\Solr\\IndexQueue\\PageIndexer->getAccessGroupsFromContent(Object(ApacheSolrForTypo3\\Solr\\IndexQueue\\Item), 0)
ApacheSolrForTypo3\\Solr\\IndexQueue\\PageIndexer->index(Object(ApacheSolrForTypo3\\Solr\\IndexQueue\\Item))
```

So this is just some generic "failed". Not useful at all. This happens because `PageIndexerRequest::getUrl()` masks the actual http error and exception.

It would be good to add a new message to logs to list the url, exception data and request parameters. It makes sense to set its log level to `INFO` to make sure that it is logged for `Testing` and `Development` contexts by default.

**Target versions**

solr 11.5.x

dmitryd:
# What this pr does

This PR adds logging for failed site requests. This helps to detect what exactly happened. Messages will appear in the log file for Testing and Development contexts by default. Preview password will be masked in logs.

# How to test

Test by placing password protection in front of the site and starting indexing without user name and password. You will get a vague error in the queue module but a good description of the problem in the log file.

We had a more complicated case, which cost us several lost hours due to missing logging of errors: the site was password protected and we supplied the password. But it still failed to index. When we added logging, we found that the site was accessed using proxy, which we had on this host and proxy did not care to pass on http authentication. It took a lot of time to understand this simple cause. If we had this logging, it could take minutes.

Fixes: #3318

### DIFF
--- a/Classes/IndexQueue/PageIndexerRequest.php
+++ b/Classes/IndexQueue/PageIndexerRequest.php
@@ -23,6 +23,7 @@ use Exception;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Exception\ServerException;
 use Psr\Http\Message\ResponseInterface;
+use Psr\Log\LogLevel;
 use RuntimeException;
 use TYPO3\CMS\Core\Http\RequestFactory;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
Thanks for the fix.
This will be included in final release.

PR #3328, opened 2022-08-24
[BUGFIX] fix infinite loop in Tsfe::getPidToUseForTsfeInitialization()

### LINKED ISSUE #3306: [BUG] Infinite loop/recursion in Indexer if root page is a sysfolder having a ts template
**Describe the bug**
\ApacheSolrForTypo3\Solr\FrontendEnvironment\Tsfe -> getPidToUseForTsfeInitialization() runs into an infinite recursion loop if the root page is a sysfolder with a ts template, resulting in a Fatal Error due to exceeded memory limit.

**To Reproduce**
Consider the following site structure:

[Root]
- [FOLDER] General storage folder (containing a TS template)
- - [FOLDER] News
- [PAGE] Home (containing a TS template)

News records (or other records, like contacts, events, ...) are stored in a subfolder under the "General storage folder". 
The default TS indexing configuration for news is used.

In this setup, the indexer runs into an infinite loop (recursion) in:
\ApacheSolrForTypo3\Solr\FrontendEnvironment\Tsfe -> getPidToUseForTsfeInitialization()

**Expected behavior**
The recursion should be detected and handled with a meaningful exception message.

This can be done by checking $pidToUse and $rootPageId before making the recursive call in line 380.

```
protected function getPidToUseForTsfeInitialization(int $pidToUse, ?int $rootPageId = null): ?int
{
    $incomingPidToUse = $pidToUse;
    $incomingRootPageId = $rootPageId;
    [...]
    if ($pidToUse == $incomingPidToUse && $rootPageId == $incomingRootPageId) {
       throw new Exception\Exception(
                "Infinite recursion detected while looking for the closest page with active template to page \"$askedPid\" . Please note that the page with active template (usually the root page of the current tree) MUST NOT be a sysfolder.",
                1637339476
        );
    }
    return $this->getPidToUseForTsfeInitialization($pidToUse, $rootPageId);
}
```

 - TYPO3 Version: 11.5.13
 - Browser: any
 - EXT:solr Version: 11.5.0-rc-2
 - Used Apache Solr Version: doesn't matter
 - PHP Version: 7.4
 - MySQL Version: 5.7

BrandSven:
…in Tsfe->getPidToUseForTsfeInitialization. Fixes #3306

# What this pr does

Before calling Tsfe->getPidToUseForTsfeInitialization recursively from itself, checks if the function parameters are different from the current call. If they are not, calling recursively will end in an infinite loop, crashing the application by running into PHPs memory limit.

# How to test

- In TYPO3, create a sysfolder named "General Storage"
- Add a typoscript root template to that sysfolder
- Create a subfolder containing some indexable records, for example "news"
- Add an indexable record (news item)
- Setup SOLR to index the record (include the default news indexing configuration typoscript)
- Build index query for news
- Run indexer

Without the fix, php memory limit will be reached ("Allowed memory size of xyz bytes exhausted").
The fix prevents this and provides a meaningful error message by throwing an exception

Fixes: #3306

### DIFF
--- a/Classes/FrontendEnvironment/Tsfe.php
+++ b/Classes/FrontendEnvironment/Tsfe.php
@@ -355,6 +355,9 @@ class Tsfe implements SingletonInterface
      */
     protected function getPidToUseForTsfeInitialization(int $pidToUse, ?int $rootPageId = null): ?int
     {
+        $incomingPidToUse = $pidToUse;
+        $incomingRootPageId = $rootPageId;
+
         // handle plugin.tx_solr.index.queue.[indexConfig].additionalPageIds
         if (isset($rootPageId) && !$this->isRequestedPageAPartOfRequestedSite($pidToUse)) {
             return $rootPageId;
@@ -377,6 +380,15 @@ class Tsfe implements SingletonInterface
         if (isset($rootPageId)) {
             return $rootPageId;
         }
+
+        // Check for recursion that can happen if the root page is a sysfolder with a typoscript template
+        if ($pidToUse === $incomingPidToUse && $rootPageId === $incomingRootPageId) {
+            throw new Exception\Exception(
+                "Infinite recursion detected while looking for the closest page with active template to page \"$askedPid\" . Please note that the page with active template (usually the root page of the current tree) MUST NOT be a sysfolder.",
+                1637339476
+            );
+        }
+
         return $this->getPidToUseForTsfeInitialization($pidToUse, $rootPageId);
     }
 


Classes/FrontendEnvironment/Tsfe.php
```diff
@@ -377,6 +380,15 @@ protected function getPidToUseForTsfeInitialization(int $pidToUse, ?int $rootPag
         if (isset($rootPageId)) {
             return $rootPageId;
         }
+
+        // Check for recursion that can happen if the root page is a sysfolder with a typoscript template
+        if ($pidToUse == $incomingPidToUse && $rootPageId == $incomingRootPageId) {
```
dkd-kaehm, line 385: @BrandSven 
Please fix this scrutinizer issue by using strict comparison operator.

dkd-kaehm (changes requested):
@BrandSven 
Thanks for your contribution.
This patch requires a little change.
Please fix it.

Classes/FrontendEnvironment/Tsfe.php
BrandSven: Should be fixed now

dkd-kaehm (approved):

Issue #3202, opened 2022-02-14
[BUG] SolrRoutingMiddleware does not work when using route enhancer on root page


EmanuelOster:
**Describe the bug**
I have a PageType route enhancer and map a PageType to the suffix menu.json like so:
```yaml
routeEnhancers:
  PageTypeSuffix:
    type: PageType
    default: ''
    map:
      menu.json: 50003
```
This works as expected for every page but the root page. On the root page the extension loggs
> Could not resolve page by path "menu.json" and language "de"

The problem seems to be revolving around how `retrievePageInformation` splits the URL and then wrongly comes to the conclusion that the page does not exist.

**To Reproduce**
1. Define a PageTypeSuffix route enhancer and use it on the root page

**Expected behavior**
The Middleware should detect the root page in the middleware and not log an error

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.24
 - Browser: N/A
 - EXT:solr Version: 11.1.3 (however code is the same on master)
 - Used Apache Solr Version: N/A
 - PHP Version: 7.4
 - MySQL Version: N/A

dkd-kaehm:
@EmanuelOster 
Thanks for reporting that issue.
Could you provide a pull-request or point to the exact trouble-place in code? 
The `SolrRoutingMiddleware::retrievePageInformation()` is about 70 rows, which of them makes the trouble?

jagodzinski:
Hello @dkd-kaehm, 

the trouble-place is 
Line 330
} elseif (empty($path)) { 

If you call a page like demo.org/de/robots.txt or demo.org/de/sitemap.xml, the path is for the first loop /robots.txt and for the second empty, because it's the rootpage. This should not log an error.

PR #3339, opened 2022-08-31
[BUGFIX] Shortcircuit work in SolrRoutingMiddleware

christophlehmann:
# What this pr does

Shortcircuit work in SolrRoutingMiddleware like in PageIndexerFinisher: When a request does not have the X-Tx-Solr-Iq header, then it's not interesting for the Solr extension.

Background: Superflous logging 'Could not resolve page by path...' to Solr logfile.

# How to test

- Index a page, see that its indexed correctly
- You don't see superflous logging for non-existing url paths

Fixes: #3202

### DIFF
--- a/Classes/Middleware/SolrRoutingMiddleware.php
+++ b/Classes/Middleware/SolrRoutingMiddleware.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 
 namespace ApacheSolrForTypo3\Solr\Middleware;
 
+use ApacheSolrForTypo3\Solr\IndexQueue\PageIndexerRequest;
 use ApacheSolrForTypo3\Solr\Routing\RoutingService;
 use Psr\Http\Message\ResponseInterface;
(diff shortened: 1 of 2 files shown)


dkd-kaehm (approved):
@christophlehmann 
Thanks for fixing that issue.

We want back-ports.
Is EXT:solr 11.2.x affected as well?

christophlehmann:
Yes, 11.2 ist affected as well. You can cherry-pick it without further changes.

josefglatz:
@christophlehmann The change is already in v 11.2 but the log entries doesn't stop (within TYPO3 10)

dkd-kaehm:
> @christophlehmann The change is already in v 11.2 but the log entries doesn't stop (within TYPO3 10)

@josefglatz 
Please make sure dev-release-11.2.x is loaded. There is no stable 11.2.x release now with this change.

josefglatz:
Thank you



Diese Nachricht wurde von meinem Smartphone versandt

PR #3350, opened 2022-09-22
[BUGFIX] Make API eID script compatible with TYPO3 v11.5

peterkraume:
# What this pr does

With TYPO3 v11.5 the API to create a siteHash doesn't work anymore:

> Uncaught TYPO3 Exception: #1425381442: Invalid target for "EXT:solr/Classes/Eid/Api.php", as it is not callable. | InvalidArgumentException thrown in file /typo3/sysext/core/Classes/Http/Dispatcher.php in line 95. Requested URL: https://www.domain.tld/index.php?eID=tx_solr_api&api=siteHash&domain=otherdomain.tld&apiKey=98edf0f46e468fbc35d21b9215e73cd8c8017a6c

# How to test

Call to https://www.domain.tld/index.php?eID=tx_solr_api&api=siteHash&domain=otherdomain.tld&apiKey=98edf0f46e468fbc35d21b9215e73cd8c8017a6c
should return a valid json encoded response like this:
```
{"sitehash":"d25b497e596326aeba48edac11294668d793481f"}
```


dkd-kaehm (changes requested):
Please refactor `Classes/Eid/SiteHash.php` to the class and method as well.

peterkraume:
What would the benefit of refactoring `Classes/Eid/SiteHash.php` to a class be? Currently it's just included in `Classes/Eid/ApiEid.php`. If someone uses SiteHash.php it'll break if I refactor it. I would rather suggest include the code from the file within `case 'siteHash':` instead of creating a new class.

dkd-kaehm:
> What would the benefit of refactoring `Classes/Eid/SiteHash.php` to a class be? Currently it's just included in `Classes/Eid/ApiEid.php`. If someone uses SiteHash.php it'll break if I refactor it. I would rather suggest include the code from the file within `case 'siteHash':` instead of creating a new class.

I'll switch the target branch to release-11.5.x for this PR without BC, 
the port for main branch must contain the refactoring on that file.

Issue #3262, opened 2022-06-01
[BUG] AnonymizeIP not anonymizing by octets


dev-rke:
**Describe the bug**
AnonymizeIP won't anonymize by octets for IPv4.

**To Reproduce**
Copy applyIpV4Mask method into an interactive PHP shell and run these commands:
```shell
php > echo applyIpV4Mask("192.168.123.5", 2) . PHP_EOL;
192.168.123
php > echo applyIpV4Mask("192.168.123.15", 2) . PHP_EOL;
192.168.123.
php > echo applyIpV4Mask("192.168.123.15", 4) . PHP_EOL;
192.168.12
php > echo applyIpV4Mask("192.168.123.15", 3) . PHP_EOL;
192.168.123
php > echo applyIpV4Mask("192.168.123.153", 3) . PHP_EOL;
192.168.123.
php > echo applyIpV4Mask("192.168.123.153", 1) . PHP_EOL;
192.168.123.15
php > echo applyIpV4Mask("192.168.123.15", 1) . PHP_EOL;
192.168.123.1
php > echo applyIpV4Mask("192.168.123.1", 1) . PHP_EOL;
192.168.123.
```

**Expected behavior**
I'd expect the last octet is set to 0.
Example:
User has the ip `192.168.123.234`. In the log it should be `192.168.123.0`

**Used versions (please complete the following information):**
 - TYPO3 Version: 10 and 11
 - EXT:solr Version: tested with 11.0 and 11.5-rc2

**Recommendation**
I'd recommend to use the IpAnonymizationUtility::anonymizeIp() function of the TYPO3 core to achieve the correct behaviour.
This class is available since TYPO3 v9.2, so it should be available in all current supported TYPO3 versions:
https://github.com/TYPO3/typo3/blob/9.5/typo3/sysext/core/Classes/Utility/IpAnonymizationUtility.php

By the way: thanks for the great extension - i just used 11.5-rc2 with the awesome API in a project.

PR #3354, opened 2022-10-01
[TASK] Use TYPO3 IpAnonymizationUtility

ayacoo:
Fixes: #3262


dkd-kaehm (approved):
@ayacoo thanks for fixing that issue.
LGTM

Issue #3359, opened 2022-10-17
[BUG] PHP 8: Warning in ext_getSetup aborts indexing


saitho:
**Describe the bug**
Fetching nested TypoScript setup via `ApacheSolrForTypo3\Solr\FrontendEnvironment\TypoScript::ext_getSetup(array $theSetup, string $theKey)` results the following warning during indexing via backend:

```
TYPO3\CMS\Core\Error\Exception: Warning: Undefined array key "tx_myextension" in /var/www/html/public/typo3conf/ext/solr/Classes/FrontendEnvironment/TypoScript.php line 129 in /var/www/html/vendor/helhum/typo3-console/Classes/Console/Error/ErrorHandler.php:88
Stack trace:
#0 /var/www/html/public/typo3conf/ext/solr/Classes/FrontendEnvironment/TypoScript.php(129): Helhum\Typo3Console\Error\ErrorHandler->handleError(2, 'Undefined array...', '/var/www/html/p...', 129)
#1 /var/www/html/public/typo3conf/ext/solr/Classes/FrontendEnvironment/TypoScript.php(127): ApacheSolrForTypo3\Solr\FrontendEnvironment\TypoScript->ext_getSetup(Array, 'tx_myextension')
#2 /var/www/html/public/typo3conf/ext/solr/Classes/FrontendEnvironment/TypoScript.php(113): ApacheSolrForTypo3\Solr\FrontendEnvironment\TypoScript->ext_getSetup(Array, 'plugin.tx_myext...')
#3 /var/www/html/public/typo3conf/ext/solr/Classes/FrontendEnvironment/TypoScript.php(85): ApacheSolrForTypo3\Solr\FrontendEnvironment\TypoScript->buildConfigurationArray(1, 'plugin.tx_myext...', 0)
#4 /var/www/html/public/typo3conf/ext/solr/Classes/FrontendEnvironment.php(52): ApacheSolrForTypo3\Solr\FrontendEnvironment\TypoScript->getConfigurationFromPageId(1, 'plugin.tx_myext...', 0, NULL)
#5 /var/www/html/public/typo3conf/ext/tx_myextension/Classes/Helper/SolrHelper.php(253): ApacheSolrForTypo3\Solr\FrontendEnvironment->getConfigurationFromPageId(1, 'plugin.tx_myext...', 0)
```

This warning then results in an error during indexing.

**To Reproduce**

TypoScript setup:
```typoscript
plugin.tx_myextension {
[...]

PR #3360, opened 2022-10-17
[BUGFIX] Fix array key access in ext_getSetup

saitho:
# What this pr does

Reading nested TypoScript will no longer log a warning when the value can not be accessed. In this case, the second value in the array returned in `TypoScript::ext_getSetup` will be an empty string. However, the second value is not used by Ext:solr anymore.

This also fixes an issue with PHP 8 where such warning would cause the indexing to fail, when the method is used inside of an indexer.

# How to test

Set in TypoScript setup:
```typoscript
plugin.tx_myextension {
  settings {
    foo = bar
  }
}
```

Place this snipped inside a Solr indexer or Solr indexer hook:
```php
use ApacheSolrForTypo3\Solr\FrontendEnvironment;

// @var $indexItem \ApacheSolrForTypo3\Solr\IndexQueue\Item

$settings =  GeneralUtility::makeInstance(FrontendEnvironment::class)
  ->getConfigurationFromPageId($indexItem->getRootPageUid(), 'plugin.tx_myextension', 0);
$config = $settings->getObjectByPathOrDefault('settings', []);
// $config should return the array ['foo' => 'bar']
```

Indexing should no longer produce a PHP warning and it should succeed in PHP 8.

Fixes: #3359

### DIFF
--- a/Classes/FrontendEnvironment/TypoScript.php
+++ b/Classes/FrontendEnvironment/TypoScript.php
@@ -115,18 +115,23 @@ class TypoScript implements SingletonInterface
     }
 
     /**
+     * Adapted from TYPO3 core
+     * @see sysext:core/Classes/TypoScript/ExtendedTemplateService until TYPO3 v11
      * @param array $theSetup
      * @param string $theKey
      * @return array
      */
     public function ext_getSetup(array $theSetup, string $theKey): array
     {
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
@saitho 
Thanks for that fix.
It will be merged after 11.5 port is created.

PR #3361, opened 2022-10-17
[BUGFIX] Fix array key access in ext_getSetup (Backport 11.5)

saitho:
Reading nested TypoScript will no longer log a warning when the value can not be accessed. In this case, the second value in the array returned in `TypoScript::ext_getSetup` will be an empty string. However, the second value is not used by Ext:solr anymore.

This also fixes an issue with PHP 8 where such warning would cause the indexing to fail, when the method is used inside of an indexer.

Fixes: #3359
Ports: #3360


dkd-kaehm (approved):
:+1:

PR #3363, opened 2022-10-19
[TASK] Do not index language with unconfigured core

### LINKED ISSUE #3362: [BUG] Fallback of solr_core_read does not work
**Describe the bug**

This is about an empty `solr_core_read` in a site language.

As per default (last parameter in line 293 = true) a site language will be indexed, whether or not `solr_core_read` is empty.

Bildschirmfoto 2022-10-18 um 11 50 53

In Line 306 `core_en` should be the default (fallback) core, if non given (=null value). But i couldn't be null, since its a string field and is might be just empty. 

An empty string leads to a wrong `path` and indexing fails.

**Expected behavior**

It could be more explizit: When a core is given and not empty, then index that language, else not. 

We can see it also as a feature, so that one can deactivate indexing a language with an empty `solr_core_read` property.

 - TYPO3 Version:  ^11
 - EXT:solr Version: ^11

christophlehmann:
# What this pr does

It disables indexing a language that has no core configured.

Further there are no connection errors in the backend module nor pages with indexing errors due to that language

# How to test

Add a site language with an empty core

Fixes #3362

### DIFF
--- a/Classes/Domain/Site/SiteRepository.php
+++ b/Classes/Domain/Site/SiteRepository.php
@@ -291,7 +291,9 @@ class SiteRepository
 
         foreach ($availableLanguageIds as $languageUid) {
             $solrEnabled = SiteUtility::getConnectionProperty($typo3Site, 'enabled', $languageUid, 'read', true);
-            if ($solrEnabled) {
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
LGTM

Issue #3364, opened 2022-10-20
[TASK] Introduce index queue type setting


dkd-friedrich:
**What should be done in the scope of this task?**

Queue types shouldn't be restricted to internal data, which is already prepared in some classes. As index queue setting 'table'
suggests that a local table is expected, this should replaced with a more generic setting 'type'.

PR #3365, opened 2022-10-20
[TASK] Introduce index queue type setting

dkd-friedrich:
# What this pr does

Queue types shouldn't be restricted to internal data, which is already prepared in some classes. As index queue setting 'table' suggests that a local table is expected, this is replaced with a new setting 'type'.

`queue.[indexConfig].table` is now deprecated and will be removed in v13, 'type' should be used instead.

# How to test

Configure an indexing configuration and configure the local table via `type` instead of `table`

Fixes: #3364


RafaelKa (approved):
LGTM

dkd-kaehm (approved):

Issue #3366, opened 2022-10-20
[BUG] PHP 8: Indexing missing fields result in warning aborting the indexing


saitho:
**Describe the bug**
Given a custom indexer that sets data conditionally and a Solr field mapping via TypoScript.
Ext:solr's `AbstractIndexer` will access the array value, which results in a PHP warning when the data does not exist.
PHP warnings then result in abortion of the indexing.

**To Reproduce**

```
plugin.tx_solr {
  index {
    queue {
      tx_myext_domain_model_entity = 1
      tx_myext_domain_model_entity {
        table = tx_myext_domain_model_entity
        indexer = Vendor\Myext\IndexQueue\EntityIndexer

        fields {
          myfield_stringS = myField
        }
      }
    }
```

```php
use ApacheSolrForTypo3\Solr\IndexQueue\Indexer;

class EntityIndexer extends Indexer
{
    /**
     * {@inheritDoc}
     */
    protected function getFullItemRecord(Item $item, int $language = 0): ?array
    {
        $record = parent::getFullItemRecord($item, $language);
        if (empty($record)) {
            return null;
        }

        if ($record['uid']%2 === 0) { // conditional field. as example myField is only set for even ids
            $record['myField'] = 'Some text';
        }
        return $record;
    }
```

**Expected behavior**

Ext:solr should not try to index fields where the data is not available.

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.16
 - EXT:solr Version: 11.5.0
 - PHP Version: 8.1.8

PR #3367, opened 2022-10-20
[BUGFIX] Do not index missing fields

saitho:
In case the data used for a Solr field is not present in the data (e.g. if it is conditionally set via a custom indexer), the field will no longer be created.

Previously the data was accessed anyways, resulting in a PHP warning which would abort the indexing in PHP 8.

Fixes: #3366

### DIFF
--- a/Classes/IndexQueue/AbstractIndexer.php
+++ b/Classes/IndexQueue/AbstractIndexer.php
@@ -85,6 +85,9 @@ abstract class AbstractIndexer
             }
 
             $fieldValue = $this->resolveFieldValue($indexingConfiguration, $solrFieldName, $data, $tsfe);
+            if ($fieldValue === null) {
+                continue;
+            }
 
             if (is_array($fieldValue)) {
                 // multi value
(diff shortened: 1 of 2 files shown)


Classes/IndexQueue/AbstractIndexer.php
```diff
@@ -187,6 +190,9 @@ protected function resolveFieldValue(
                 $fieldValue = unserialize($fieldValue);
             }
         } else {
+            if (!array_key_exists($indexingConfiguration[$solrFieldName], $data)) {
```
dkd-friedrich, line 193: If think we should also check `$indexingConfiguration[$solrFieldName]` , a non-empty string is required

E.g.:

```
if (is_string($indexingConfiguration[$solrFieldName] ?? null)  && !array_key_exists($indexingConfiguration[$solrFieldName], $data))
```

dkd-friedrich (changes requested):
I think we should improve the check in resolveFieldValue

Linter is also failing, `composer install && composer exec php-cs-fixer fix` should this.

dkd-kaehm (approved):
:+1:

dkd-friedrich (approved):

Issue #3368, opened 2022-10-20
[TASK] Implement an indexing configuration icon fallback


dkd-friedrich:
**What should be done in the scope of this task?**

The Index queue module uses the registered record icon for each indexing configuration, if no valid icon is found or no local table
is used, the default fallback icon is used and indicates an error.  We should use a default icon that doesn't imply an error.

PR #3369, opened 2022-10-20
[TASK] Indexing configuration icon fallback

dkd-friedrich:
# What this pr does

Index queue module uses the registered record icon for each indexing configuration, if no valid icon is found or no local table is used, the default fallback icon is used and indicates an error. This commit implements an icon check and a default icon that doesn't imply an error.

# How to test
Configure an indexing configuration that uses a indexing source that doesn't configure an icon. With this changeset a default icon (`mimetypes-other-other`) is used.

Resolves: #3368


dkd-kaehm (approved):
LGTM

Issue #3376, opened 2022-10-24
Epic [TASK] Make EXT:solr TYPO3 12 LTS compatible


3l73:
This task contains all the sub task to get EXT:solr woking within TYPO3 12 LTS:
**Note:** The compatibility to TYPO3 11 LTS MUST NOT be assured.

### Must have

- [x] #3382
- [x] Reports: Migrate configuration from ext_tables to Service.yaml
- [x] [TASK:T12] Solarium: Upgrade to version 6.2.7 (minimal requirement 6.2.4 due dependencies)
- [x] #3476
- [x] #3473
- [x] #3470
- [x] #3514
- [x] #3470
- [x] #3452 
- [x] #3454 
- [x] #3456 
- [x] #3207 
- [x] #3437
- [x] #3460
- [x] #3490
- [x] #3529
- [x] #3564
- [x] #3614
- [x] #3511
- [x] [TASK:T12] ViewHelper: Fix inheritance of final ViewHelper issue
- [x] #3513
- [x] #3619
- [x] #3632
- [x] #3653
- [x] #2775
- [x] #3567
- [x] #3488
- [x] #3580
- [x] #3674
- [x] #3676
- [x] #3677
- [x] #3666
- [x] #3681
- [x] #3631
- [x] #3679
- [x] #3685
- [x] #3629
- [x] #3698
- [x] #3747
- [x] #3725
- [x] #3691
- [x] #3489
- [x] #3800
- [x] #3798
- [x] #3806
- [x] #3043 
  - [x] #3518
  - [x] #3665 
  - [x] #3630
- [x] #3735
- [x] #3810
- [x] #3823
- [x] #3703
- [x] #3722
- [ ] #3737

### Deprecations

- [x] #3471
- [x] #3499
- [x] #3506

### Tasks after merge TYPO3 12 compatibility into main (Pull-Request: #3377)

- [x] Merge #3377 into main branch
- [x] #3661
- [x] Rebase PRs and change target to main branch: 
   - [x] #3585
   - [x] #3587
   - [x] #3516
   - [x] #3649

### Nice to have

- [x] #3502 
   via #3503 
- [ ] #3501

# Final 12.0.0

- [x] #3829

---

#### 2023.03.24

The indexing stack almost working now with following restriction:
* TypoScrip templates on subpages are not loaded, but only from the root page is used. This will be fixed near alpha state.

Next steps:

* Fix the search context
* Fix small issues in BE-Modules 

N/A

PR #3378, opened 2022-10-24
[TASK] Remove dependencies to Core testing classes

saitho:
Classes from TYPO3 core testing namespace are not available
when the package is installed via Composer.

This causes local tests to fail.
Therefore the classes from TYPO3 core are copied into Ext:solr.

Fixes: #3376

### DIFF
--- a/Tests/Integration/Fixtures/Frontend/PhpError.php
+++ b/Tests/Integration/Fixtures/Frontend/PhpError.php
@@ -1,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+/*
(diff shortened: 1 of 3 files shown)


dkd-kaehm:
No needs to copy:
See the change in
composer.json
from 
```
"TYPO3\\CMS\\Core\\Tests\\": ".Build/Web/typo3/sysext/core/Tests/"
```
to
```
"TYPO3\\CMS\\Core\\Tests\\": ".Build/Web/vendor/typo3/cms-core/Tests/"
```
in #3384

Issue #3375, opened 2022-10-21
[TASK] Disable all TSFE hacks in Search context


dkd-kaehm:
**What should be done in the scope of this task?**

One of the simplest ways to solve the issues like #3065 within search could be a disabling all the hacks belonging to TSFE side effects done by EXT:solr.

The method `\ApacheSolrForTypo3\Solr\Controller\AbstractBaseController::initializeAction()` must set the EXT:solr in Search context, by setting static class property(or similar patterns).
All the TSFE hacks methods must skip the execution if EXT:solr is in Search context.

- [ ] #3065

Target Versions:
* 11.2.x
* 11.5.x
* main

PR #3385, opened 2022-11-01
[TASK] #3375: prepend wrong fe language on empty cache,

goldi42:
# What this pr does

Prepend wrong FE language on empty cache,
use a clone of context on indexing to encapsulate the index from cli process

# How to test

Clear all caches and call the frontend of a site with nondefault language

Fixes: #3375


dkd-kaehm (approved):
@goldi42 
Thanks for fixing that issue.

Issue #3387, opened 2022-11-03
[FEATURE] Add signal before search in SearchController::resultsAction


stat1x:
**Description**
It may be interesting to have a signal before the search is done to be able to modify the arguments.

**Solution**
The solution is quiet easy to implement.
It can be done like this :
```php
    public function resultsAction()
    {
        try {
            $arguments = (array)$this->request->getArguments();
            $pageId = $this->typoScriptFrontendController->getRequestedId();
            $languageId = Util::getLanguageUid();

            // New signal needed
            $arguments = $this->emitActionSignal(__CLASS__, 'beforeSearch', [$arguments]);

            $searchRequest = $this->getSearchRequestBuilder()->buildForSearch($arguments, $pageId, $languageId);
            $searchResultSet = $this->searchService->search($searchRequest);
            ...
        } catch (SolrUnavailableException $e) {
            $this->handleSolrUnavailable();
        }
    }
```

**Target versions**
10 and 11

dkd-kaehm:
@stat1x 
Thanks for the feature.
Signals are deprecated.
Please make the pull requests against release-11.2.x and release-11.5.x, not against main branch.

Note: Currently only branches release-11.2.x+ are supported. Also:
* release-11.2.x 
* release-11.5.x 
* main

dkd-kaehm:
@stat1x
on main branch we want PSR-14 on this place.

dkd-kaehm:
Events tracked on: #3437

PR #3389, opened 2022-11-03
[FEATURE] Add signal before search in resultsAction

stat1x:
What this pr does
Add a signal before search in resultsAction to be able to modify the arguments.

How to test
Check that the signal is correctly emitted by connecting to it from another extension.

Fixes: #3387


dkd-kaehm:
### Note: 

release-10.0.x is not supported, therefore no release will be tagged anymore.

Issue #3423, opened 2022-12-15
[BUG] Solr plugins affect frontend environment


dkd-friedrich:
**Describe the bug**
Search form and search results plugins are initializing Solr sites to get the right Solr connection. During initialization own `TypoScriptFrontendController` instances are created and `TypoScriptFrontendController` automatically sets the language in the global `PageRenderer`

Due to this manipulation the frontend language might be affected and content rendered after the Solr plugins might be in the wrong language.

A similar issues was fixed in #3385 for TYPO3 10 and issue relates to #3375, but is only a part of it.

**To Reproduce**
- Place a Solr plugin 
- Place another Plugin relying on the SiteLanguage 

**Expected behavior**
Global language is not affected by EXT:solr

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.14+
 - EXT:solr Version: 11.5
 - PHP Version: 8.1

PR #3425, opened 2022-12-16
[BUGFIX:BP:11.5] Fix frontend Solr connection initialization

dkd-friedrich:
# What this pr does

Solr plugins are initializing Solr connections for all available
languages, this causes wrong language initializations in the
PageRenderer, as all created TypoScriptFrontendController instances
will overwrite the language setting.

By allowing to fetch the Solr configuration by TYPO3 site and a
specific language this issue is solved and the initialization
overhead is reduced.

# How to test

- Place a Solr plugin
- Place another Plugin relying on the SiteLanguage

Fixes: #3423


dkd-kaehm (approved):
LGTM

Issue #3413, opened 2022-11-25
[BUG] PHP Warning Trying to access array offset on value of type null in Util.php


froemken:
**Describe the bug**

PHP Warning: Trying to access array offset on value of type null in /var/www/html/public/typo3conf/ext/solr/Classes/Util.php line 119

**To Reproduce**
Delete a record which is not referenced to an existing col_pos. Happen when you switch backend layouts.

**Expected behavior**
Check value of BackendUtility::getRecord before accessing it as array

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.19
 - Browser: Chrome
 - EXT:solr Version: 11.5.0
 - Used Apache Solr Version: 8.11.1
 - PHP Version: 8.1.11
 - MySQL Version: 10.2.44-MariaDB-1:10.2.44+maria~bionic-log - mariadb.org binary distribution

dkd-kaehm:
@froemken 
Thanks for reporting that issue.
I can not reproduce this issue currently within solr-ddev-site
ON:
* TYPO3 11.5.19 + 21
* PHP 8.1.8 
* EXT:solr 11.5.0, [release-11.5.x@01d568b](https://github.com/TYPO3-Solr/ext-solr/commit/01d568ba44557f430086639ae5bd3b9d1809af75)

Could you please check if #3427 resolves this issue?

froemken:
Hi @dkd-kaehm 

yes, this will solve the issue, too.

The return value of BU::getRecord() is array|null that's why I thought you will add a test for array instead, before accessing the array value one line below.

Stefan

PR #3427, opened 2022-12-20
[BUGFIX] Handle deleted workspace records

jschlier:
# What this pr does

This pull request handles deleted records when checking if the changed record is part of a workspace.

# How to test

Delete a record with workspaces enabled.

Fixes: #3413


jschlier:
I may not fully understand in which cases this function is used, but I think that deleted records should be checked as well.
A simple check if the $record is `null` before accessing the array keys could also be sufficient. Thinking about it it leads to deleted records from a workspace as being return as "live" records.

Since this only happens when records are deleted it probably does not add any unwanted behaviour - except possibly a  call to Solr to delete a record that is not indexed - but checking the deleted record anyway is cleaner in my opinion.

froemken:
Can someone please restart the tests? It's not possible to see whats wrong with this patch.

dkd-kaehm:
@froemken 
Rebased to current main state, which starts the build.

Classes/Util.php
```diff
@@ -114,7 +114,7 @@ public static function isDraftRecord(string $table, int $uid): bool
         $isWorkspaceRecord = false;
 
         if ((ExtensionManagementUtility::isLoaded('workspaces')) && (BackendUtility::isTableWorkspaceEnabled($table))) {
-            $record = BackendUtility::getRecord($table, $uid, 'pid, t3ver_state');
+            $record = BackendUtility::getRecord($table, $uid, 'pid, t3ver_state', '', false);
 
             if ($record['pid'] == '-1' || $record['t3ver_state'] > 0) {
```
froemken, line 119: Officially getRecord() can return either array or null. This patch works, but I would prefer to check for array first, before accessing the array keys.

froemken (changes requested):

jschlier:
Sorry, I lost track of this PR.

The new version of the main branch already includes a `null` check in the following IF statement.

This could be merged then?

PR #3444, opened 2023-01-02
[BUGFIX] add empty string as fallback

lukasniestroj:
# What this pr does

ensure a string is always returned, even if the query does not exist

# How to test

Define a queryGroup "A", filter by that new group, manipulate the url by hand, submit form. Page should not show error.


dkd-friedrich:
Thanks for reporting and providing a patch.

Ensuring that a decoded string is returned is unfortunately not sufficient. An invalid filter will be applied and Solr will return an error (`400 Bad Request`) that leads to an exception.

I've added a check for invalid filter parts, invalid filters will be ignored and logged.

dkd-friedrich (approved):

Issue #3448, opened 2023-01-09
[BUG] regression in HighlightResultViewHelper


n3amil:
**Describe the bug**
the introduction of strict_types=1 in combination with str_replace within the HighlightResultViewHelper resulted in a regression, if there is a configured highlightFields, say "text", but some result items don't have a field "text" for highlighting. For example a search with multiple item types like news items (having a field "text") and products (not having a field "text").  
Before the regression escapeEverythingExceptAllowedTags would simply return an empty string in that case. Now an Exception is thrown:
Uncaught TYPO3 Exception: str_replace(): Argument 3 ($subject) must be of type array|string, null given

**To Reproduce**
1. configure two index queue items, one with a "text" (or any) field, one without
2. configure the field to be highlighted with highlightFields = text
3. use the viewhelper, pass the fieldname  fieldName:'text'

should work the same for the default field "content" or any field

**Expected behavior**
search results (the partial) are just rendered as before, if highlighting can be applied it is, rest of the items missing the configured highlightFields are just displayed normally. 

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5
 - EXT:solr Version: 11.5.1
 - Used Apache Solr Version: 8.11.1
 - PHP Version: 8.0

**Additional context**
regression was introduced with https://github.com/TYPO3-Solr/ext-solr/commit/61076e3ed99ba2c4b9f17d44c2e50f278218780d

einpraegsam:
I can confirm this issue for our website update from 10 to 11 with ApacheSolrForTypo3/Solr in version 11.5.1.
For now we're going to copy the VH to our sitepackage and do a simple typecast.

PR #3449, opened 2023-01-09
return empty string for renderStatic if there is no content to highlight

n3amil:
# What this pr does

fix regression in HighlightResultViewHelper if there is no content to highlight, by checking if getHighlightedContent is a string before passing it to escapeEverythingExceptAllowedTags, otherwise an empty string is returned for the renderStatic method, as before the regression. 

# How to test
see issue

Fixes: #3448


dkd-friedrich (approved):
LGTM

dkd-friedrich:
@n3amil Thanks for reporting and providing a pr

I'll take care of the backport to 11.5.

Issue #3420, opened 2022-12-07
Missing  Custom Template  when  Extend to Subpages = 1 


drblitz-weblab:
1. For testing you can add this code to your root page

```
plugin.tx_solr.view.templateFiles.results.availableTemplates {
    resultss = Results
    default {
        label = Default Searchresults Template
        file = Results
    }
    finder {
        label = Event
        file = ProductFinder
    }
}
```
And here is working fine 
<img width="978" alt="Zrzut ekranu 2022-12-7 o 10 25 44" src="https://user-images.githubusercontent.com/44512564/206140195-d9f3571c-5e68-4206-b46c-fc7a2b4e473a.png">

2. Create two example pages  and set in page property  Extend to Subpages = 1 
<img width="2017" alt="Zrzut ekranu 2022-12-7 o 10 28 32" src="https://user-images.githubusercontent.com/44512564/206140828-2d9622f6-0e22-4584-9eb9-fe8f8e9d041c.png">

the page tree should look like this

<img width="249" alt="Zrzut ekranu 2022-12-7 o 10 29 37" src="https://user-images.githubusercontent.com/44512564/206141128-0149849f-78f1-4497-bd98-ecbf80b73bf6.png">

3.  Go to page "Sub 2.1" and add solr search 
<img width="1885" alt="Zrzut ekranu 2022-12-7 o 10 30 59" src="https://user-images.githubusercontent.com/44512564/206141453-f6441adc-1f24-4cec-b7cf-8dfe588114ed.png">

**Custom template is missing**

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.19
 - EXT:solr Version: 11.5.0
 - PHP Version: 8.0

micron:
Just ran into this issue too. Any estimate when this will be fixed ?

dkd-kaehm:
@drblitz-weblab @micron 
Thanks for reporting and confirming that issue.
That is very strange behavior. We want to find out the cause of that issue. 
If you find something, please make notice there.

sabbelasichon:
@dkd-kaehm The thing is that the retrieval for the templates uses the FrontendEnvironment class but the page is access restricted so the template configuration cannot be fetched and nothing is found.

sabbelasichon:
@dkd-kaehm I am not sure why the plugin configurations are using the simulated frontend environment. Maybe i am missing something here.

dkd-kaehm:
[...]

PR #3486, opened 2023-02-10
[BUGFIX] Use ConfigurationManager to get typscript in FlexForm

sabbelasichon:
# What this pr does

Do not get TypoScript from simulated frontend environment but from the ConfigurationManager. 

# How to test

Check the description in the issue and you will see the error but with this patch it works

Fixes: #3420

---

- [x] Port/cherry-pick into release-11.5.x
- [x] Port/cherry-pick into release-11.2.x

### DIFF
--- a/Classes/System/UserFunctions/FlexFormUserFunctions.php
+++ b/Classes/System/UserFunctions/FlexFormUserFunctions.php
@@ -18,7 +18,6 @@ declare(strict_types=1);
 namespace ApacheSolrForTypo3\Solr\System\UserFunctions;
 
 use ApacheSolrForTypo3\Solr\ConnectionManager;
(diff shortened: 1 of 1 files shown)


dkd-kaehm (changes requested):
The code is fine, but the commit message misses the `Fixes: #3420` line 
and name/header should be:
`[BUGFIX] Use ConfigurationManager to get typscript in plugin FlexForm`

dkd-kaehm:
- [ ] After the tests are green and commit message is changed, the fix must be back-ported/cherry-picked into the release-11.5.x.

dkd-kaehm:
@drblitz-weblab @micron
There is the patch, please check and test this out.

dkd-kaehm (approved):

Issue #3502, opened 2023-02-19
[FEATURE] Disable cores in docker image easily


christophlehmann:
The docker image comes with a lot of cores. Most of the time they are not all needed and produce superflous log output etc.

It would be nice to have an easy possibility to disable unneeded cores.

PR #3503, opened 2023-02-19
[FEATURE] Introduce TYPO3_SOLR_ENABLED_CORES docker env variable

christophlehmann:
This adds a startup script which disables all unneeded cores.

Usage:

docker run -e 'TYPO3_SOLR_ENABLED_CORES=english german' -it typo3solr/ext-solr

This enables only the english and german core (core_en, core_de) during startup.

If the env variable is not set, then all cores are enabled (= default behaviour).

Resolves: #3502


Docker/SolrServer/docker-entrypoint-initdb.d/disable-cores.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ ! -z "${TYPO3_SOLR_ENABLED_CORES}" ]
+then
+    mv /var/solr/data/cores /var/solr/data/cores_disabled
+    mkdir /var/solr/data/cores
+    for CORE in ${TYPO3_SOLR_ENABLED_CORES}
+    do
+        echo "Enable core ${CORE}"
+        mv /var/solr/data/cores_disabled/${CORE} /var/solr/data/cores/${CORE}
+    done
+    rm -rf /var/solr/data/cores_disabled
```
dkd-kaehm, line 12: This disallows to enable additional cores afterwards.

Docker/SolrServer/docker-entrypoint-initdb.d/disable-cores.sh
```diff
@@ -0,0 +1,13 @@
...
+fi
```
dkd-kaehm, line 13: I would suggest to use the version with possibility to enable additional cores.

```suggestion
#!/usr/bin/env bash

if [ -n "${TYPO3_SOLR_ENABLED_CORES}" ]
then
  [ -d /var/solr/data/cores_disabled ] || mkdir /var/solr/data/cores_disabled
  mv /var/solr/data/cores/* /var/solr/data/cores_disabled/.
  for CORE_TO_DISABLE in /var/solr/data/cores_disabled/*/core.properties
  do
    mv "${CORE_TO_DISABLE}" "${CORE_TO_DISABLE}_disabled";
  done

  for CORE_to_ENABLE in ${TYPO3_SOLR_ENABLED_CORES}
  do
    echo "Enable core ${CORE_to_ENABLE}"
    mv "/var/solr/data/cores_disabled/${CORE_to_ENABLE}" "/var/solr/data/cores/${CORE_to_ENABLE}"
    mv "/var/solr/data/cores/${CORE_to_ENABLE}/core.properties_disabled" "/var/solr/data/cores/${CORE_to_ENABLE}/core.properties"
  done
fi
```

dkd-kaehm (changes requested):
@christophlehmann 
Thanks for this nice feature.
Didn't know about `docker-entrypoint-initdb.d/` possibility.

dkd-kaehm:
The script of solr-ddev-site 

https://github.com/TYPO3-Solr/solr-ddev-site/blob/dd776dad79d3ca405652124bb91e5b5d1fd732be/.ddev/solr/docker-overrides-entrypoint.sh

```bash
overrideSolrUnixUidAndGidAndSetPermissions ()
{
  if [[ -v SOLR_OVERRIDE_UID ]] && grep -E -q '^[0-9]+$' <<<"${SOLR_OVERRIDE_UID:-}"; then
    usermod --non-unique --uid "$SOLR_OVERRIDE_UID" solr
  fi

  if [[ -v SOLR_OVERRIDE_GID ]] && grep -E -q '^[0-9]+$' <<<"${SOLR_OVERRIDE_GID:-}"; then
    groupmod --non-unique --gid "$SOLR_OVERRIDE_GID" solr
  fi

  chown -R solr:solr /var/solr
}

overrideSolrUnixUidAndGidAndSetPermissions
```

can be adapted and moved there as well.

brandung-sjorek:
@christophlehmann In think the [`cibuild_docker.sh`](https://github.com/TYPO3-Solr/ext-solr/blob/main/Build/Test/cibuild_docker.sh) should be updated accordingly; or the github-action workflow which uses this script. 

@dkd-kaehm Right?

dkd-kaehm:
> @christophlehmann In think the [`cibuild_docker.sh`](https://github.com/TYPO3-Solr/ext-solr/blob/main/Build/Test/cibuild_docker.sh) should be updated accordingly; or the github-action workflow which uses this script.
> 
> @dkd-kaehm Right?

The scripts inside of `/docker-entrypoint-initdb.d` are executed automatically, no needs to do something else as provided there.
See: https://github.com/apache/solr-docker/blob/3a61966de4da0ef7a3e9f2d29959d6bcb2ea117e/8.11/scripts/run-initdb

Also not more than the Docker-Image tests could be enriched for this stuff.

brandung-sjorek:
@dkd-kaehm Sorry, maybe I got you wrong. You want (as I suggested) @christophlehmann to implement a test, if disabling cores works as expected, in [`cibuild_docker.sh`](https://github.com/TYPO3-Solr/ext-solr/blob/main/Build/Test/cibuild_docker.sh). Correct?

dkd-kaehm:
> @dkd-kaehm Sorry, maybe I got you wrong. You want (as I suggested) @christophlehmann to implement a test in [`cibuild_docker.sh`](https://github.com/TYPO3-Solr/ext-solr/blob/main/Build/Test/cibuild_docker.sh), to check if disabling cores works as expected. Correct?

Yes, if possible it would be nice to cover it with test cases.
For example:
First start with "de" and "en" and check that they are enabled, 
then remove "de" and add "da" and restart then check if "de" is gone and "da" is there.

---

PS: @brandung-sjorek I thought, your suggestion was for enabling this feature in image at all, not about test scenarios in build process.

PR #3516, opened 2023-02-24
[TASK:T12] DI for IndexQueueWorkerTaskAdditionalFieldProvider::$siteRepository

dkd-kaehm:
Relates: #3376


sfroemkenjw (approved):
Now, that the FQCN are shortened, it may be useful to remove a lot of RETURNS

Issue #3043, opened 2021-09-30
[BUG] `solr_path_read` does not work properly


dkd-kaehm:
**Describe the bug**
There are to many confusions about "/solr/" path segment for connection configurations. 
Current code base is inconsistent and should be refactored.
The Site configuration has already split settings for path and core, so we should propagate them as is to the node initialization. 

We should remove the processing/splitting from:
https://github.com/TYPO3-Solr/ext-solr/blob/c7a07832989e01111b8cf8416624a70d6dde0ef0/Classes/System/Solr/Node.php#L47-L54

and concatenation to "path":
https://github.com/TYPO3-Solr/ext-solr/blob/c7a07832989e01111b8cf8416624a70d6dde0ef0/Classes/Domain/Site/SiteRepository.php#L295-L297

and propagate the path and core as is without any processing.

**Used versions (please complete the following information):**
 - TYPO3 Version: [e.g. 10.4]
 - EXT:solr Version: [e.g. 11.1.x]
 
**Additional context**
Discussion started on https://typo3.slack.com/archives/C02FF05Q4/p1632990610130900
and #3041

biebricher-of:
Hi,

I have maybe a similar problem with `solr_path_read` but in the opposite direction.
I work with a solr-server without a path-element before the core-name.

**Example**:
schema: `https`
host: `solr***.solr-hosting.info`
core: `core_de`
full-path: https://solr***.solr-hosting.info/core_de

1. Naturally I didn't config the `solr_path_read`. But then it will build a URI like `https://solr***.solr-hosting.info/solr/core_de`.
2. Then I configured the `solr_path_read` as `/` ... the trim-escalation ;-) removed it and it ended again with a solr-URI like `https://solr***.solr-hosting.info/solr/core_de`.

When I removed the path handling in the repository:

**Before**:
https://github.com/TYPO3-Solr/ext-solr/blob/c7a07832989e01111b8cf8416624a70d6dde0ef0/Classes/Domain/Site/SiteRepository.php#L295-L297

**After**:
```
'path' => SiteUtility::getConnectionProperty($typo3Site, 'core', $languageUid, 'read', 'core_en') . '/' ,
```
**Then**:
a) it works for me on a server without path (like my example)
b) I can use the core-variable to add a path if needed (`solr_core_read: /solr/core_de`)

**My questions**:
- Is this the way?
- Or didn't I miss a configuration-option to enable server without a path?

Thank you.


Versions:
[...]

PR #3518, opened 2023-02-27
[BUGFIX] Use correct property when creating Node (context has to be s…

stat1x:
…et instead of path)

# What this pr does

Settings the path when creating makes no sense as a 'context' property exists.

# How to test

Log in to TYPO3 backend and check the solr connection status (should be green).

Fixes: #3043


dkd-kaehm:
Replaced by: #3803

Issue #3531, opened 2023-03-03
[BUG] PHP 8 warning in Util::isDraftRecord


dogawaf:
**Describe the bug**

Since an upgrade to php 8, I see this error in the logs:

```
PHP Warning: Trying to access array offset on value of type null in /var/www/html/web/typo3conf/ext/solr/Classes/Util.php line 119
```

**To Reproduce**

The error occurs in a php 8 setup, with TYPO3 11.5, when an editor save a content in the BE.
I was not able to reproduce the issue on my own, but the issue happens several times a day.

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.23
 - EXT:solr Version: [e.g. 11.5.0]
 - PHP Version:  8.0.27

dogawaf:
The fix is maybe not the good one, because it can hide the real issue: why the record was not loaded?.

koehnlein:
I can confirm this error and know how to reproduce it:

It occurs, when a record is deleted. For example a content element.

baschny:
The fix https://github.com/TYPO3-Solr/ext-solr/commit/315ddf397e3f9cd8574288282d27bc49675ee6f4 does not really fix it completely. See https://github.com/TYPO3-Solr/ext-solr/pull/3642 for a more complete fix, this should / could also be backported to the 11 branch.

PR #3532, opened 2023-03-03
[BUGFIX] Ensure record exists before asserting if draft

dogawaf:
# What this pr does

This fix an error in case a record was not found in the garbage collector process.

Fixes: #3531

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -116,7 +116,7 @@ class Util
         if ((ExtensionManagementUtility::isLoaded('workspaces')) && (BackendUtility::isTableWorkspaceEnabled($table))) {
             $record = BackendUtility::getRecord($table, $uid, 'pid, t3ver_state');
 
-            if ($record['pid'] == '-1' || $record['t3ver_state'] > 0) {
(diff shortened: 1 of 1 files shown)


dogawaf:
The fix is maybe not the good one, as it can hide the real issue: why the record was not found?

dkd-kaehm:
> The fix is maybe not the good one, as it can hide the real issue: why the record was not found?

Did you find out why it was so on your system?

koehnlein:
I just added some information how to reproduce it in #3531:

It happens if a record is deleted. `BackendUtility::getRecord()` is executed with the uid of the deleted record and returns `null`

dkd-friedrich:
I'm also not exactly sure why an already deleted record gets processed by the RecordMonitor or the GarbageCollector, but as draft records will be ignored anyway it should be safe to fix this issue that way.

dkd-friedrich (approved):
LGTM

dkd-friedrich:
@dogawaf Thanks for providing a pr, I'll do a backport to 11.5.

PR #3538, opened 2023-03-07
[BUGFIX] Use correct eval for site TCA

georgringer:
Instead of using 

```php
'eval' => 'required',
```

use 

```php
'required' => true,
```


dkd-kaehm (approved):
Thanks for the fix.

Issue #2903, opened 2021-03-19
[FEATURE] Don't disturb editors to do their job, if Apache Solr Server is not responsible


dkd-kaehm:
**User story:**
As an editor I want to edit contents in TYPO3 even the Apache Solr Server is down, to be able to do my job.

**Describe the solution you'd like**
Currently no exact suggestion for solution. The behavior must be checked. May be some EXT:Solr addon is the source of troubles. 

**Additional context**
Editors get errors like `Solr http error ....` by editor actions.
Internal Issue: 115597
Reporter: @bibusdax

**Target versions**
current development state, 11, 10

dkd-kaehm:
@dkd-friedrich 
IMHO it is solved by one of your patches for exceptions handling.
Can you please check and close this issue?

dkd-friedrich:
Unfortunately this issue still exists, but it seems that Solarium clients and adapters react differently in recent versions. Exception `HttpException` is still thrown if a `ClientException` occurs and if so, the exception is not catched in EXT:solr. I am currently not sure what exactly triggers a `ClientException`, if a Solr core is unavailable no exception is thrown and just the response shows the server error.

IMHO there is currently no simple way to handle communication issues during garbage collection. Garbage collection e.g. removes hidden or deleted records from queue and from the index, but if the index update fails, there is currently no process that ensures that the index will be cleaned later.

I think in a first step we can log failing index updates, by checking the response status and catching the exception. This will ensure that the failed update is logged and that the editor is not disturbed. Later on a new index queue status could be introduced that allows the Scheduler task to remove the index documents later.

PR #3539, opened 2023-03-10
[TASK] Handle Solr connection exception

dkd-friedrich:
# What this pr does

Solr requests may fail due to server errors or faulty queries, thrown exceptions or just an invalid status code indicate such errors.

This commit ensures that exceptions are handled and a valid response with a meaningful status code is returned. Handling is centralized in SolrService so that other components just have to deal with the status code.

Additionally the GarbageCollection is considering this status and is logging missing index update.

# How to test

- Configure an indexing configuration e.g. for news
- Simulate invalid Solr responses
- Try to hide an indexed record in TYPO3 backend

Record update shouldn't fail, but a log entry should be in place.

Resolves: #2903


dkd-kaehm:
Possible related issues/prs:
* #3108 / #3109

PR #3543, opened 2023-03-13
[BUGFIX] Fix Solr response handling during indexing

dkd-friedrich:
# What this pr does

If a Solr response is not valid (=200) the index queue item should be marked as failed to prevent the indexer of being blocked. This commit fixes this issue by throwing an `IndexingException` which causes the `IndexService` to mark the item as failed.

# How to test

- Add an indexing configuration
- Initialize the queue
- Simulate an indexing error on Solr side, e.g. by stopping the Solr container
- Try to index 

Queue items should now be marked as failed and the indexing should not fail silently.

Resolves: #3108


dkd-kaehm (approved):
:+1:
LGTM

PR #3546, opened 2023-03-16
[TASK] Rename misnamed util function getExceptionTime to getExecution…

twojtylak:
# What this pr does

Rename misnamed util function `getExceptionTime` to `getExecutionTime`

In older versions, the function was named `getExectionTime` and was then renamed to `getExceptionTime`, presumably via rector


dkd-kaehm (approved):
Thanks for fixing that issue.
Nice catch!

Issue #3534, opened 2023-03-05
[FEATURE] Make settings for statistics dynamic


ayacoo:
**Is your feature request related to a problem? Please describe.**
We are in the backend in the statistics module with a core. At the moment, the days and the number of results of the statistics are firmly defined in the code.  (30 days, 5 entries)

**Describe the solution you'd like**
The settings should be controllable via TypoScript.

**Additional context**
https://github.com/TYPO3-Solr/ext-solr/blob/main/Classes/Controller/Backend/Search/InfoModuleController.php#L127

In the course of this, the old ticket https://github.com/TYPO3-Solr/ext-solr/issues/1583 could also be addressed.

**Target versions**
11.5.x / 12

As we need this with a client, we will take over the implementation in the coming period.

PR #3548, opened 2023-03-19
[FEATURE] Make statistics more dynamic

ayacoo:
# What this pr does

The statistics are hard-coded so far. It is now defined in Typoscript. If no values are defined, the previous values are set as the default value in the code to prevent a breaking change.

# How to test

Simply add TypoScript and look at the statistics in the backend info module.

# Hint

The documentation is adapted if everything fits in terms of content and the variable naming is also good enough.

Fixes: #3534

### DIFF
--- a/Classes/Controller/Backend/Search/InfoModuleController.php
+++ b/Classes/Controller/Backend/Search/InfoModuleController.php
@@ -24,6 +24,7 @@ use Psr\Http\Message\ResponseInterface;
 use TYPO3\CMS\Core\Messaging\FlashMessage;
 use TYPO3\CMS\Core\Registry;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
 
 /**
  * Info Module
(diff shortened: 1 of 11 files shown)


bmack:
How about pageTsConfig instead?

ayacoo:
Would certainly be more elegant but would then also open up another place where you can configure. Therefore I would rather stay with the TypoScript.

dkd-kaehm (changes requested):
The docs part is missing.

ayacoo:
@dkd-kaehm I will gladly add the documentation if I get the OK for variable naming and location ;-)

dkd-kaehm:
> @dkd-kaehm I will gladly add the documentation if I get the OK for variable naming and location ;-)

If you mean following names, they are fine ;):

```
statistics = 1
  statistics {
    anonymizeIP = 1
    topHits {
      days = 30
      limit = 5
    }
    noHits {
      days = 30
      limit = 5
    }
    queries {
      days = 30
      limit = 100
    }
  }
```

dkd-kaehm (approved):
Thanks for your work.
We'll proceed with the merge.

Todo Maintainers: 
- [ ] merge suggestions
- [ ] **use squash button for merging!**

Issue #2984, opened 2021-07-20
[BUG] Configuring tx_solr[q] as speaking URL part crashes faceting


calien666:
Our customer wanted to have the search param as part of the uri speaking. After introducing the faceting didn't work and produced bad link requests.

Configure a page with a Route enhancement for search query:
```yaml
  Search:
    type: Plugin
    namespace: tx_solr
    routePath: '/{q}'
    limitToPages: [{your search page}]
    defaults:
      q: ''
    requirements:
      q: '^.*$'
```
Search is now callable like
```
https://example.com/search/my%20Search%20word
```
With active faceting the generated links for toggling will look like this:

```
https://example.com/search/%23%23%23tx_solr:q%23%23%23/?tx_solr%5Bfilter%5D%5B0%5D=keywords%3AWaste+plastics
```
As you see, the building structure and replacement of the placeholders could not work, because of url encoding the # to %23.

**Expected behavior**
Clean search URLs with speaking parameters like:

```
https://example.com/search/recycling/?tx_solr%5Bfilter%5D%5B0%5D=keywords%3AWaste+plastics
```

**Used versions:**
 - TYPO3 Version: 9.5.27
 - Browser: All browsers, server side problem
 - EXT:solr Version: 11.0.4
 - Used Apache Solr Version: 8.5.2
 - PHP Version: 7.4.21
 - MySQL Version: 5.7.34

**Additional context**

After debugging I found one line changing could solve the problem, but I am not sure, if this will fit to all cases.

Inside solr/Classes/Domain/Search/Uri/SearchUriBuilder.php, method buildLinkWithInMemoryCache the xtbase URI-Builder gets $structure as arguments, but $arguments is useful, too, since Extbase UriBuilder can accept multiple associative arrays, too.

changing

```php
            self::$missCount++;
            $this->uriBuilder->reset()->setTargetPageUid($pageUid);
            $uriCacheTemplate = $this->uriBuilder->setArguments($structure)->setUseCacheHash(false)->build();
```

to 

```php
            self::$missCount++;
            $this->uriBuilder->reset()->setTargetPageUid($pageUid);
            $uriCacheTemplate = $this->uriBuilder->setArguments($arguments)->setUseCacheHash(false)->build();
```

solved the problem for me.

trokohl:
I have a similar problem and this BugFix works

trokohl:
But i think you must also change
$hash = md5($pageUid . json_encode($structure));

to 

[...]

PR #3562, opened 2023-03-29
[BUGFIX] Retry Uri Building after exception

saitho:
# What this pr does

When there are route enhancers defined for a page containing the Solr search plugin, e.g. to realize pretty URLs for the search term, filters or pagination, there are issues with the URI caching that utilizes placeholders. Placeholders usually do not get replaced, leading to the issue described in #2984.

This PR implements a solution for route enhancers that define `requirements` in their parameters. The requirements are checked against the placeholders, leading to an exception. This exception is caught and it will then proceed the URI building with the original arguments instead of the placeholder arguments.

Fixes: #2984

# How to test

## Scenario 1: Speaking URL for search term

See https://github.com/TYPO3-Solr/ext-solr/issues/2984

## Scenario 2: Speaking URLs for pagination

1. Set up the following route enhancer (set the page containing the Solr list plugin in `limitToPages`):
```yaml
routeEnhancers:
  SolrPagination:
    type: Plugin
    namespace: 'tx_solr'
    routePath: '/{page}'
    defaults:
      page: 0
    requirements:
      page: '\d+'
    limitToPages: [1]
```
2. Clear TYPO3 cache
3. Open the Solr list plugin (e.g. https://mysite.dev/search)
4. The pagination links should be pretty (e.g. https://mysite.dev/search/1, https://mysite.dev/search/2)

### DIFF
--- a/Classes/Domain/Search/Uri/SearchUriBuilder.php
+++ b/Classes/Domain/Search/Uri/SearchUriBuilder.php
@@ -27,6 +27,7 @@ use ApacheSolrForTypo3\Solr\System\Configuration\TypoScriptConfiguration;
 use ApacheSolrForTypo3\Solr\System\Url\UrlHelper;
 use ApacheSolrForTypo3\Solr\Utility\ParameterSortingUtility;
 use Psr\EventDispatcher\EventDispatcherInterface;
+use Symfony\Component\Routing\Exception\InvalidParameterException;
 use TYPO3\CMS\Core\Http\Uri;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\Mvc\Web\Routing\UriBuilder;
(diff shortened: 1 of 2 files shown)


Classes/Domain/Search/Uri/SearchUriBuilder.php
```diff
@@ -277,11 +278,26 @@ protected function buildLinkWithInMemoryCache(?int $pageUid, array $arguments):
...
+            } catch (InvalidParameterException $exception) {
+                // the placeholders may result in an exception when route enhancers with requirements are active
+                // In this case, try to build the URL with original arguments
+                $hash = md5($pageUid . json_encode($arguments));
+                if (isset(self::$preCompiledLinks[$hash])) {
+                    self::$hitCount++;
+                    $uriCacheTemplate = self::$preCompiledLinks[$hash];
+                } else {
+                    $uriCacheTemplate = $this->uriBuilder->setArguments($arguments)->build();
+                    /* @var UrlHelper $urlHelper */
```
dkd-kaehm, line 296:
```suggestion
                    /** @var UrlHelper $urlHelper */
```

dkd-kaehm (changes requested):

dkd-kaehm (approved):
LGTM

PR #3576, opened 2023-03-31
[TASK] Migrate Integration tests to fixture extensions

bmack:
This change introduces fixture extensions

* fake_extension
* fake_extension2
* fake_extension3

which can be loaded into a IntegrationTest class.

This way, manual definition of TCA and SQL db files is removed.

In addition, the PageIndexerTest now uses
real frontend subrequests (the fixture extensions
are required for this change), to reduce the usage to IntegrationTest->fakeTSFE(), which hopefully
will be killed at a later stage.

At the same time the be_users.xml fixture has now
become a CSV, in order to move towards non-deprecated items.

# What this pr does

Please add a description here

# How to test


dkd-kaehm (approved):
LGTM!

PR #3583, opened 2023-03-31
[TASK] Code CleanUp

sfroemkenjw:
# What this pr does

Apply various inspections of PhpStorm:

* Remove unneeded annotations
   * Remove `@var`, `@param`, `@return`, which are without proper description 
   and its type is provided by PHPs type-hinting definition.
* Add void as method return type
* Replace mb_substr/mb_strpos with str_starts_with
* Simplify if conditions
* Remove default values
* Remove superflous type casts

Relates: #3376


dkd-kaehm (approved):
Now the whole Classes is cleaned up.

PR #3584, opened 2023-03-31
[TASK] Simplify further Integration tests

bmack:
# What this pr does

This change removes some abstractions from
ControllerTests.

This is now done in favor of a new
method in the god-class "IntegrationTest",
which now ships with "IntegrationTest->indexPages()"
a method where page Ids are added to the index
via a Frontend Subrequest (can be done separately
via "addPageToIndex()") and can also run
a indexed item via "executePageIndexer".

The new method "indexPages()" should be used
instead of "indexPageIds()" which does not use
EXT:solr's regular indexing method in Frontend
but fakes this in the test.

In addition, further cleanups are made in the
XML fixtures to always include the slug field
as well. More to come in this area.

# How to test

Run the tests

Related: #3376


dkd-kaehm (dismissed):
LGTM

dkd-kaehm (approved):
TOP!

PR #3587, opened 2023-04-03
[TASK:T12] Reports: Make them compatible to \TYPO3\CMS\Reports\Status…

taieb123:
…ProviderInterface

# What this pr does

Please add a description here

# How to test


dkd-kaehm:
@taieb123 
What exactly do or fix the wraps with `if(!empty($this->getStatus())){` and returning empty strings?
Why do we need empty Labels?

PR #3618, opened 2023-04-25
[TASK] Migrate test fixtures to CSV - part 1

bmack:
In order to use the latest TYPO3 Core Testing Framework, all test fixtures need to be migrated to a CSV
format.

This change marks the first part for the migration.

See #3614


dkd-kaehm (approved):
Thanks Benni.
I'll merge this now!

dkd-kaehm:
The srutinizer is to ignore for now, because it ignores the composers autoload settings.
Most probably now is the time for switching to the PhpStan....

PR #3621, opened 2023-05-02
[TASK] Use stable testing framework v7

bmack:
This change updates the TYPO3 testing framework to v7.0.1, which is compatible with TYPO3 v11 and v12.


bmack:
merged already

Issue #3622, opened 2023-05-03
[BUG] Cannot instantiate interface ConfigurationManagerInterface


chroenmeister:
**Describe the bug**
Using the latest 11.2.2 release getting this exception while opening plugin instance:

Cannot instantiate interface TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface

**To Reproduce**
1. Open Plugin Search: Form, Result, Additional Components

**Screenshots**

![grafik](https://user-images.githubusercontent.com/5887116/235876303-c35b851e-7e31-4290-bea8-0806eb108eda.png)

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.37
 - EXT:solr Version: 11.2.2
 - PHP Version: 7.4.11

brandung-sjorek:
The same applies to:

- TYPO3 Version: 11.5.26
- EXT:solr Version: **11.5.2**
- PHP Version: 7.4.11

dkd-kaehm:
We can not reproduce the issue.

@lfehr, @chroenmeister, @brandung-sjorek 
Please provide the installed extension list, or better those, which implementing the classes   
`TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface`.

Most probably, there is some extension which implements this interface and therefore the DI can not resolve the class.

brandung-sjorek:
@dkd-kaehm We'll (@chroenmeister and me) look into it, and give you Feedback within the next 2 weeks. Please keep the issue open, in the case we find the root-cause in the TYPO3 Solr-Extension itself.

Thanks for your reply …

rr-it:
https://github.com/TYPO3-Solr/ext-solr/blob/3fc0a33f627530eb9b7c4ad0b6a7f7e894977467/Classes/System/UserFunctions/FlexFormUserFunctions.php#L206-L216

We are trying to make an instance of an `Interface::class` here. Not sure if this must be an instance of an actual implementation of this interface?

```php
 $configurationManager = GeneralUtility::makeInstance(\ApacheSolrForTypo3\Solr\System\Configuration\ConfigurationManager::class);
$configurationManager = GeneralUtility::makeInstance(\TYPO3\CMS\Extbase\Configuration\ConfigurationManager::class);
$configurationManager = GeneralUtility::makeInstance(\TYPO3\CMS\Core\Configuration\ConfigurationManager::class);
```


[...]

PR #3626, opened 2023-05-08
[BUGFIX] make CE search form in backend editable again

rr-it:
Especially on TYPO3 10 the userfunction to load the flex form configuration produced an error.

This implements `TYPO3\CMS\Extbase\Object->get()` - deprecated by TYPO3 12:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.4/Deprecation-90803-DeprecationOfObjectManagergetInExtbaseContext.html

# Technical details

`FlexFormUserFunctions` does not implement the extbase controller and thereby does not have dependency injection.
With missing dependency injection the instantiation of the `Interface::class` by `GeneralUtility::makeInstance(ConfigurationManagerInterface::class)` fails with:
`Cannot instantiate interface TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface`

The solution to instantiate an object of `ConfigurationManager::class` is not feasible here - there are different implementations of `ConfigurationManagerInterface::class` for the frontend and backend.

So we are falling back to the solution provided in `GeneralUtility::makeInstance()`:
https://github.com/TYPO3/typo3/blob/808ef61a53bc1e7ef65b715c13727b0f2d086c24/typo3/sysext/core/Classes/Utility/GeneralUtility.php#L3441-L3445
> You may want to use \TYPO3\CMS\Extbase\Object\ObjectManager::get() if you
> want TYPO3 to take care about injecting dependencies of the class to be 
> created. Therefore create an instance of ObjectManager via
> GeneralUtility::makeInstance() first and call its get() method to get
> the instance of a specific class.

# Further reading on extbase and dependency injection

- https://www.derhansen.de/2022/10/problem-with-all-extbase-plugins-during-typo3-11-update.html
- https://github.com/FluidTYPO3/vhs/pull/1368#issuecomment-327141164

# How to test

This is reproducible by multiple users.

Fixes: #3622

---

Maintaners comments:

Todos: 

- [x] Rebase and change target into release-11.5.x

Port into:
- [ ] release-11.2.x

### DIFF
--- a/Classes/System/UserFunctions/FlexFormUserFunctions.php
+++ b/Classes/System/UserFunctions/FlexFormUserFunctions.php
@@ -26,6 +26,7 @@ use Doctrine\DBAL\Driver\Exception as DBALDriverException;
 use function str_starts_with;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 use TYPO3\CMS\Extbase\Configuration\ConfigurationManagerInterface;
+use TYPO3\CMS\Extbase\Object\ObjectManager;
 use TYPO3\CMS\Extbase\Utility\LocalizationUtility;
 
 /**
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
Please don't do thin against main branch. Rebase the changes against release-11.5.x and change the target branch to the release-11.5.x branch.

dkd-kaehm (approved):

Issue #3641, opened 2023-05-22
[BUG] Garbage collector throws error in backend when creating a workspace placeholder (PHP 8)


baschny:
If you have a workspace and edit an existing record, you get an Exception when trying to save it for the first time:
```
PHP Warning: Undefined array key 69665 in /app/private/typo3conf/ext/solr/Classes/GarbageCollector.php line 227
```

**To Reproduce**
1.  Login as an user with a workspace
2. Edit some tt_content which is already published
3. Hit "Save"
4. See above mentioned exception

_Note_: The record is still saved correctly.

**Expected behavior**
No error

**Used versions:**
 - TYPO3 Version: 11.
 - EXT:solr Version: 11.5.2
 - PHP Version: 8.1

**Findings**

The problem is that the workspace "Save" will create a shadow copy of the record and `Util::isDraftRecord` is not checking for it correctly.

baschny:
I added a fix for this in #3642 and pointed it to 11.5 branch so that it can be fixed also in the current "stable" version.

kitzberger:
In the meantime the relevant code parts have been "fixed" by some backport to the 11.5 branch (315ddf397e3f9cd8574288282d27bc49675ee6f4). But it's not working!

@baschny's fix is though. Please merge.

PR #3642, opened 2023-05-22
[BUG] Fix detection of "draft records" in workspaces

baschny:
Also consider "modified placeholder records" as draft records to avoid them being indexed and also to avoid exceptions in the backend (PHP 8) in the garbage collector.

Resolves: #3641
Relates: #3531

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -114,9 +114,14 @@ class Util
         $isWorkspaceRecord = false;
 
         if ((ExtensionManagementUtility::isLoaded('workspaces')) && (BackendUtility::isTableWorkspaceEnabled($table))) {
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
LGTM!
Thanks for fixing that issue.

kitzberger:
I can confirm that this fix is working for version 11.5. Thanks @baschny :muscle:

dkd-kaehm:
@baschny 
Thanks for this fix.
Is that fix only for TYPO3 <=11.5 or can we port it for TYPO3 12 as well?

baschny:
This fix should work also on all newer versions, this part has not changed in the core AFAIK

PR #3650, opened 2023-05-27
[TASK] Raise phpstan to level 5 and remove prophecy tests

bmack:
# What this pr does

This change removes prophecy from all tests, as the framework is not actively maintained anymore.

Some tests which check for TypeErrors are removed as this is now part of PHP 8 language.

At the same time, phpstan is raised to level 5.


dkd-kaehm:
I'll contunue on that.

dkd-kaehm (approved):

PR #3652, opened 2023-06-01
[TASK] Simplify phpunit invocations

bmack:
# What this pr does

This change adds more information to the phpunit
xml files so the calls to phpunit can be simplified.

# How to test

The PR just needs to have CI green.


dkd-kaehm (approved):
LGTM

PR #3654, opened 2023-06-01
Use f:translate instead of s:translate in Index.html

sfroemkenjw:
# What this pr does

Replace removed s:translate VH with fluids f:translate


dkd-kaehm (approved):
LGTM

PR #3664, opened 2023-06-02
[TASK] Housekeeping: Remove cache key from ci.yaml

bmack:
# What this pr does

This change removes the cache key from the GitHub workflow as this is "OK" from a time-consuming performance.

Fixes: #3663


dkd-kaehm (approved):
LGTM!

Issue #3669, opened 2023-06-02
[TASK] Remove usages of Util class when possible


bmack:
Reduce Util class usages to show the actual dependencies to TSFE etc.

PR #3670, opened 2023-06-02
[TASK] Reduce usages of Util class

bmack:
The Util class methods "getLanguageUid()" and "getFrontendUserGroupsList()" have been removed as they can be retrieved differently.

Fixes: #3669


dkd-kaehm (approved):
LTGM!

Issue #3677, opened 2023-06-05
[TASK] Remove unneeded code pieces


bmack:
There are still old interfaces around

PR #3678, opened 2023-06-05
[!!!][TASK] Remove deprecated and unused code

bmack:
This change removes some interfaces which haven't been in use for a while now, or are not needed anymore.

The SiteInterface was not needed anymore, as only Site-based TYPO3 installations are out there.

Fixes: #3677


dkd-kaehm (approved):
LGTM!

Issue #3698, opened 2023-06-14
[BUG] Wrong return type for AbstractSolrService->getClient()


o-ba:
Return type has to be `Client`, since `null` is not possible here.

PR #3699, opened 2023-06-14
[BUGFIX] Correct return type of AbstractSolrService->getClient()

o-ba:
Fixes: #3698


dkd-kaehm (approved):
Nice catch!

dkd-kaehm:
Build is due of troubles on https://get.typo3.org/ broken. We want to restart the build after it is available again.

PR #3701, opened 2023-06-15
[BUGFIX] Force score to float

georgringer:
Force the score to be a float.

TBH: I didn't yet that test that myself but a colleague had the issue and just patched it locally


dkd-kaehm:
@georgringer thanks for the patch.
Can you please peek up a little bit more infos about this issue from your colleague?
* the versions stack
* how to reproduce the issue
* do we want ports in previous versions?

dkd-kaehm (approved):
LGTM, thanks for the fix.

Issue #2151, opened 2018-10-09
FlexForm filter values are force quoted


rschaufler:
Version: 8.1.2

When trying to add more complex filter rules to the FlexForm filter section, all values are quoted which prevents them to work.

**Examples:**

Works:

![screenshot 2018-10-09 at 11 45 58](https://user-images.githubusercontent.com/3216408/46661182-f5058200-cbb8-11e8-904d-9568dd7fcddc.png)

Does not work:

![screenshot 2018-10-09 at 11 46 37](https://user-images.githubusercontent.com/3216408/46661217-0484cb00-cbb9-11e8-83e7-7a8df5140fcf.png)

Reason: https://github.com/TYPO3-Solr/ext-solr/blob/8.1.2/Classes/System/Service/ConfigurationService.php#L135

Maybe the quotes should not be applied at all and the user should just be aware of that. If using a phrase that contains spaces, the user could apply quotes himself

markuskobligk:
I can confirm this behaviour, it breaks the autoauggest functionality in my project sind the value of filter gets quoted and breaks the request.

rschaufler:
also found that this was intentionally implemented, see [https://github.com/TYPO3-Solr/ext-solr/issues/1742](https://github.com/TYPO3-Solr/ext-solr/issues/1742)

i think this fix broke the functionality more than it helped.

rschaufler:
@markuskobligk if you want to fix this, just XCLASS the class mentioned in my issue description and remove the quoting. That's how i solved this problem for now

timohund:
I would propose to add functional tests for this AND for #1742 instead of just removing it.

benwalch:
@timohund @dkd-kaehm 
I think the best way would be to remove the force quoting entirely and update the documentation accordingly that string values with spaces have to be quoted.
This would add the most flexibility, and users can completely rely on the lucene query parser syntax when inserting custom filters. https://lucene.apache.org/core/3_0_3/queryparsersyntax.html

echavaillaz:
I make a PR: https://github.com/TYPO3-Solr/ext-solr/pull/3703

I add a regex to prevent too big breaking change. The old behaviour is still working, so if a string value contains space(s) and no special characters of the solr query parser, the string is always wrapped with double quotes.

But if the string contains special characters no wrapping happen.

There is some cases where this PR can break, for example if the filter value is something like `toto AND tata` or `music (rock)` but I think this is not usual use case. So the breaking is as little than possible.

PR #3703, opened 2023-06-21
!!![FEATURE] Allow using complex filter values in FlexForm

echavaillaz:
With this PR it is now possible to use complex query in FlexForm filter value.

I add a regex to prevent a too big breaking change. The old behaviour is still working, so if a string value contains space(s) and no special characters of the solr query parser, the string is always wrapped with double quotes. But if the string contains special characters no wrapping happen (special characters are `+ - && || ! ( ) { } [ ] ^ " ~ * ? : \`.

There is some cases where this PR can break, for example if the filter value is something like `toto AND tata` or `music (rock)` or `my "flow" is`. Here the wrapping and the escaping of the inner double quote have to be manually updated like this `"toto AND tata"`, `"music (rock)"` and `"my \"flow\" is"`.

Fixes: #2151

### DIFF
--- a/Classes/System/Service/ConfigurationService.php
+++ b/Classes/System/Service/ConfigurationService.php
@@ -98,12 +98,13 @@ class ConfigurationService
      */
     protected function getFilterFromFlexForm(array $flexFormConfiguration): array
     {
-        $filterConfiguration = [];
         $filters = ObjectAccess::getPropertyPath($flexFormConfiguration, 'search.query.filter');
 
         if (empty($filters)) {
-            return $filterConfiguration;
+            return [];
         }
+
+        $filterConfiguration = [];
 
         foreach ($filters as $filter) {
             $filter = $filter['field'];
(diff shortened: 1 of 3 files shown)


dkd-kaehm (changes requested):
The code base LGTM!
Please add 
* the hints from your issue comment https://github.com/TYPO3-Solr/ext-solr/issues/2151#issuecomment-1600623995
to the docs, so users are aware of.
* and a little sentence in Documentation/Releases/solr-release-12-0.rst about this feature.

dkd-kaehm:
@echavaillaz 
Please fix the issues from GH-actions:
https://github.com/TYPO3-Solr/ext-solr/actions/runs/5334225100/jobs/9668256731?pr=3703

For CS fixes just run this command:
`TYPO3_VERSION=12.4.x-dev composer tests:setup && composer t3:standards:fix`

The unit test and/or method `ConfigurationService::getFilterFromFlexForm()` must be adjusted:
```
1) ApacheSolrForTypo3\Solr\Tests\Unit\System\Service\ConfigurationServiceTest::canOverrideConfigurationWithFlexFormSettings with data set "simpleInteger" ('id', 4711, 'id:4711')
TypeError: ApacheSolrForTypo3\Solr\Tests\Unit\System\Service\ConfigurationServiceTest::canOverrideConfigurationWithFlexFormSettings(): Argument #2 ($filterValue) must be of type string, int given, called in /home/runner/work/ext-solr/ext-solr/.Build/vendor/phpunit/phpunit/src/Framework/TestCase.php on line 1608

/home/runner/work/ext-solr/ext-solr/Tests/Unit/System/Service/ConfigurationServiceTest.php:107
/home/runner/work/ext-solr/ext-solr/.Build/bin/phpunit:123

--

There was 1 failure:

1) ApacheSolrForTypo3\Solr\Tests\Unit\System\Service\ConfigurationServiceTest::canOverrideConfigurationWithFlexFormSettings with data set "escapedString" ('id', 'foo"bar', 'id:"foo\"bar"')
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
-    0 => 'id:"foo\"bar"'
+    0 => 'id:foo"bar'
 )
```

echavaillaz:
Ready to test?

dkd-kaehm (approved):

PR #3704, opened 2023-06-21
[BUGFIX] Make BeforeSearchResultIsShownEvent more usable

georgringer:
It would be cool if properties of the event could be changed. Otherwise the event doesn't help much - IMO

I need it e.g. for the pagination to be changed

### DIFF
--- a/Classes/Event/Search/BeforeSearchResultIsShownEvent.php
+++ b/Classes/Event/Search/BeforeSearchResultIsShownEvent.php
@@ -19,6 +19,7 @@ namespace ApacheSolrForTypo3\Solr\Event\Search;
 
 use ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\SearchResultSet;
 use ApacheSolrForTypo3\Solr\Pagination\ResultsPagination;
+use TYPO3\CMS\Core\Pagination\PaginationInterface;
 
 /**
  * This event is triggered before adding the search result to the fluid template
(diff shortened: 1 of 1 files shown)


Classes/Event/Search/BeforeSearchResultIsShownEvent.php
```diff
@@ -79,4 +80,9 @@ public function getCurrentPage(): int
     {
         return $this->currentPage;
     }
+
+    protected function setPagination(PaginationInterface $pagination): void
```
dkd-kaehm, line 84: Why not public?

Classes/Event/Search/BeforeSearchResultIsShownEvent.php
```diff
@@ -19,6 +19,7 @@
 
 use ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\SearchResultSet;
 use ApacheSolrForTypo3\Solr\Pagination\ResultsPagination;
```
dkd-kaehm, line 21: This line should is not needed anymore.

dkd-kaehm (changes requested):
Thanks for this patch.
There are most probably two changes required.

dkd-kaehm (approved):
LGTM!
The PR can be squashed and merged.

georgringer:
@dkd-kaehm I don't know if other properties should be setable as well but certainly all other events must be changed as well to be usable for extension authors. I suggest to use a different PR for that

georgringer:
@dkd-kaehm wanna merge that?

bmack:
Can we add them to the other Events that are related to Actions/Controllers?

PR #3708, opened 2023-06-28
[DOC] Solr claims to be not configured in backend context, although I did it well. What can be the reason?

haraldwitt:
…t well. What can be the reason?

# What this pr does

Please add a description here

# How to test

Todo:

- [ ] port into release-11.5.x

### DIFF
--- a/Documentation/FAQ/Index.rst
+++ b/Documentation/FAQ/Index.rst
@@ -678,10 +678,13 @@ With Typo3 10 and solr 11.0.1 it was possible for solr to read it's configuratio
 
 The solution:
 
-If the root page is a shortcut, the destination MAY NOT be access restricted! This can e. g. be achieved in the following way.
+If the root page is a shortcut, the destination MAY NOT be access restricted!  
+In addition a login page itself cannot get hidden after login. This would lead to errors and no redirections will take place.  
+This can e. g. be achieved in the following way.
 
-* The root page is a shortcut to the first sub-page.
-* The first sub-page is the login page, which is not access restricted. Instead this page gets hidden if a user is logged in. And it redirects to the next access restricted page or sub-tree after successful login.
-* The next sub-page is the access restricted page or sub-tree.
+* The root page is a shortcut to the 1st subpage.
+* The 1st subpage is a shortcut to an unprotected login page, and will be hidden after successfull login.
+* The 2nd subpage is an pagetree with the access restricted content.
+* And last but not least there is the login page (e. g. 3rd subpage) in an unprotected area, which will never get hidden. So ext-solr can read its configuration here.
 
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
LGTM!
:+1: 

We want backport into release-11.5.x branch.

Documentation/FAQ/Index.rst
```diff
@@ -663,3 +663,25 @@ Example:
...
+The reason for that:  
+
+With Typo3 10 and solr 11.0.1 it was possible for solr to read it's configuration from shortcut pages directly. With Typo3 11 and solr 11.5 these shortcuts are followed. And if the access to the destination of the shortcut is restricted, solr cannot read the configuration because solr is not logged in.
+
+The solution:
+
+If the root page is a shortcut, the destination MAY NOT be access restricted! This can e. g. be achieved in the following way.
+
+* The root page is a shortcut to the first subpage.
+* The first subpage is the login page, which is not access restricted. Instaed this page gets hidden if a user is logged in. And it redirects to the next access restricted page or subtree after successful login.
```
dmitryd, line 684: Instaed -> Instead

dkd-kaehm, line 684:
```suggestion
* The root page is a shortcut to the first sub-page.
* The first sub-page is the login page, which is not access restricted. Instead this page gets hidden if a user is logged in. And it redirects to the next access restricted page or sub-tree after successful login.
```

Documentation/FAQ/Index.rst
```diff
@@ -663,3 +663,25 @@ Example:
...
+
+The solution:
+
+If the root page is a shortcut, the destination MAY NOT be access restricted! This can e. g. be achieved in the following way.
+
+* The root page is a shortcut to the first sub-page.
+* The first sub-page is the login page, which is not access restricted. Instead this page gets hidden if a user is logged in. And it redirects to the next access restricted page or sub-tree after successful login.
+* The next subpage is the access restricted page or subtree.
+
+This way already logged in users will be redirected directly to the restricted page / subtree, because the login page is hidden and so the restricted page is the first subpage now.
```
dkd-kaehm, line 687:
```suggestion
* The next sub-page is the access restricted page or sub-tree.

This way already logged in users will be redirected directly to the restricted page / sub-tree, because the login page is hidden and so the restricted page is the first sub-page now.
```

Documentation/FAQ/Index.rst
haraldwitt: Sorry dmitryd,
there is an error inside this comment. The login page cannot be hidden after sucessful login. This leads to an error and no redirect will occour. Please change this to:

* The root page is a shortcut to the 1st subpage.
* The 1st subpage is a shortcut to an unprotected login page, and will be hidden after successfull login
* The 2nd, 3rd, ... subpage maybe any protected content.
* And last but not least there is the login page in an unprotected region, which will never gets hidden.

Greetings 
Harald

Issue #3709, opened 2023-06-28
[TASK] Improve site and document id determination


dkd-friedrich:
**What should be done in the scope of this task?**

Determination of sites and/or document ids is partly inefficient as the root page is already available but methods are used that can handle any page in the page tree. By using methods that support any pages the RootPageResolver is unnecessarily triggered.

In EXT:solr versions prior 11.5 this may lead to errors as the RootPageResolver might be influenced by global TSFE objects.

PR #3710, opened 2023-06-28
[TASK] Improve site and document id determination

dkd-friedrich:
# What this pr does

Site and document id determination is improved by using root page ids if already available, so we can dispense with the unnecessary use of the RootPageResolver

Resolves: #3709


dkd-friedrich:
This change has an impact on the PageContext of solrfal, we have to discuss/solve this issue first.

PR #3711, opened 2023-06-29
[TASK:BP:11.5] Improve site and document id determination

dkd-friedrich:
# What this pr does

Site and document id determination is improved by using root page ids if already available, so we can dispense with the unnecessary use of the RootPageResolver

Ports: #3710
Resolves: #3709


dkd-friedrich:
This change has an impact on the PageContext of solrfal, we have to discuss/solve this issue first.

dkd-friedrich:
> This change has an impact on the PageContext of solrfal, we have to discuss/solve this issue first.

We discussed this internally, as this would be a breaking change for solrfal and is just a minor improvement and not a bugfix, we skip this for older solr versions.

PR #3719, opened 2023-07-07
Backend user should not be logged out after editing content on hidden pages

### LINKED ISSUE #3702: [BUG] Backend User gets logged out when editing page content on a hidden page beneath a Sys-Folder
Given the following page tree:

image

the backend user (admin or/and editor) that is trying to create or edit a content element on the page is logged out after saving the changes.
The behaviour is the same as described in #3246 

I was able to track it down in a similar way:

1. `ApacheSolrForTypo3\Solr\FrontendEnvironment\Tsfe->initializeTsfe()`
2. `$tsfe->determineId($serverRequest)`
3. `GeneralUtility::makeInstance(ErrorController::class)->pageNotFoundAction()`

After looking into pageNotFoundAction() I tracked it down to `\TYPO3\CMS\Core\Error\PageErrorHandler\PageContentErrorHandler->handlePageError()`

There, `\TYPO3\CMS\Core\Error\PageErrorHandler\PageContentErrorHandler->sendSubRequest()` is called, wich causes another reinitialisation of TSFE that may be messing up other $GLOBALS variables that are not backed up before like $GLOBALS['BE_USER'].

Only if the subrequest is created, the current backend user gets logged out, so instead of tracking down the root cause even further (to second initialization of tsfe / running through all middlewares) I implemented a workaround:

I created a custom PageContentErrorHandler that tries to check if the created TSFE is a solr-initialized one: https://gist.github.com/phth/fdb2ac5f648e5b73b130306320ba7f10
and configured it in config.yaml:

```
errorHandling:
  -
    errorCode: '404'
    errorHandler: PHP
    errorPhpClassFQCN: Webicorns\SitePackage\Error\PageContentErrorHandler
    errorContentSource: 't3://page?uid=637'
```

From this, I have made the following conclusions:

- solr is not really responsible for the behaviour
- the behaviour is implemented in TYPO3 core due to the internal subrequests
- the former workaround, described in #3246 - setting the feature flag subrequestPageErrors to false - is not an option anymore because it was removed in TYPO3 12.0, see https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-96107-DeprecatedFunctionalityRemoved.html
- the error may not occure if someone already uses a custom ErrorHandler

**Used versions**
 - TYPO3 Version: 12.4.2
 - EXT:solr Version: dev-main / f4d1395

dmitryd:
# What this pr does

This PR adds a check for page's "hidden" property because hidden pages cannot normally be used for initializing TSFE.

# How to test

Have a tree like this:

```
|- Root
    |- Folder
        |- Hidden page
```

Try editing content on the hidden page.

Fixes: #3702

### DIFF
--- a/Classes/FrontendEnvironment/Tsfe.php
+++ b/Classes/FrontendEnvironment/Tsfe.php
@@ -326,7 +326,7 @@ class Tsfe implements SingletonInterface
         }
         $pageRecord = BackendUtility::getRecord('pages', $pidToUse);
         $isSpacerOrSysfolder = ($pageRecord['doktype'] ?? null) == PageRepository::DOKTYPE_SPACER || ($pageRecord['doktype'] ?? null) == PageRepository::DOKTYPE_SYSFOLDER;
-        if ($isSpacerOrSysfolder === false) {
+        if ($isSpacerOrSysfolder === false && !$pageRecord['hidden']) {
             return $pidToUse;
         }
         /** @var ConfigurationPageResolver $configurationPageResolver */


dmitryd:
I would recommend to merge #3720 instead because it does the same as this one plus more.

PR #3722, opened 2023-07-11
[BUGFIX] Handle float values in options facet parser

saschanowak:
# What this pr does

Cast solr value to string to handle float values correctly.
When the result values are 0.01, 0.02, 0.03 the last count wins with the array key 0
see https://3v4l.org/2LOHk#v8.1.20

# Version
Tested on dev-master with TYPO3 12.4.3

### Maintainers notes:

Port:

- [x] release-11.0.x/ELTS
- [x] release-11.2.x/ELTS
- [x] release-11.5.x


dkd-kaehm (changes requested):
The changes are fine, 
but we want to add Unit/Integration test case for that case.

dkd-kaehm (approved):

PR #3728, opened 2023-07-14
[BUGFIX] Fix TypeError in SearchFormViewHelper

tillhoerner:
Problem:
Calling the `s:searchForm` ViewHelper with the `pageType` or `suggestPageType` argument with a value from a typoscript setting will lead to a TypeError

```html
<s:searchForm pageType="{settings.pageType}" suggestPageType="{settings.suggestPageType}">
```

`TYPO3\CMS\Extbase\Mvc\Web\Routing\UriBuilder::setTargetPageType(): Argument #1 ($targetPageType) must be of type int, string given, called in /var/www/html/private/typo3conf/ext/solr/Classes/ViewHelpers/SearchFormViewHelper.php`

Solution:
Cast both arguments to int

### DIFF
--- a/Classes/ViewHelpers/SearchFormViewHelper.php
+++ b/Classes/ViewHelpers/SearchFormViewHelper.php
@@ -198,7 +198,7 @@ class SearchFormViewHelper extends AbstractSolrFrontendTagBasedViewHelper
         $suggestUrl = $this->uriBuilder
             ->reset()
             ->setTargetPageUid($pageUid)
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
LGTM.

Which versions are affected?
Do we want the backports?

holgerkraemer:
@dkd-kaehm 
Version 11.5.5 is affected as well (Using PHP 8.2). Will there be a backport of this bugfix?

PR #3733, opened 2023-07-25
[BUGFIX] fix null-pointer-exception in AbstractStrategy

adamkoppede:
# What this pr does

Fixes a null-pointer-exception that occurred when the "page type" field of a hidden root page was changed.

# How to test

We weren't able to re-create the issue on a mirrored instance, since the corresponding index-queue item could not be created via the "Index Queue" TYPO3 backend module. We directly imported the database record for reproduction. It's possible that such an item cannot be created during normal operations.

We checked the patch on TYPO3 11 with PHP 8 against the 11.5.3 release.


dkd-kaehm:
Was fixed in #3770 and ported to 11.5.x in #3778

Issue #3734, opened 2023-07-25
[BUG] Indexing fails for SOLR_RELATION


kitzberger:
When trying to index a record that has a SOLR_RELATION without a `foreignLabelField` leads to an exception.

## To Reproduce

* Link two tables by an intermediate table
* Create a index config like below
* Try to index those records

### TCAs

TCA of the table `employee`:

```php
        'employments' => [
            'config' => [
                'type' => 'inline',
                'foreign_table' => 'tx_myextension_domain_model_employment',
                'foreign_field' => 'employee',
                'foreign_label' => 'department',
            ],
       ],
```

TCA of the MM table `employment` :

```php
        'employee' => [
            'config' => [
                'type' => 'select',
                'renderType' => 'selectSingle',
                'foreign_table' => 'tx_myextension_domain_model_record',
            ],
        ],
        'department' => [
            'config' => [
                'type' => 'select',
                'renderType' => 'selectSingle',
                'foreign_table' => 'tx_myextension_domain_model_department',
            ],
        ],
```

TCA of the table `departments`:

```php
        'employments' => [
            'config' => [
                'type' => 'inline',
                'foreign_table' => 'tx_myextension_domain_model_employment',
                'foreign_field' => 'department',
                'foreign_label' => 'employee',
            ],
        ],
```

### Index config

```
      employment_intM = SOLR_RELATION
      employment_intM {
        localField = employments
        multiValue = 1
      }
```

## Expected behavior
To work like before ;-) With TYPO3 10.4 and EXT:solr 11.2.3 this was working flawlessly.

## Screenshots
![image](https://github.com/TYPO3-Solr/ext-solr/assets/1405149/8595c10b-4cd7-4358-b8cf-7fd16aca6b5d)

## Used versions
 - TYPO3 Version: 11.5.30
 - EXT:solr Version: 11.5.3
 - PHP Version: 8.1

kitzberger:
I'm now almost sure that the problem is the missing `foreignLabelField`. Is that now mandatory? Doesn't say anything about that here: https://docs.typo3.org/p/apache-solr-for-typo3/solr/main/en-us/Configuration/Reference/TxSolrIndex.html#solr-relation

PR #3740, opened 2023-08-01
[BUGFIX] Prevent indexing error on missing 'foreignLabelField'

kitzberger:
Prevent indexing error on missing 'foreignLabelField' of SOLR_RELATION

Related: #3734


dkd-kaehm (approved):
Thanks for fixing that issue.

Issue #3741, opened 2023-08-02
[BUG] SolrRoutingMiddleware ineffective


dkd-friedrich:
**Describe the bug**

Currently the SolrRoutingMiddleware required by the SolrFacetMaskAndCombineEnhancer is ineffective, EXT:solr parameters in route path are ignored.

Cause is a bugfix for #3202, probably due to a misunderstanding the middleware was limited to indexing processes. However, the middleware is not part of the indexing and must be active in the normal frontend rendering.

**Expected behavior**

- SolrFacetMaskAndCombineEnhancer  should work as expected
- No warning should be logged if on root page and the PageTypeSuffix enhancer is enabled

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4, 11.5
 - EXT:solr Version: 12-beta, 11.5.1

PR #3743, opened 2023-08-02
[BUGFIX:BP:11.5] Fix EXT:solr route enhancer

dkd-friedrich:
Backport of #3742 
---

# What this pr does

SolrRoutingMiddleware was limited to indexing processs, but as the SolrFacetMaskAndCombineEnhancer requires this middleware to add the required parameters from the route path, this limitation causes that the route enhancer is ineffective.

This commit fixes this issue by reenabling the middleware for frontend requests and also ensures that route paths can be resolved correctly if PageTypeSuffix enhancer is active.

# How to test

Configure route enhancer:
- SolrFacetMaskAndCombineEnhancer
- PageTypeSuffix 

With this commit
- EXT:solr can handle and react to facet parameters in route path
- no warning `Could not resolve page by path "menu.json" and language "de"` if requesting a specific page type on root page
- no TypeError occurs if no options are passed to the SolrFacetMaskAndCombineEnhancer via `solr`

Ports: #3742 
Relates: #3202
Resolves: #3741

### DIFF
--- a/Classes/EventListener/EnhancedRouting/CachedPathVariableModifier.php
+++ b/Classes/EventListener/EnhancedRouting/CachedPathVariableModifier.php
@@ -52,7 +52,7 @@ class CachedPathVariableModifier
 
         $this->routingService = GeneralUtility::makeInstance(
             RoutingService::class,
-            $enhancerConfiguration['solr'],
+            $enhancerConfiguration['solr'] ?? [],
             (string)$enhancerConfiguration['extensionKey']
         );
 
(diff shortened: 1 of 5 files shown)


dkd-kaehm (approved):
LGTM

vbdetemplephilipp:
After upgrading from 11.5.3 to 11.5.4 some frontend requests are stuck in the do while loop in:
\ApacheSolrForTypo3\Solr\Middleware\SolrRoutingMiddleware::retrievePageInformation

For those requests $scan is not set to false and therefore the loop runs infinitely.
Maybe a check for $path if it is the same as before for each loop run could help.

dkd-kaehm:
@vbdetemplephilipp thanks for reporting.
Could you please create new issue for that and if you know the exact case why `$scan` is never false, please add it as well.

PR #3749, opened 2023-08-08
[TASK] Upgrade solarium/solarium to 6.4.0+

dkd-kaehm:
Relates: #3725

Do not merge before real solarium release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
     "ext-libxml": "*",
     "ext-pdo": "*",
     "ext-simplexml": "*",
-    "solarium/solarium": "6.2.7",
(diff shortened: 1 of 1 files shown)


dkd-friedrich:
Main point mentioned for Solarium 6.4 is part of 6.3.0, support of PHP 7 is dropped.

Documentation is misleading/faulty at this point:
> Support for PHP 7 was removed in Solarium 6.4.0. Upgrade to PHP 8 first to use the latest Solarium version.

As there are currently no hints for an upcoming version 6.4 and we're planning to release EXT:solr 12 in the next days, we switch to solarium/solarium 6.3.2. I'm closing this PR in favor of #3799

Issue #3753, opened 2023-08-14
[BUG] Configset for Solr 9 is broken


jschlier:
**Describe the bug**

When trying to create a Solr core with the 12_0_0 configset Solr returns an error:
`Error CREATEing SolrCore 'core_de' Unable to create core [tecselect_de] Caused by: solr.scripting.xslt.XSLTResponseWriter`

Solr 9 moved the XLST respoonse writer to the "scripting" Module.
See [Changelog](https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html)

The namespace was changed in `Resources/Private/Solr/configsets/ext_solr_12_0_0/conf/solrconfig.xml`, but the scripting module is not activated.

**To Reproduce**

1. Install solr 9
2. copy the EXT:solr configset
3. try to create a new core (e. g. with `http://localhost:8983/solr/admin/cores?action=CREATE&name=core_de&configSet=ext_solr_12_0_0&schema=german/schema.xml`)

**Expected behavior**

The configuration for Solr should not produce errors.
This is easily fixed by enabling the scripting module with the others at the beginning of `solrconfig.xml`:

`<lib dir="${solr.install.dir:../../../..}/modules/scripting/lib/" regex=".*\.jar" />`

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.5
 - EXT:solr Version: main
 - Used Apache Solr Version: 9.2.0

PR #3754, opened 2023-08-14
[BUGFIX] Enable scripting module

jschlier:
# What this pr does

This PR fixes the Configuration for Solr by enabling the scripting module.
The XSLTResponseWriter is now part of this module and referenced later in the configuration.
Without enabling the scripting module Solr returns an error when trying to use the current configuration.

# How to test

Create a new Solr core with the new configuration, which now no longer results in an error.

Fixes: #3753

### DIFF
--- a/Resources/Private/Solr/configsets/ext_solr_12_0_0/conf/solrconfig.xml
+++ b/Resources/Private/Solr/configsets/ext_solr_12_0_0/conf/solrconfig.xml
@@ -9,6 +9,7 @@
 	<lib dir="${solr.install.dir:../../../..}/modules/langid/lib/" regex=".*\.jar" />
 	<lib dir="${solr.install.dir:../../../..}/modules/analytics/lib/" regex=".*\.jar" />
 	<lib dir="${solr.install.dir:../../../..}/modules/analysis-extras/lib/" regex=".*\.jar" />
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
@jschlier Thanks for fixing that issue, please add a test case for core creation and deletion in `main/Build/Test/cibuild_docker.sh`

dkd-kaehm:
@jschlier, @dkd-friedrich 
Why this line does not do the same?

https://github.com/TYPO3-Solr/ext-solr/blob/1b75d382ffb76c85e045fd7e13fa47507e0ca1d2/Resources/Private/Solr/solr.xml#L16

jschlier:
Ah, I did not see that. That might actually already fix this. I don't think I used this `solr.xml` configuration in my use case. 

In fact I don't think we use this file at all in any of our projects, but install and set up Solr automatically using the Solr install script. Might be a misconfiguration on our end then with TYPO3 12. In that case we can simply close this pull request and issue.

The documentation says this:

> As mentioned before the configSet is one single package, that contains all to customize a plain Apache Solr Server, to an Apache Solr Server, ready for EXT:solr & TYPO3.

But I guess that's not right then. The documentation is a bit contraditory, that might have confused me.

PR #3760, opened 2023-08-16
[BUGFIX] Less strict return types on resolving values

### LINKED ISSUE #3758: Too strict return values during indexing
**Describe the bug**
We have an integer field to be indexed by the page indexer.
Something like `plugin.tx_solr.index.queue.pages.fields.order_intS = tx_extension_some_field`.
The field is of course of type int in the database.

The method `\ApacheSolrForTypo3\Solr\IndexQueue\FrontendHelper\PageFieldMappingIndexer::resolveFieldValue` only allows arrays and strings as return values. But when it processes our field, it has type int. Thus throwing an exception.
Can the return value of this method be changed accordingly?

**To Reproduce**
1. Index an integer field with the PageFieldMappingIndexer

**Expected behavior**
No exception because of wrong return type.

 - TYPO3 Version:12.4.3]
 - Browser: -
 - EXT:solr Version: [e.g. 12.0.0]
 - Used Apache Solr Version: dev-main (commit b15df83)
 - PHP Version: 8.2.7
 - MySQL Version: mariadb 10.5.18

RazielleS:
Resolves #3758

# What this pr does

Use less strict return types on some methods to allow all possible field types.

# How to test

Use fields like dateS, boolS and intS during indexing.

Fixes: #3758

### DIFF
--- a/Classes/IndexQueue/AbstractIndexer.php
+++ b/Classes/IndexQueue/AbstractIndexer.php
@@ -119,7 +119,7 @@ abstract class AbstractIndexer
         string $solrFieldName,
         array $data,
         TypoScriptFrontendController $tsfe
-    ): array|float|int|string|null {
+    ): mixed {
         if (isset($indexingConfiguration[$solrFieldName . '.'])) {
             // configuration found => need to resolve a cObj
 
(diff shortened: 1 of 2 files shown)


stegmatze (approved):
Resolves Issues with 12.0.0-beta-1

dkd-kaehm (approved):
:+1:

Issue #3769, opened 2023-08-29
[BUG] Call to a member function getSolrConfiguration() on null


eliashaeussler:
**Describe the bug**
When running an upgrade wizard for EXT:container, I run into issues with EXT:solr due to missing site configuration.

Stack trace:

```
$ vendor/bin/typo3 container_deleteChildrenWithWrongPid
                                                            
  [ Error ]                                                 
  Call to a member function getSolrConfiguration() on null  
                                                            
Exception trace:
#0 ()
   typo3conf/ext/solr/Classes/Domain/Index/Queue/GarbageRemover/AbstractStrategy.php:118
#1 ApacheSolrForTypo3\Solr\Domain\Index\Queue\GarbageRemover\AbstractStrategy->deleteIndexDocuments()
   typo3conf/ext/solr/Classes/Domain/Index/Queue/GarbageRemover/AbstractStrategy.php:96
#2 ApacheSolrForTypo3\Solr\Domain\Index\Queue\GarbageRemover\AbstractStrategy->deleteInSolrAndUpdateIndexQueue()
   typo3conf/ext/solr/Classes/Domain/Index/Queue/GarbageRemover/PageStrategy.php:50
#3 ApacheSolrForTypo3\Solr\Domain\Index\Queue\GarbageRemover\PageStrategy->collectPageGarbageByContentChange()
   typo3conf/ext/solr/Classes/Domain/Index/Queue/GarbageRemover/PageStrategy.php:32
#4 ApacheSolrForTypo3\Solr\Domain\Index\Queue\GarbageRemover\PageStrategy->removeGarbageOfByStrategy()
   typo3conf/ext/solr/Classes/Domain/Index/Queue/GarbageRemover/AbstractStrategy.php:63
#5 ApacheSolrForTypo3\Solr\Domain\Index\Queue\GarbageRemover\AbstractStrategy->removeGarbageOf()
   typo3conf/ext/solr/Classes/Domain/Index/Queue/UpdateHandler/GarbageHandler.php:78
#6 ApacheSolrForTypo3\Solr\Domain\Index\Queue\UpdateHandler\GarbageHandler->collectGarbage()
   typo3conf/ext/solr/Classes/Domain/Index/Queue/UpdateHandler/EventListener/ImmediateProcessingEventListener.php:138
[...]

PR #3770, opened 2023-08-29
[BUGFIX] Delete index documents without available site

eliashaeussler:
This PR fixes a bug where an index document does not a have a site object assigned. In this case, the index document can be safely removed.

Since this bug occurred in a TYPO3 v11.5 environment, the PR should be backported to the 11.5.x branch.

Resolves: #3769
Releases: main, 11.5


dkd-friedrich (approved):
LGTM

PR #3778, opened 2023-09-11
[BUGFIX] Delete index documents without available site

dkd-kaehm:
Resolves: #3769
Releases: main, 11.5
Ports: #3770


dkd-friedrich (approved):
LGTM

PR #3784, opened 2023-09-15
[TASK] Sync with EXT:solr addons :: Fix Type-Hinting issues

dkd-kaehm:
Fixes regression after #3764

### DIFF
--- a/Classes/Event/IndexQueue/AfterIndexQueueItemHasBeenMarkedForReindexingEvent.php
+++ b/Classes/Event/IndexQueue/AfterIndexQueueItemHasBeenMarkedForReindexingEvent.php
@@ -27,7 +27,7 @@ final class AfterIndexQueueItemHasBeenMarkedForReindexingEvent
 {
     public function __construct(
         private readonly string $itemType,
-        private readonly int|string $itemUid,
+        private readonly int $itemUid,
         private readonly int $forcedChangeTime,
         private int $updateCount,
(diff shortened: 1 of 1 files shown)


Classes/Event/IndexQueue/AfterIndexQueueItemHasBeenMarkedForReindexingEvent.php
```diff
@@ -27,7 +27,7 @@ final class AfterIndexQueueItemHasBeenMarkedForReindexingEvent
 {
     public function __construct(
         private readonly string $itemType,
-        private readonly int|string $itemUid,
+        private readonly int $itemUid,
```
dkd-friedrich, line 30: To be more flexible and independent from the type of items to index a non-numeric item id must be supported.

This is also a preparation for indexing external data, the QueueInterface explicitly supports strings and integer values as item ids, if this is an issue in other Solr extensions we might have to reconsider the concept.

dkd-friedrich (changes requested):

Classes/Event/IndexQueue/AfterIndexQueueItemHasBeenMarkedForReindexingEvent.php
dkd-kaehm: OK,
i'll not merge this PR.

PR #3785, opened 2023-09-18
[BUGFIX] Do not index translations on default language fallbackType: free

### LINKED ISSUE #3560: [BUG] wrong indexing behaviour with fallbackType: free
**Describe the bug**
I have a problem with the correct indexing behaviour for ext:news records with fallbackType: free and 3 langs set. I have en, dk and de as languages set ( core_dk, core_de and core_en for the solr cores). But inside core_en dk, de & en news getting indexed! Expected behaviour should be that only en news are indexed. If I remove the fallbackType: free in the site config then only en news get indexed correctly inside core_en. 

**To Reproduce**

- solr base & news ts inclusion (see attached img)
- 3 news in all 3 languages (total 9 records see attached img)
- fallbackType free for dk & de
- for dk & de indexing is correct
- for en (default lang) alle records in all languages are getting indexed (see attached img). 

**Expected behavior**
for en only en news shall be indexed or found via frontend search.

**Screenshots**
sole-ddev-templates
news-list
wrong-indexing

**Used versions:**
 - Setup with https://github.com/TYPO3-Solr/solr-ddev-site
 - TYPO3 Version: 11.5.25
 - EXT:solr Version: tested with 11.5.x-dev & 11.5.1
 - ext:news: tested with v9.4 to v11.0

DDEV Database Dump:

[solr-ddev-sites-data.sql.gz](https://github.com/TYPO3-Solr/ext-solr/files/11089753/solr-ddev-sites-data.sql.gz)

schliesser:
# What this pr does

With fallbackType: free the index queue contains an entry for each available language. Without this check all the entries are additionally indexed in the default language. They should be only indexed for the language, the record has defined as sys_language_uid.
See extensive debuggin in the issue.

# How to test

Have a record based indexing configuration and change `fallbackType` to `free` in your site config. Do not set any `fallbacks`

Fixes #3560

### DIFF
--- a/Classes/Access/RootlineElementFormatException.php
+++ b/Classes/Access/RootlineElementFormatException.php
@@ -22,6 +22,4 @@ use ApacheSolrForTypo3\Solr\Exception\InvalidArgumentException;
  *
  * @author Ingo Renner <ingo@typo3.org>
  */
-class RootlineElementFormatException extends InvalidArgumentException
-{
-}
+class RootlineElementFormatException extends InvalidArgumentException {}
--- a/Classes/Backend/SettingsPreviewOnPlugins.php
+++ b/Classes/Backend/SettingsPreviewOnPlugins.php
@@ -38,8 +38,7 @@ class SettingsPreviewOnPlugins
 
     public function __construct(
         protected FlexFormService $flexFormService
-    ) {
-    }
+    ) {}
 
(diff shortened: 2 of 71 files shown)


dkd-kaehm (approved):
Thanks for the fix.

Current GH-Actions build fails, because of new rule `single_line_empty_body`,
which I'll fix soonish.

I'll merge this PR after port for release-11.5.x is available.

Issue #3800, opened 2023-09-22
[TASK] Always delegate simulated TSFE via PSR-14 events instead of Site and SiteLanguage


dkd-kaehm:
Almost all in EXT:solr 12-ALPHA introduced events propagate the Site and SiteLanguage objects, which are fetched from simulated TSFE object. That does not allow to use the required TSFE in subsequent processing directly, without instantiate TSFE objects manually.
EXT:solr* stack requires exact the same TSFE object, which is used by indexed Records/Pages, to process correctly.

**Target version: main|12 only.**

PR #3801, opened 2023-09-22
[TASK] Always delegate simulated TSFE via PSR-14 events instead of Site/SiteLanguage

dkd-kaehm:
Almost all in EXT:solr 12-ALPHA introduced events propagate the Site and SiteLanguage objects, which are fetched from simulated TSFE object. That does not allow to use the required TSFE in subsequent processing directly, without instantiate TSFE objects manually. EXT:solr* stack requires exact the same TSFE object, which is used by indexed Records/Pages, to process correctly.

This change makes it possible to reuse simulated TSFE object in listeners to be able to delegate right Core-Context objects for TYPO3 CORE components.  The Site and SiteLanguage getters are still available.

Fixes: #3800

### DIFF
--- a/Classes/Event/Indexing/AfterPageDocumentIsCreatedForIndexingEvent.php
+++ b/Classes/Event/Indexing/AfterPageDocumentIsCreatedForIndexingEvent.php
@@ -22,6 +22,7 @@ use ApacheSolrForTypo3\Solr\System\Configuration\TypoScriptConfiguration;
 use ApacheSolrForTypo3\Solr\System\Solr\Document\Document;
 use TYPO3\CMS\Core\Site\Entity\Site;
 use TYPO3\CMS\Core\Site\Entity\SiteLanguage;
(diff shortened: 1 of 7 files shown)


bmack:
Hey, I suggest to keep both Site + Site Language. They do no harm (might be in use already...) and if --- at some point --- we would remove TSFE (in TYPO3 Core), we could have some information in place already.

dkd-kaehm:
> Hey, I suggest to keep both Site + Site Language. They do no harm (might be in use already...) and if --- at some point --- we would remove TSFE (in TYPO3 Core), we could have some information in place already.

I keep currently the getters for both. 
We'll try to remove our ugly hard dependency to faked/simulated TSFE as well in the future as discussed in our sprints about detached sub-requests/processes. 
Currently we'll do that this way, and then on some point between 12 and 13, we'll remove TSFE simulation at all 
See: #3649

Issue #3798, opened 2023-09-21
[BUG] Setting results.resultsHighlighting.fragmentSize does not work


dkd-kaehm:
The setting `results.resultsHighlighting.fragmentSize` does not work on TYPO3 12 LTS.

We should check why. Possible sources:
* Changes on Apache Solr 9+
* Changes on solarium/solarium
* etc.

**Target version:** TYPO3 12 / EXT:solr main only

PR #3802, opened 2023-09-25
[BUGFIX] Fix result highlighting fragment size

dkd-friedrich:
# What this pr does

Since Apache Solr 9 the Unified Highlighter is used by default,  causing the fragment size setting to be ignored in our setup.

By configuring the Original Highlighter as default, this issue is fixed till we update our highlighting setup to use the recommended highlighter.

# How to test

1. Index contents
2. Activate the results highlighting 
`plugin.tx_solr.search.results.resultsHighlighting = 1`
3. Test the results in the frontend with different `fragmentSize` settings, e.g. `20` (fastVector) and `15` (original) 

With this bugfix the fragment size will be considered and ensured that the `fastVector` or `original` hightlighter is used.

Resolves: #3798


dkd-kaehm (approved):
LGTM.

But we should make a notice in change-log, that on 12.0.0 stable release the contents of Docker-Volume must be flushed to take effect of that fix.

PR #3807, opened 2023-09-28
[BUGFIX] Allow setting documents

georgringer:
To be in sync with old implementations it is helpful to e.g. remove the documents completly or just modify it how I like it

### DIFF
--- a/Classes/Event/Indexing/BeforeDocumentsAreIndexedEvent.php
+++ b/Classes/Event/Indexing/BeforeDocumentsAreIndexedEvent.php
@@ -29,7 +29,7 @@ use TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController;
  * Previously used with
  * $GLOBALS['TYPO3_CONF_VARS']['EXTCONF']['solr']['IndexQueueIndexer']['preAddModifyDocuments']
  */
-class BeforeDocumentsAreIndexedEvent
+final class BeforeDocumentsAreIndexedEvent
 {
     public function __construct(
         private readonly Document $document,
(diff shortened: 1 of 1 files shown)


Classes/Event/Indexing/BeforeDocumentsAreIndexedEvent.php
```diff
@@ -74,6 +74,15 @@ public function getDocuments(): array
...
     }
+    
+    /**
+     * @param Document[] $documents
+     */
+    public function setDocuments(array $documents): void
+    {
+        $this->documents = $documents;
+    }
+
```
dkd-kaehm, line 85: setter on readonly prop?

dkd-kaehm (changes requested):

Classes/Event/Indexing/BeforeDocumentsAreIndexedEvent.php
georgringer: it is $documents, not, $documents!

dkd-kaehm: > it is $documents, not, $documents!

What do you mean?

dkd-kaehm: > > it is $documents, not, $documents!
> 
> What do you mean?

Aaah OK, you mean docoment is read only not documents.

georgringer: correct! and I need to change the documents, e.g. setting it to an empty array if it shouldn't be indexed or whatsover

dkd-kaehm (approved):

Issue #3846, opened 2023-10-26
[FEATURE] Make debugging of failed page indexing easier


DanielSiepmann:
**Is your feature request related to a problem? Please describe.**
I updated to TYPO3 and solr v12 and now have a lot of `Error adding field 'created'='1681287239' msg=Invalid Date String:'1681287239'` issues.
This is easy to find for records, but pages don't reveal the issue.

**Describe the solution you'd like**
Better error logging in order to actually see the error without extra debugging efforts.

**Describe alternatives you've considered**
I used xdebug to retrieve the solr response containing the error.

**Additional context**

**Target versions**
next patch release for latest major version.

dkd-kaehm:
@DanielSiepmann 
is #2590 the same and can be closed?

DanielSiepmann:
I guess so. Didn't had issues for a long time.

Issue #2590, opened 2020-04-30
[FEATURE] Provide useful error message on page index failure


DanielSiepmann:
**Is your feature request related to a problem? Please describe.**
Whenever indexing via PageIndexer failes, an error is logged. This error basically just says "it didn't work" with no explanation.
This causes frustration and extra efforts and time, leading to costs.

**Describe the solution you'd like**
It would be cool if the error causing the indexing to fail, would be logged as the cause.
Maybe it is possible to extract the error from the response, e.g. in case of 500 HTTP Status.

**Describe alternatives you've considered**
Right now I'll disable the authentication feature in `ApacheSolrForTypo3\Solr\IndexQueue\PageIndexerRequest`. Then I extract the requested url + http headers for indexing from log, and execute a curl on command line. I redirect the result to a `test.html` and open it with the browser.

PR #3847, opened 2023-10-26
[TASK] Log solr response in case of Page indexing issue

DanielSiepmann:
Logging the response eases debugging.
Integrators and developers can check the response in order to retrieve the actual reported issue, a crucial info in order to fix indexing issues.

# What this pr does

Add an error log entry with solr response body in case indexing a page did not succeed.

# How to test

Create a broken page indexing and check the log for the new entry.

Fixes: #3846, #2590


dkd-kaehm (approved):
@DanielSiepmann 
Thanks for the fix.

Issue #3851, opened 2023-10-30
TypeError TYPO3\CMS\Extbase\Mvc\Web\Routing\UriBuilder::setRequest(): Argument #1 ($request) must be of type TYPO3\CMS\Extbase\Mvc\RequestInterface, TYPO3\CMS\Core\Http\ServerRequest given, called in Classes/ViewHelpers/Uri/AbstractUriViewHelper.php on line 57


DanielSiepmann:
**Describe the bug**
EXT:solr fails in considering none Extbase Environments when building URIs for some facets.

**To Reproduce**
1. Create facets outside of Extbase context (There is no native way provided by EXT:solr)
2. Create links to facet items
3. See error

**Expected behavior**
No error is shown, but URIs are generated in expected namespace.

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.7
 - Browser: Firefox
 - EXT:solr Version: 12.0.0
 - Used Apache Solr Version:  9.3.0
 - PHP Version: 8.1.25
 - MySQL Version: 8.0.34-0ubuntu0.20.04.1

**Additional context**
We created custom content elements (FLUIDTEMPLATE insteaf of USER/EXTBASE) which will render filter for executed search results.
Those now break with update to TYPO3 and EXT:solr v12.

I'll provide a patch that solves the issue for us right now. But I don't promise that it won't break other stuff.
<details>
  <summary>Readable Error</summary>

```
(1/1) TypeError

TYPO3\CMS\Extbase\Mvc\Web\Routing\UriBuilder::setRequest(): Argument #1 ($request) must be of type TYPO3\CMS\Extbase\Mvc\RequestInterface, TYPO3\CMS\Core\Http\ServerRequest given, called in solr/Classes/ViewHelpers/Uri/AbstractUriViewHelper.php on line 57
```

</details>

PR #3852, opened 2023-10-30
[BUGFIX] Provide proper Uri Builder Request in VH

DanielSiepmann:
Checking RenderingContext is not enough as it might provide either a ServerRequest or Extbase Request.
We use the RequestBuilder to ensure there is a proper Extbase Request.

A none Extbase Request might exist in case of usage within StandaloneView or other contexts without providing an Extbase Request.

Fixes: #3851

# What this pr does

Ensure that an Extbase request is provided.

# How to test

I don't have an easy step by step way … Maybe test that this won't break anything instead?

### DIFF
--- a/Classes/ViewHelpers/Uri/AbstractUriViewHelper.php
+++ b/Classes/ViewHelpers/Uri/AbstractUriViewHelper.php
@@ -23,6 +23,7 @@ use ApacheSolrForTypo3\Solr\Domain\Search\Uri\SearchUriBuilder;
 use ApacheSolrForTypo3\Solr\Exception\InvalidArgumentException;
 use ApacheSolrForTypo3\Solr\ViewHelpers\AbstractSolrFrontendViewHelper;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Mvc\Web\RequestBuilder;
 use TYPO3\CMS\Extbase\Mvc\Web\Routing\UriBuilder;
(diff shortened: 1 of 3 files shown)


dkd-kaehm:
@DanielSiepmann 
Integration tests are broken. Please check them.

Classes/ViewHelpers/Uri/AbstractUriViewHelper.php
```diff
@@ -40,21 +41,31 @@ abstract class AbstractUriViewHelper extends AbstractSolrFrontendViewHelper
...
         }
+        if (!isset(self::$requestBuilder)) {
+            self::$requestBuilder = GeneralUtility::makeInstance(RequestBuilder::class);
+        }
 
         if ($renderingContext instanceof RenderingContext) {
             $uriBuilder = GeneralUtility::makeInstance(UriBuilder::class);
-            /** @phpstan-ignore-next-line */
-            $uriBuilder->reset()->setRequest($renderingContext->getRequest());
+            $request = self::$requestBuilder->build($renderingContext->getRequest());
```
DanielSiepmann, line 67: Maybe we should check whether the Request of Rendering Context already is an extbase request and keep original logic in that case. Only use new one in case we didn't receive an extbase request?

dkd-kaehm: Its ok so.

dkd-kaehm (approved):
Thanks for fixing that issue.
:+1:

Issue #3856, opened 2023-11-01
[BUG] indexing not possible if enableRecursiveValueResolution and empty foreignLabelField 


clickstorm:
**Describe the bug**
If enableRecursiveValueResolution and the record has no relation, then the item can not be indexed.

**To Reproduce**

```
plugin.tx_solr.index.queue ...
... {
foreignLabelField =child.title
localField = ref_areas
enableRecursiveValueResolution = 1
multiValue = 1
}
```
1. Index items
2. Click on error log
4. Open an item 
5. See error 
```
1476107295: TYPO3\CMS\Core\Error\Exception: PHP Warning: Undefined array key "foreignLabelField" in /var/www/html/projekte/zoo_leipzig/website/site-12/vendor/apache-solr-for-typo3/solr/Classes/ContentObject/Relation.php line 173 in /var/www/html/projekte/zoo_leipzig/website/site-12/vendor/typo3/cms-core/Classes/Error/ErrorHandler.php:138
Stack trace:
```

**Expected behavior**
Prevent the error. We will provide a Pull Request.

**Screenshots**
![01-11-_2023_16-48-05](https://github.com/TYPO3-Solr/ext-solr/assets/9073453/39450fdd-52c3-43a2-ba3f-868668990cd0)

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.7
 - Browser: chrome
 - EXT:solr Version: 12.0.0
 - PHP Version: 8.2

PR #3857, opened 2023-11-01
[BUGFIX] prevent empty array key if foreignLabelField is null

clickstorm:
# What this pr does

If you enableRecursiveValueResolution and the foreignLabelField is not set because the record has no relation, the item can not be indexed.

# How to test

See Ticket.

Fixes: #3856


dkd-kaehm (approved):
Thanks for fixing that issue.

Issue #3869, opened 2023-11-08
[TASK] Document `grouping.groups.[groupName].sortBy`


julianhofmann:
The `sortBy` in group configuration is supported but was *not* documented.  You'll need it, especially when using `query.sortBy` and grouping.

PR #3870, opened 2023-11-08
Add `grouping.groups.[groupName].sortBy`

julianhofmann:
The `sortBy` in group configuration is supported but was *not* documented.  Especially when using `query.sortBy` and grouping, you'll need it.

Fixes: #3869


dkd-kaehm (approved):
Thanks for fixing the docs.

Issue #3871, opened 2023-11-08
[TASK] Fix TS paths in documentation


julianhofmann:
For `grouping.groups.[groupName].*` there is the `groups`-key missing in "TS Path" of documentation.

PR #3872, opened 2023-11-08
[TASK] Add missing groups-key to TS Path

julianhofmann:
For `grouping.groups.[groupName].*` there was the groups-key missing in "TS Path" of documentation.

Fixes: #3871


dkd-kaehm (approved):
Thanks for fixing the docs.

Issue #3873, opened 2023-11-09
[BUG] Infinite loop in SolrRoutingMiddleware


vbdetemplephilipp:
The function \ApacheSolrForTypo3\Solr\Middleware\SolrRoutingMiddleware::retrievePageInformation contains a do-while loop.
For specific frontend requests, this loop runs infinitely.

use old school requests like /index.php?id=123

**Expected behavior**
The loop should be exited if no page could be found based on the request path, should scan for page ids or middleware should be executed after redirect to speaking url.

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.32
 - Browser: any
 - EXT:solr Version: 11.5.4
 - PHP Version: 8.2.12

PR #3878, opened 2023-11-14
[BUGFIX:11.5]  Infinite loop in SolrRoutingMiddleware 

superyaro:
# What this pr does

Fixes infinite loop and therefore server crashing if route doesn't exist or URl is in the scheme like domain/id=123456

# How to test

use old school requests like /index.php?id=123 or any URI which will not be found in the slug Provider class (any none existing URI with speaking URL)

Fixes: #3873


dkd-kaehm (approved):
Thanks for the fix.

dkd-kaehm (approved):

tehplague:
The fix also covers an issue with the suggest controller if invoked with an old-style URIs and without a `queryString` argument (see https://typo3.slack.com/archives/C02FF05Q4/p1700048519979689).

PR #3893, opened 2023-11-21
[TASK] Remove unused field `cookie` in tx_solr_statistics

derhansen:
This PR removes the unused field `cookie` from the tx_solr_statistics table. The field has never been used in the extension and has most likely been introduced by accident since the initial statistics table structure partly seems to be similar to ext:indexedsearch statistics table (see https://github.com/TYPO3/typo3/blob/TYPO3_4-5/typo3/sysext/indexed_search/ext_tables.sql#L107C2-L107C2).


ohader (approved):

ohader (approved):
Looks good to me! 👍

saschanowak (approved):

darth-hader (approved):

dkd-kaehm:
@derhansen 
Thanks for your work. The feature for "users" last searches is gone on some point of EXT:solr.
There are still multiple remains on that feature in docs and in our code, like 

https://github.com/TYPO3-Solr/ext-solr/blob/635883f6d65ab2b8ab4ed56b4b501f73d074817c/Documentation/Configuration/Reference/TxSolrSearch.rst?plain=1#L608-L617

Do we really want to remove this feature from EXT:solr like EXT:indexed_search did: 
* https://forge.typo3.org/issues/92993 
* https://github.com/TYPO3/typo3/commit/d774505326b66697a1818ddac154332017f2d53d

**May be restoring the functionality is better solution?**
I can not find any valid argument to remove this nice feature except unexplainable wiping from EXT:indexed_search.
IMHO the EXT:indexed_search should get that feature back as well.

dkd-kaehm:
The "user" mode works currently without that column. It stores the searches in users session.
So the pull request will be merged and back-ported to ELTS Versions as well.

Issue #3897, opened 2023-11-28
[BUG] open_basedir Issue with IndexQueueModule.css


DanielSiepmann:
**Describe the bug**
The backend module for queue does not load the css file in a proper way.

**To Reproduce**
Steps to reproduce the behaviour:
1. Enable open_basedir restriction
2. Go to the backend module for queues
3. Check the error log and check loaded CSS files.

**Expected behavior**
No error should be shown, the css file should be loaded with a timestamp, if `$GLOBALS['TYPO3_CONF_VARS']['BE']['versionNumberInFilename']` is configured.

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.33
 - Browser: Firefox
 - EXT:solr Version: 11.5.3
 - Used Apache Solr Version: xxx
 - PHP Version: 8.1.25
 - MySQL Version: xxx

**Additional context**
The ViewHelper in use has wrong documentation. One should use EXT: syntax in order to allow proper handling by TYPO3 which will properly add timestamp if configured and passes open_basedir checks.

I'll provide a PR soon.

PR #3898, opened 2023-11-28
[BUGFIX] Broken IndexQueueModule.css asset path in backend

DanielSiepmann:
The path should use EXT: syntax.
That way TYPO3 can properly handle the file and add version numbers and check for existence.

# How to test

Open the queue backend module and check that the css file has a proper timestamp as parameter when setting `$GLOBALS['TYPO3_CONF_VARS']['BE']['versionNumberInFilename']` to `0`

Fixes: #3897


DanielSiepmann:
I've tested in TYPO3 11.5.33, please test with v12 as well and maybe backport?

Issue #3900, opened 2023-12-01
[BUG] I found a problem with indexing field with value 0


taieb123:
**Describe the bug**
I have a field in the queue configuration with _boolS the value inside the field is 1 or 0 as boolean, the problem is the field was not added to the document when the field value is 0, 

The problem inside `ApacheSolrForTypo3\Solr\IndexQueue\AbstractIndexer` with the function `addDocumentFieldsFromTyposcript` the problem comes from the condition !empty()

**Expected behavior**
It should also add the document to the value when the value field is 0

**Screenshots**
![image](https://github.com/TYPO3-Solr/ext-solr/assets/152389963/b603cc81-2c27-4357-90ac-4c70ade561d0)

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.8
 - EXT:solr Version: 12.0.1
 - Used Apache Solr Version: 9.3.1
 - PHP Version: 8.2.0
 - MySQL Version:  8.0.0

taieb-rekik-cr:
@dkd-kaehm 
Thank you for the update, yes working fine for me.

PR #3901, opened 2023-12-03
[BUGFIX] false as field value is not allowed

dkd-kaehm:
`AbstractIndexer::addDocumentFieldsFromTyposcript()` disallows any PHPs empty value. This change adds explicit checks for disallowed values:
* null
* empty string
* empty array

Fixes: #3900


mr-typo3:
I would like to see this change reversed! "FALSE" is also a sensible value that should be saved.

dkd-kaehm:
> I would like to see this change reversed! "FALSE" is also a sensible value that should be saved.

@mr-typo3 
I don't understand what you mean.
That bugfix must not be reversed, because it explicitly allows the `false` value in fields but not empty values.

PR #3908, opened 2023-12-05
[BUGFIX] Resolve custom partialName in hierarchy facet

tillhoerner:


dkd-kaehm (approved):
Thanks for fixing that issue.

Issue #3920, opened 2023-12-13
[BUG] AfterFacetIsParsedEvent is never dispatched


dkd-kaehm:
**Describe the bug**
The event AfterFacetIsParsedEvent is never dispatched, because the `\ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\Facets\OptionBased\Options\OptionsFacetParser::injectEventDispatcher()` is not called, because the Parser classes are not public and not autowired.

**To Reproduce**
1. Setup some option based facet and make sure it is rendered in search plugin
2. Set xDebug Breakpoint to `AfterFacetIsParsedEvent::__construct()`
3. trigger the search, which renders tht facet

**Expected behavior**
* The xDebug MUST hold on `AfterFacetIsParsedEvent::__construct()`
* OR The registered `AfterFacetIsParsedEvent`-`EventListener` MUST react on that event.

**Screenshots**

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.9
 - EXT:solr Version: dev-main ~12.0.1 on 966a6ab8a637847819c3df20d8e1756c30aeaebb

**Additional context**

Besides that, the `AfterFacetIsParsedEvent` is triggered for Options Facet type.

PR #3921, opened 2023-12-13
[BUGFIX] `AfterFacetIsParsedEvent` was never dispatched

dkd-kaehm:
The `AfterFacetIsParsedEvent` was newer dispatched, because the `::injectEventDispatcher()`  was never called due of missing dependency injection settings. This change sets almost all classes within `ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\Facets\` namespace  to autowire and public DI. 
The non-wireable classes are excluded from autowire and public DI and should behave the same as before this change.

**In addition the `AfterFacetIsParsedEvent` is added to all available facet types.**

Fixes: #3920

---

## Mainteners Notes:

We want the integration tests to check if the events are fired properly.

### DIFF
--- a/Classes/Domain/Search/ResultSet/Facets/AbstractFacetItem.php
+++ b/Classes/Domain/Search/ResultSet/Facets/AbstractFacetItem.php
@@ -48,6 +48,11 @@ abstract class AbstractFacetItem
         return $this->label;
     }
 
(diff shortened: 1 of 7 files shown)


dkd-friedrich (approved):
LGTM

Issue #3566, opened 2023-03-30
[BUG] Subpage of page with restricted access and extendToSubpages set is removed from index queue


timhorstmann:
Scenario:

- rootpage
  - page A (with restricted access, fe_group is selected and extendToSubpages is set)
    - page B (without any access restrictions set)

When the index queue is build for the first time, both pages A+B get indexed and can be found (with proper FE access rights).
If page B gets updated, it is instantly removed from index queue and also from solr index and therefore can not be found anymore.

There seems to be no way to get page B indexed again than rebuilding the index queue. But with any update to page B it is removed again, so this can not be valid workaround.

I investigated a little and found the responsible lines of code:

In [FrontendEnvironment/Tsfe.php](https://github.com/TYPO3-Solr/ext-solr/blob/main/Classes/FrontendEnvironment/Tsfe.php#L147) the necessary usergroups are fetched only from the current site. These groups are later used for building a tsfe and checking the access to the page. This fails in the described scenario with a page-not-found-response.

This bug is a variation of [Issue 3563](https://github.com/TYPO3-Solr/ext-solr/issues/3563), although the problem here might be solved easier. From my point of view it should be sufficient to collect the usergroups from the rootline until rootpage or a page with extendToSubpages set.

Used versions:

TYPO3 Version: 11.5.25
Browser: firefox
EXT:solr Version: 11.5.1
Used Apache Solr Version: 8.11.1
PHP Version: 8.1
MariaDB: 10.5

timhorstmann:
If this bug can be confirmed and the described solution is valid, I could try to fix it.

dkd-kaehm:
@timhorstmann the 3563 seems to be wrong in description.
Please change it to the right one.

dkd-kaehm:
Yes, it did never work.

timhorstmann:
> @timhorstmann the 3563 seems to be wrong in description. Please change it to the right one.

You're right, I updated to 3271

georgringer:
I run into similiar issues but I don't get page B to be indexed - no matter if on 1st run or later on :(

dkd-kaehm:
> I run into similiar issues but I don't get page B to be indexed - no matter if on 1st run or later on :(

I guess not a run, but initializing the index queue for pages.

kitzberger:
Issue still exists for TYPO3 12 and EXT:solr 12.x. Page B isn't added to the `tx_solr_indexqueue_item` table, no matter what I do.

PR #3926, opened 2024-01-11
[BUGFIX] Issue #3566: Check rootline for feGroups from page with extendToSubpages set

timhorstmann:
# What this pr does

When fetching usergroups to build a TSFE, the fe_groups field from current page could be empty. But there might be a page in the rootline with feGroups and  extendToSubpages set.
 
This PR collects the feGroups from the rootline in this case.
The code is inspired from [TYPO3\CMS\Backend\Utility\BackendUtility::ADMCMD_previewCmds](https://github.com/TYPO3/typo3/blob/bedee37f92616a0d21f073d8049b32befe5fe512/typo3/sysext/backend/Classes/Utility/BackendUtility.php#L3648)

# How to test

See https://github.com/TYPO3-Solr/ext-solr/issues/3566

Fixes: https://github.com/TYPO3-Solr/ext-solr/issues/3566


Classes/FrontendEnvironment/Tsfe.php
```diff
@@ -144,6 +144,20 @@ protected function initializeTsfe(int $pageId, int $language = 0, ?int $rootPage
                 $userGroups = explode(',', $pageRecord['fe_group']);
             } else {
                 $userGroups = [0, -1];
+
+                // Check rootline for feGroups from page with extendToSubpages set
+                $rootline = BackendUtility::BEgetRootLine((int)($pageInfo['uid'] ?? 0));
```
dkd-kaehm, line 149: Variable `$pageInfo` is never defined.
Please check that issue.

dkd-kaehm (changes requested):
@timhorstmann 
Please check the implementation.
There is trouble with `$pageInfo` variable which is not defined.

Classes/FrontendEnvironment/Tsfe.php
dkd-kaehm, line 149:
```suggestion
                $rootline = BackendUtility::BEgetRootLine((int)($pageRecord['uid'] ?? 0));
```

timhorstmann:
@dkd-kaehm Yes. Thank you for noticing. I updated the PR.

timhorstmann:
This does not work as aspected.
$pageRecord['uid'] is not present. If i fix this, the page is not removed from the index queue, but still can't be found. I will update the PR.

dkd-kaehm:
> I'm a little confused by the last comment. Is there another TODO for me?

You mean following?:
> Please check the implementation.
> There is trouble with $pageInfo variable which is not defined.

If so, you has already answered and fixed it already.

Issue #3928, opened 2024-01-16
[BUG] TypeError in getRootPageId method


cepheiVV:
**Describe the bug**
PHP Type Error of method return value.
The method must return an integer.
`\ApacheSolrForTypo3\Solr\System\Page::getRootPageId()`
But the returned value of `$page['uid']` is a string.

`$rootPageId = $page['uid'];` should be changed to `$rootPageId = (int)$page['uid'];`.

**Expected behavior**
The method `\ApacheSolrForTypo3\Solr\System\Page::getRootPageId()` return the correct type.

**Screenshots**
<img width="1083" alt="Screenshot 2024-01-16 at 12 08 27" src="https://github.com/TYPO3-Solr/ext-solr/assets/2963482/3c4da1d8-7e5e-4427-a1c3-4d55bdb58550">

**Used versions (please complete the following information):**
 - TYPO3 Version: 	11.5.33
 - EXT:solr Version: 11.5.5
 - PHP Version: 8.1.26

PR #3929, opened 2024-01-16
[BUGFIX] Ensure method return value of root-page-UID is an integer

dkd-kaehm:
Ensure method return value of root-page-UID is an integer `System\Page::getRootPageId()`

Fixes: #3928


dkd-kaehm:
Thanks @cepheiVV
But the pull requests shouldn't go against fork.
You can close the https://github.com/cepheiVV/ext-solr/pull/1

PR #3930, opened 2024-01-17
[DOC] Add infos about adding own field processors

kitzberger:
# What this pr does

Update the wrong section on that it's still impossible to extend the file processors


dkd-kaehm (approved):
Nice Catch.
Thanks for extending the docs.

PR #3938, opened 2024-01-24
[BUGFIX] Add record to index queue when published

### LINKED ISSUE #3718: [BUG] Publishing records does not fire VersionSwappedEvent
**Describe the bug**
In `typo3/public/typo3conf/ext/solr/Classes/IndexQueue/RecordMonitor.php`, the method `processCmdmap_postProcess` (line 99) checks:
`        if ($command === 'version' && $value['action'] === 'swap') {`
However,  `$value['action'] `now seems to be `'publish'`. Therefore, the VersionSwappedEvent is not fired and the record is not written into the index queue.

**To Reproduce**
1. CREATE any record (e.g. type pages)
2. Publish the record via the workspace (send to LIVE workspace)
3. Note the record's uid <UID>
3. Check the database for `select * from tx_solr_indexqueue_item where item_uid=<UID>;`
4. The published record <UID> is not there
5. Therefore, when executing the "Index Queue Worker" task in the scheduler,  the record is not sent to the solr core.

**Expected behavior**
After publishing a new record, the record's uid is written into the table `tx_solr_indexqueue_item`.

 - TYPO3 Version: 11.5.28
 - Browser: Firefox 114.0.2 (64-bit)
 - EXT:solr Version: 11.5.3
 - Used Apache Solr Version: 8.11.2
 - PHP Version: 7.4.3-apache-buster
 - MySQL Version: 5.7.31

**Additional context**
This seems connected to https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Breaking-92206-RemoveWorkspaceSwappingOfElements.html

spoonerWeb:
# What this pr does

Checks if action is "publish" and add record to index queue.
Removed "swap" action as it's removed with https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Breaking-92206-RemoveWorkspaceSwappingOfElements.html.

# How to test

1. Create a workspace
2. Switch to workspace and create page
3. Page is not in index queue -> correct
4. Publish the page to LIVE workspace
5. Page is not in index queue -> incorrect

Fixes:  #3718

### DIFF
--- a/Classes/IndexQueue/RecordMonitor.php
+++ b/Classes/IndexQueue/RecordMonitor.php
@@ -99,7 +99,7 @@ class RecordMonitor
 
         // track publish / swap events for records (workspace support)
         // command "version"
-        if ($command === 'version' && $value['action'] === 'swap') {
+        if ($command === 'version' && $value['action'] === 'publish') {
             $this->eventDispatcher->dispatch(
                 new VersionSwappedEvent($uid, $table),
             );
--- a/Classes/IndexQueue/RecordMonitor.php
+++ b/Classes/IndexQueue/RecordMonitor.php
@@ -99,7 +99,7 @@ class RecordMonitor
 
         // track publish / swap events for records (workspace support)
         // command "version"
-        if ($command === 'version' && $value['action'] === 'publish') {
+        if ($command === 'version' && ($value['action'] === 'publish' || $value['action'] === 'swap')) {
             $this->eventDispatcher->dispatch(
                 new VersionSwappedEvent($uid, $table),
             );


dkd-kaehm (approved):
Thanks for the fix.
We'll backport it to 11.5 and then merge the both on same time.

@spoonerWeb Could you please sqaush the commits?

dkd-friedrich:
I'm not sure if we can drop 'swap' completely, EXT:workspaces is internally still handling swap actions and even building command maps with 'swap', e.g.

- https://github.com/TYPO3/typo3/blob/main/typo3/sysext/workspaces/Classes/Service/WorkspaceService.php#L168
- https://github.com/TYPO3/typo3/blob/main/typo3/sysext/workspaces/Classes/Hook/DataHandlerHook.php#L109

I think we should use both options to be on the safe side

dkd-kaehm:
@spoonerWeb 

Could you please check https://github.com/TYPO3-Solr/ext-solr/pull/3938#issuecomment-1960861595

etnicoliver:
> As far as I understand it, "swap" has been removed? "publish" is sufficient, no?
> 
> https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Breaking-92206-RemoveWorkspaceSwappingOfElements.html

That's what I understood. I tested this patch and it seems to work; it's just not the same line in version 13.0.3: https://patch-diff.githubusercontent.com/raw/TYPO3-Solr/ext-solr/pull/3938.patch

Classes/IndexQueue/RecordMonitor.php
```diff
@@ -98,7 +98,7 @@ public function processCmdmap_postProcess(
 
         // track publish / swap events for records (workspace support)
         // command "version"
-        if ($command === 'version' && $value['action'] === 'swap') {
+        if ($command === 'version' && $value['action'] === 'publish') {
```
dkd-kaehm:
```suggestion
        if ($command === 'version' && ($value['action'] === 'publish' || $value['action'] === 'swap')) {
```


The command map for 'swap' is still present on TYPO3 13.4.19, 
See mention on https://github.com/TYPO3-Solr/ext-solr/pull/3938#issuecomment-1960861595:

https://github.com/TYPO3/typo3/blob/v13.4.19/typo3/sysext/workspaces/Classes/Service/WorkspaceService.php#L167

What does it mean for EXT:solr exactly? Does it break some functionality of EXT:solr?
May be just leave 'swap' check there as well?

dkd-kaehm (changes requested):
The tests must be adjusted.

Issue #3940, opened 2024-01-31
[BUG] Function getTagContent should be case insensitive


thomashohn:
**Describe the bug**
The method getTagContent in class HtmlContentExtractor - should also for work for tags that are not lower case

**Used versions (please complete the following information):**
 - TYPO3 Version:11.5.34
 - Browser: Not relevant
 - EXT:solr Version: 11.5.5/12.x/main
 - Used Apache Solr Version: Not relevant
 - PHP Version: 8.2.*
 - MySQL Version: 8.*

PR #3941, opened 2024-01-31
[BUGFIX] Handle if some tags are upper-case and strict-comparison

thomashohn:
# What this pr does

Make take content case-in sensitive to running it by strtolower

# How to test

Have content like <I>foo</i>  in the DB

Fixes: #3940


dkd-friedrich (approved):
LGTM

Issue #3955, opened 2024-02-21
[TASK] Allow several tested Solr versions


dkd-friedrich:
**What should be done in the scope of this task?**

As incompatibilities can also occur in minor Solr version updates, all versions should be explicitly tested and added to a list of supported versions.

E.g. minor version 8.11.3 includes a breaking change to fix a security issue, but EXT:solr relies on stream body and won't work with 8.11.3 out of the box.

To allow customers to recognize supported versions, a corresponding list of tested versions is to be introduced.

PR #3956, opened 2024-02-21
[TASK] Support several Apache Solr versions

dkd-friedrich:
As incompatibilities can also occur in minor versions, all versions will be explicitly tested and listed.

Currently tested and supported versions are: 9.3.0, 9.4.0, 9.4.1, 9.5.0.

It is always recommended to use the latest tested version.

Resolves: #3955


dkd-kaehm (approved):
Top!
LGTM.

Issue #3975, opened 2024-03-05
[FEATURE] Allow to configure value of resolved relation while indexing


DanielSiepmann:
**Is your feature request related to a problem? Please describe.**
We have a relation to a table where we don't have a single field representing the value, but a combination of fields.

**Describe the solution you'd like**
Something like the following TypoScript:
```
authorsNames_stringM = SOLR_RELATION
authorsNames_stringM {
    localField = author
    foreignLabel.dataWrap = {field : first_name} {field : last_name}
    multiValue = 1
}
```
The new feature is `foreignLabel` with applied stdWrap. That way it is possible to build values however you like. Using userFunc, data from the resolved record, etc.

**Describe alternatives you've considered**
I digged into the code and it is very hard limited to resolving a single field.
So one could use dataHandler hooks and fill a specific field only for resolving.

**Additional context**

**Target versions**
12

jonnsn:
If I got the problem right, one could use COA for that. We use something like the following - but for a singlevalue (_stringS) field. Not sure if one could make a multivalue from it.
`
authorsNames_stringS = SOLR_CONTENT
authorsNames_stringS {
    cObject = COA 
    cObject {
        5 = SOLR_RELATION
        5 {
		localField = author
		foreignLabelField = first_name
		singleValueGlue = | |
	}
	#spacer
	10 = TEXT
	10.noTrimWrap = || |

	15 = SOLR_RELATION
	15 {
		localField = author
		foreignLabelField = last_name
		singleValueGlue = | |
	}                
        localField = author
        foreignLabel.dataWrap = {field : first_name} {field : last_name}
        multiValue = 1
    }
}
`
Keep in mind that the index will contain all first_names followed by all last_names. So not possible to exactly match <first_name last_name>. And not very useful if one would want to output the values directly from solr.

Maybe it helps someone.

Regards

PR #3976, opened 2024-03-05
[FEATURE] Support stdWrap to resolve relation label field

DanielSiepmann:
# What this pr does

Adds a new TypoScript option to allow stdWrap usages to build the actual label for indexing when resolving relations via SOLR_RELATION.

# How to test

Index a field using SOLR_RELATION with the new option. E.g. try EXT:tt_address and use the provided example.

Fixes: #3975


dkd-kaehm (approved):
Thanks for that feature.

Issue #3997, opened 2024-04-05
[BUG] PHP Warning: Undefined array key "label" if the label of a sorting field only consists of a stdWrap


serens:
**Describe the bug**
If a sorting field is configured so that its label is only created via a stdWrap, then the TYPO3 log will be flooded with a PHP warning.

**To Reproduce**

Create a simple sorting field with a label using stdWrap:

```
plugin.tx_solr {
  search {
    sorting = 1
    sorting {
      options >
      options {
        isnew {
          label.data = LLL:EXT:extension/Resources/Private/Language/locallang.xlf:solr.sorting.isnew
          field = sort_isnew_boolS
        }
      }
    }
  }
}
```

Trigger a search in the frontend.

**Expected behavior**
No PHP Warning 'PHP Warning: Undefined array key "label" in /vendor/apache-solr-for-typo3/solr/Classes/Domain/Search/ResultSet/ResultSetReconstitutionProcessor.php line 90'.

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.11
 - Browser: Chrome
 - EXT:solr Version: 12.0.2
 - Used Apache Solr Version: 
 - PHP Version: 8.2.17
 - MySQL Version: 8.0.32

**Additional context**
In `\ApacheSolrForTypo3\Solr\Domain\Search\ResultSet\ResultSetReconstitutionProcessor::parseSortingIntoObjects` the code assumes that key "label" will always be set.

PR #3998, opened 2024-04-05
[BUGFIX] Solves PHP Warning: Undefined array key "label".

serens:
If the label of a sorting field only consists of a stdWrap, the key “label” is not present, only “label.”.

# What this pr does

Removes possible flooded TYPO3 log messages.

# How to test

Create a simple sorting field with a label using stdWrap.

```
plugin.tx_solr {
  search {
    sorting = 1
    sorting {
      options >
      options {
        isnew {
          label.data = LLL:EXT:extension/Resources/Private/Language/locallang.xlf:solr.sorting.isnew
          field = sort_isnew_boolS
        }
      }
    }
  }
}
```

Fixes: #3997


dkd-kaehm (approved):
Thanks for this fix.

Issue #4004, opened 2024-04-12
[BUG] Can't submit form with empty autosuggest input field


felixranesberger:
`suggest_controller.js` adds an autofocus on form submit to all forms with an autosuggest input containing `.tx-solr-suggest`. We have a form that contains an autosuggest input field and multiple other facets to filter for values. If you leave the search input empty and add activate some other facet filters, you can't submit the form.

PR #4005, opened 2024-04-12
[BUGFIX] Improve autosuggest_controller focus behavior

felixranesberger:
# What this pr does

`suggest_controller.js` adds an autofocus on form submit to all forms with an autosuggest input containing `.tx-solr-suggest`. We have a form that contains an autosuggest input field and multiple other facets to filter for values. If you leave the search input empty and add activate some other facet filters, you can't submit the form. 

This pull request changes the submit event listener, to check for `.tx-solr-suggest-focus` instead of `.tx-solr-suggest` which is more intuitive, as it's already used for the initial focus made by `$form.find('.tx-solr-suggest-focus').focus();`.

# How to test

Create an autosuggest searchfield and remove the `.tx-solr-suggest-focus` CSS class.

Fixes: #4004

### DIFF
--- a/Resources/Public/JavaScript/suggest_controller.js
+++ b/Resources/Public/JavaScript/suggest_controller.js
@@ -11,19 +11,19 @@ function SuggestController() {
                 $formAutoComplete = $('body');
             }
 
-            $form.find('.tx-solr-suggest-focus').focus();
-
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
Thanks for that fix.

Issue #4003, opened 2024-04-11
[TASK] Be able to disable tracking of visitors search inputs


christophlehmann:
**What should be done in the scope of this task?**

Tracking of user search input seems enabled by default. This

* floods the tables fe_sessions/tx_solr_last_searches
* starts sessions and thus sends session cookies.

There should ben an easy possibility to disable it.

Offtopic: Search inputs are only tracked, when they provide results.

Current TypoScript

```
plugin.tx_solr {
  search {
    lastSearches = 0
    lastSearches {
      limit = 10
      // tracking mode "user" or "global"
      mode = user
    }
  }
}
```

`lastSearches = 0` and `mode = ''` or `limit = 0` could be candidates to achieve disabling the tracking. What do you think?

dkd-kaehm:
Yes, that is good idea.
Would propose `mode = disabled` to be more self explaining notation.

christophlehmann:
And change it to be new default?

dkd-kaehm:
For main in separate commit please.
It would be nice to backport that feature in non-Breaking manner.

PR #4006, opened 2024-04-14
Be able to disable tracking of last searches

christophlehmann:
# What this pr does

It introduces `plugin.tx_solr.search.lastSearches.mode = disabled` and in the 2nd commit set it to be new default

# How to test

* Set the TypoScript.
* Make a search, that returns a result
* See no tracked keywords in database 

Fixes: #4003


dkd-kaehm (approved):
@christophlehmann 
Thanks for that feature we'll merge that soonish.

PR #4009, opened 2024-04-16
[BUGFIX] Fix debug output if response body is read once stream

Kanti:
# Problem:

We had the problem that the requests where cached by staticfilecache.
But we could not find any hint about that in the logs. 

Because the log would only show an empty body.

That is because the body was already read in the lines above the log message.
And it was a read once stream. (read more about read once streams: https://stackoverflow.com/a/6518288/5440709 )

# Solution:

Our solution is to keep the body in a variable so if there is a Problem we can send the body to the log message.

@dkd-kaehm


dkd-kaehm:
@Kanti 
Thanks for the fix, but what does it exactly?
Please add the description properly.

PR #3996, opened 2024-04-05
[TASK:P:main] Full fixed french translation

dkd-kaehm:
Fixes: #3983
Ports: #3984 

@megamisan please review this port

PR #3984, opened 2024-03-12
[TASK] Full fixed french translation

megamisan:
# What this pr does

Adds missing translations for French and normalize some terms likes "Index Queue" to "File d'indexation" instead of "File d'attente de l'index" or "File d'attente d'indexation".

Also fixes some other small mistakes.

# How to test

Using a clean setup (no var/labels folder), activate and download French for backend. Change your user to French. Check that pages.no_search_sub_entries field has its label and description translated. In the module bar, check module names are in French. In about module, check the module and submodules information are translated. Explore the SOLR module and find no word in English. If found, the text is not found in translation files.

Fixes: #3983

### Maintainers notes:

- [x] port into main branch

Issue #3983, opened 2024-03-12
[TASK] Complete French translation and check existing translations


megamisan:
It is expected to:

- Have all translation files translated to French
- Fix internal glossary translation
- Check and fix existing translations

PR #4032, opened 2024-05-08
[FIX] Typos in last FR translation PR

megamisan:
# What this pr does

Fixed two typos I introduced in PR #3984 related to #3983.

In #3996 which is the port to main not yet merged, it should be fixed too.


dkd-kaehm (approved):
@megamisan 
Thanks for fixing that.

Issue #4043, opened 2024-05-16
[TASK] Eliminate deprecations in EXT:solr backend modules


dkd-friedrich:
**What should be done in the scope of this task?**

Make EXT:solr backend modules compatible with TYPO3 12 and TYPO3 13, eliminate deprecations.

sfroemkenjw:
PR #4054
PR #4049
PR #4045

PR #4054, opened 2024-05-16
[TASK] Use new template module API

sfroemkenjw:
Relates: #4043


dkd-kaehm (approved):
Can be merged if CI succeeds.

Issue #4041, opened 2024-05-15
[TASK] Solr 9.6 compatibility


dkd-friedrich:
**What should be done in the scope of this task?**

Upcoming EXT:solr 13 for TYPO3 13 should support Apache Solr 9.6.

dkd-kaehm:
Done in #4056

PR #4056, opened 2024-05-16
[TASK] Apache Solr 9.6 compatibility

dkd-friedrich:
Allows the usage of EXT:solr with Apache Solr 9.6, Apache Solr 9.6 is now the recommended version and tests will run against this version.

Resolves: #4041


dkd-kaehm (approved):
:100:

Issue #4018, opened 2024-04-29
[FEATURE] Add contentObjectData to fluid view


mhirdes:
**Is your feature request related to a problem? Please describe.**
I would like to insert the search form on two different places in our template with f:cObject. To seperat these two layouts from each other, it would be awesome to have access to the content object. Then I can call the view via:
`<f:cObject typoscriptObjectPath="tt_content.list.20.solr_pi_search" data="{template: 'myTemplate'}"/>`

and in my template Form.html

`<f:if condition="{contentObjectData.template} == 'myTemplate'">`

**Describe the solution you'd like**
Add to the controller the following line:
`$view->assign('contentObjectData', $this->configurationManager->getContentObject()->data);`

**Target versions**
TYPO3 v12

christophlehmann:
The current content object should also be available as request attribute (since v11/v12 ?!)

PR #4059, opened 2024-05-16
[FEATURE] Add contentObjectData to searchController

spoonerWeb:
# What this pr does

Add record data of current content object in all actions of Search controller

Fixes: #4018


dkd-kaehm (approved):
:+1:

PR #4069, opened 2024-05-17
[TASK] Add access plugin tests

dkd-friedrich:
Add tests to explicitly test the filtering done by the Solr Access plugin.

Relates: https://github.com/TYPO3-Solr/solr-typo3-plugin/issues/33


dkd-kaehm (approved):
:100:

Issue #3750, opened 2023-08-08
[TASK] Use POST for Solr queries to be able to avoid enabled remote and body streaming


dkd-kaehm:
**What should be done in the scope of this task?**

EXT:solr uses remote and body streams for delete queries. This should be avoided in future.
See: 
* https://solr.apache.org/guide/solr/latest/upgrade-notes/major-changes-in-solr-9.html#security-2
* https://issues.apache.org/jira/browse/SOLR-14853

### Todos:

- [ ] Rewrite all queries using both features to POST
- [ ] Disable both features in Dockerfile 
- [ ] provide notes in the docs
- [ ] revert #3967 for 13+

Related: #3748 

### Target Version:

- [ ] 13 aka main, but only after branching release-12.0.x

dkd-friedrich:
Done in #4073

Upcoming rebase of main Branch will include this adaptions for EXT:solr 13

PR #4073, opened 2024-05-17
[TASK] Remove content stream usage

dkd-friedrich:
Drops the usage of content streams for Solr queries, this removes the dependency to SOLR_ENABLE_STREAM_BODY.

Resolves: #3750


dkd-kaehm (approved):
:100:

Issue #4080, opened 2024-05-28
[BUG] Retrieving plugin name in combination with gridelements leads to warnings within the backend page module


3l73:
**Describe the bug**

**To Reproduce**
1. Install extension `gridelementsteam/gridelements`
2. Install extension `apache-solr-for-typo3/solr`
3. Add a page and place the search plugin into it
4. After saving and go back to the page module the warning is displayed at the top of the page

**Expected behavior**
No warning should be shown at the top of the page

**Screenshots**
![Backend_Warning](https://github.com/TYPO3-Solr/ext-solr/assets/4559579/f2331169-8904-4a00-8022-8ab1b7680da5)

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.14
 - Browser: Chromium
 - EXT:solr Version: 12.0.3
 - Used Apache Solr Version: -
 - PHP Version: 8.3
 - MySQL Version: -

**Additional context**
Within class `\ApacheSolrForTypo3\Solr\Backend\SettingsPreviewOnPlugins` the backend utility to get the merged labels.

https://github.com/TYPO3-Solr/ext-solr/blob/9f5ed12031e868bbd00f0c0a8c9decd0b893fa99/Classes/Backend/SettingsPreviewOnPlugins.php#L152

The method `BackendUtility::getLabelFromItemListMerged()` offers a 5th argument `$row` which is at that point optional.

However the option is passed threw other methods and switch from optional to a kind of required argument.

The transition occurred in https://github.com/TYPO3/typo3/blob/85cb1b09b03366d4cf690064d9f2afb013b27c82/typo3/sysext/core/Classes/DataHandling/ItemProcessingService.php#L50

which is triggered by

https://github.com/TYPO3/typo3/blob/85cb1b09b03366d4cf690064d9f2afb013b27c82/typo3/sysext/backend/Classes/Utility/BackendUtility.php#L1360

A solution could be to call `BackendUtility::getLabelFromItemListMerged()` with the row information, that seams be available at that point.

```php
$label = BackendUtility::getLabelFromItemListMerged($this->pluginsTtContentRecord['pid'], 'tt_content', 'list_type', $this->pluginsTtContentRecord['list_type'], $this->pluginsTtContentRecord);
```

dkd-kaehm:
@3l73 
Hehe `Refs: #4080` seems to be not functional on GitHub.

The fix will be ported automatically to main/dev-13.0.x on next rebase.

PR #4081, opened 2024-05-28
[TASK] Call getLabelFromItemListMerged with the current row data

3l73:
# What this pr does

This commit add the row information as 5th argument in the function call of 'BackendUtility::getLabelFromItemListMerged()'.

Even the argument is optional for this function, it becomes a kind of required in the following nested function calls within TYPO3.

# How to test

1. Install extension `gridelementsteam/gridelements`
2. Install extension `apache-solr-for-typo3/solr`
3. Add a page and place the search plugin into it
4. After saving and go back to the page module
5. There should not be any warnings be visible related to a missing pid

Fixes: #4080


dkd-kaehm (approved):
:100:

Issue #4082, opened 2024-06-03
[BUG] numeric facet range slider sends lot of requests to server


hvomlehn-sds:
**Describe the bug**
When using a numeric range with a slider facet, the facet updates continously while sliding the slider with the mouse. Updating the UI to show the curretly slided to value is appreciated. However, the component sends the new facet-setting to the server during the slide as well.

**Expected behavior**
The jQuery-API of the slider component provides a stop-event in addition to the sliede event. The stop-event should be used to trigger the url update only, so that server requests are only sent, once the user releases the slider.

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.15
 - Browser: Firefox, Chromium, probably any browser
 - EXT:solr Version: 12.0.3

PR #4084, opened 2024-06-03
[BUGFIX] numeric facet range slider sends lot of requests to server

hvomlehn-sds:
Previously the numeric range slider updated the UI and the facet setting continously during sliding. This caused a lot of get requests to the server. This PR implents the desired beviour to send update requests only once the user releases the slider. The UI still gets updated continously by using the separate slide-event for that.

Fixes: https://github.com/TYPO3-Solr/ext-solr/issues/4082


dkd-kaehm (approved):
:100: 
Thanks for the fix

Issue #4087, opened 2024-06-06
[BUG] Wrong range string generation in DateRange facet


derhansen:
**Describe the bug**
The range string in the `DateRange` facet is generated, so that the time of the start- and enddate is 0000 (00:00). For the enddate, this is wrong and should be 2359 (23:59), so items for the whole enddate are selected.

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4
 - EXT:solr Version: 12.x
 - Used Apache Solr Version: 9.6

dkd-kaehm:
@derhansen 
Please check the Tests.

PR #4088, opened 2024-06-06
[BUGFIX] Fix range string calculation in DateRange facet

derhansen:
This change fixed the range string calculation, so items for the given date are included until 23:59 of the target enddate.

Fixes: #4087


derhansen:
Sorry, messed this up. Will create a new PR

PR #4090, opened 2024-06-07
[BUGFIX] Fix range string calculation in DateRange facet

derhansen:
This change fixes the range string calculation, so items for the given date are included until 23:59 of the target end date.

Fixes: #4087

---

### Maintainers Notes:

Ports wanted, before merging:

- [x] release-11.5.x
- [x] main (if not rebaseable to release-12.0.x)


dkd-kaehm (approved):
@derhansen 
Thanks for the fix.

PR #4106, opened 2024-07-11
[TASK] Apache Solr 9.6.1 compatibility

dkd-kaehm:
Allows the usage of EXT:solr with Apache Solr 9.6.1, 
Apache Solr 9.6.1 is now the recommended version and tests will run against this version.


dkd-friedrich (approved):
👍

PR #4109, opened 2024-07-16
[FEATURE] Add flag to site config to skip DataHandler hooks

### LINKED ISSUE #4108: [FEATURE] Please describe your feature wish here

Currently the 2 hooks of EXT:solr into DataHandler are fired **for all sites**, their pages and records. This is time consuming. Especially the RecordMonitor is taking up to 10s to do its thing.

For sites that don't use solr, it'll be great to have them skip the DataHandler hooks completely.

**Target versions**

EXT:solr 12.x

kitzberger:
# What this pr does

Currently the 2 hooks of EXT:solr into DataHandler are fired for all sites, their pages and records. This is time consuming. Especially the RecordMonitor is taking up to 10s to do its thing.

# How to test

Add the flag `solr_skip_hooks` to a site config file and edit/save one of it's pages. Both GarbageCollector and RecordMonitor should skip their `processDatamap_afterDatabaseOperations` method.

Fixes: #4108

---

### Maintainers Notes:

- **Don't merge into release-12.0.x**

### DIFF
--- a/Classes/Util.php
+++ b/Classes/Util.php
@@ -25,8 +25,11 @@ use Doctrine\DBAL\Exception as DBALException;
 use TYPO3\CMS\Backend\Utility\BackendUtility;
 use TYPO3\CMS\Core\Context\Context;
 use TYPO3\CMS\Core\Context\Exception\AspectNotFoundException;
+use TYPO3\CMS\Core\Exception\SiteNotFoundException;
+use TYPO3\CMS\Core\Site\SiteFinder;
 use TYPO3\CMS\Core\Utility\ExtensionManagementUtility;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Core\Utility\MathUtility;
 
 /**
  * Utility class for tx_solr
@@ -132,6 +135,34 @@ class Util
         }
 
         return $isWorkspaceRecord;
+    }
+
+    public static function skipHooksForRecord(
+        string $table,
+        int $uid,
+        ?int $pid
+    ): bool {
+        if (is_null($pid) && MathUtility::canBeInterpretedAsInteger($uid)) {
+            $recordInfo = BackendUtility::getRecord($table, $uid, 'pid');
+            if (!is_null($recordInfo)) {
+                $pid = $recordInfo['pid'] ?? null;
+            }
+        }
+
+        if (!is_null($pid)) {
+            /** @var SiteFinder $siteFinder */
+            $siteFinder = GeneralUtility::makeInstance(SiteFinder::class);
+            try {
+                $site = $siteFinder->getSiteByPageId($pid);
+            } catch (SiteNotFoundException) {
+                return false;
+            }
+            if ((bool)($site->getConfiguration()['solr_skip_hooks'] ?? false)) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     /**
--- a/Classes/GarbageCollector.php
+++ b/Classes/GarbageCollector.php
@@ -189,6 +189,10 @@ class GarbageCollector implements SingletonInterface
             return;
         }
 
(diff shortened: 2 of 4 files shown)


dkd-kaehm:
Thanks for that feature.
Why not using the "Enable Solr for this site" checkbox from Site-Config?
What happens with mounted pages between different sites?

grafik

I'll change the base and target branch to release-12.0.x to be able to run the tests. Currently the main branch not stable enough for new features. We'll switch back to main, if that feature should be merged. 
Please use `git reset --hard origin/skip-hooks` to work on that feature.

kitzberger:
@dkd-kaehm, to be honest I don't know the implications behind that `solr_enable_read` flag.

Thing is, we are using EXT:solr for the site in question but not its indexing. That's why we needed a separate flag. Does this make sense? I not 100% sure ;-)

I've noticed that RecordMonitor checks on `no_search_sub_entries` being present in the rootline of the modified record. That's a neat feature but a bit magic too. I wouldn't assume (as an integrator) that this flag had any effect on the RecordMonitor. I'd rather disable the hook in the site config.

dkd-kaehm:
> @dkd-kaehm, to be honest I don't know the implications behind that `solr_enable_read` flag.
> 
> Thing is, we are using EXT:solr for the site in question but not its indexing. That's why we needed a separate flag. Does this make sense? I not 100% sure ;-)
> 
> I've noticed that RecordMonitor checks on `no_search_sub_entries` being present in the rootline of the modified record. That's a neat feature but a bit magic too. I wouldn't assume (as an integrator) that this flag had any effect on the RecordMonitor. I'd rather disable the hook in the site config.

OK, you see, there is another way with `no_search_sub_entries`, so why introducing the new setting?
Would you like to reuse the existing possibilities/settings and maybe add a hint in the docs, for example in FAQ?
Beside of that, there is a feature for your purpose: https://docs.typo3.org/p/apache-solr-for-typo3/solr/12.0/en-us/Configuration/Reference/ExtensionSettings.html#monitoringtype

I would propose to fix the existing things if they not working well, instead of introducing new settings.

kitzberger:
> Beside of that, there is a feature for your purpose: https://docs.typo3.org/p/apache-solr-for-typo3/solr/12.0/en-us/Configuration/Reference/ExtensionSettings.html#monitoringtype

Thanks for pointing that out! (didn't know about that). But unfortunately that's a global option, right? I need a per site option. Hence my PR.

dkd-kaehm:
> > Beside of that, there is a feature for your purpose: https://docs.typo3.org/p/apache-solr-for-typo3/solr/12.0/en-us/Configuration/Reference/ExtensionSettings.html#monitoringtype
> 
> Thanks for pointing that out! (didn't know about that). But unfortunately that's a global option, right? I need a per site option. Hence my PR.

This could be a option for that feature.

dkd-friedrich:
My first thought was also that it might be useful to be able to set the `monitoringType` per site. We are planning to add site information anyway to improve the maintenance of the event queue.

But if Solr connections are deactivated for a site, it makes sense to cancel the monitoring, but I'm not sure what effect this will have on mounted sites. I think there are some possible side effects to consider here.

kitzberger:
I've come across another problem now: trying to save `be_users` is running into memory exhaustion. Since those records are "global" (pid=0) the `pages.no_search_sub_entries` setting cannot be used to deactivate those hooks.

Classes/Util.php
```diff
@@ -134,6 +137,34 @@ public static function isDraftRecord(
         return $isWorkspaceRecord;
     }
 
+    public static function skipHooksForRecord(
+        string $table,
+        int $uid,
+        ?int $pid
+    ): bool {
```
kitzberger, line 144: @dkd-kaehm, maybe have some kind of exclude list defined here? `sys_template`, `be_groups` , `be_users`

PR #4110, opened 2024-07-17
[FIX] deprecations in Dockerfile

dkd-kaehm:
Fixes deprecation notices in Dockerfile:

 - MaintainerDeprecated: Maintainer instruction is deprecated in favor of using label (line 2)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 3)

### DIFF
--- a/Docker/SolrServer/Dockerfile
+++ b/Docker/SolrServer/Dockerfile
@@ -1,6 +1,6 @@
 FROM solr:9.6.1
-MAINTAINER dkd Internet Service GmbH <info@dkd.de>
-ENV TERM linux
+LABEL maintainer="dkd Internet Service GmbH info@dkd.de"
+ENV TERM=linux
 
 ARG SOLR_UNIX_UID="8983"
 ARG SOLR_UNIX_GID="8983"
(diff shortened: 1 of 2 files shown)


Docker/SolrServer/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM solr:9.6.1
-MAINTAINER dkd Internet Service GmbH <info@dkd.de>
-ENV TERM linux
+LABEL maintainer="dkd Internet Service GmbH info@dkd.de"
```
dkd-friedrich, line 2: The recommended replacement for `MAINTAINER` differs from this solution, see
https://docs.docker.com/reference/dockerfile/#maintainer-deprecated

IMHO we should adjust this and use `LABEL org.opencontainers.image.authors`

dkd-friedrich (changes requested):

Docker/SolrServer/Dockerfile
dkd-kaehm, line 2:
```suggestion
LABEL org.opencontainers.image.authors="dkd Internet Service GmbH info@dkd.de"
```

dkd-friedrich (approved):
LGTM

Issue #4107, opened 2024-07-15
[BUG] Index document not deleted if not in index queue


dkd-friedrich:
**Describe the bug**

If an index document has no representation in the index queue, EXT:solr fails to remove the item on deletions or updates of the corresponding record in TYPO3.

**Expected behavior**
Index documents should be deleted on deletions or updates even if there is no index queue entry.

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.5
 - EXT:solr Version: 12.0.3
 - Used Apache Solr Version: 9.6.1
 - PHP Version: 8.3

dkd-friedrich:
Fixed in #4113

Issue #3547, opened 2023-03-16
[BUG] Unable to add pages with future starttime to index queue


DanielSiepmann:
**Describe the bug**
Pages with an `starttime` set to the future are not indexed, until they are modified after reaching starttime, or reindexed via scheduler after reaching starttime.

**To Reproduce**
1. Create a new page
2. Set `starttime` to the future
3. Save page
4. Inspect the table `tx_solr_indexqueue_item`, you won't find an entry for the page,.

**Expected behavior**
I would expect to create or edit an TYPO3 page record with a `starttime` in the future. This page should then be added as index queue item to the database table to be indexed as soon as the page is available.

**Screenshots**
No Screenshots available

**Used versions (please complete the following information):**
 - TYPO3 Version: e.g. 11.5.24
 - Browser: Firefox, but should not be related.
 - EXT:solr Version: 11.5.1
 - Used Apache Solr Version: 8.11.1, but should not be related
 - PHP Version: 8.1.16, but should not be related
 - MySQL Version: 8.0.32-0ubuntu0.20.04.2. but should not be related

**Additional context**
We already wrote about that on slack: https://typo3.slack.com/archives/C02FF05Q4/p1678968753747799

Related Code:
A TSFE will be initialized prior adding the entry to the DB
That initialization will fail. The TSFE can not find the page as it is not available yet at https://github.com/TYPO3/typo3/blob/v11.5.24/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L1084. This will result in an Exception at https://github.com/TYPO3/typo3/blob/v11.5.24/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L998. This then leads to missing TSFE: https://github.com/TYPO3-Solr/ext-solr/blob/11.5.1/Classes/FrontendEnvironment/Tsfe.php#L186 which then will prevent further processing and adding the db entry.

DanielSiepmann:
Our current hotfix for our project:

```patch
diff --git a/Classes/FrontendEnvironment.php b/Classes/FrontendEnvironment.php
index 80ae2c67..88077e0a 100644
--- a/Classes/FrontendEnvironment.php
+++ b/Classes/FrontendEnvironment.php
@@ -74,11 +74,11 @@ class FrontendEnvironment implements SingletonInterface
             $rootPageRecordUid = $pageRecord['l10n_parent'];
         }
 
[...]

PR #4113, opened 2024-07-24
[BUGFIX] Ensure index document is deleted

dkd-friedrich:
# What this pr does

If an index document has no representation in the index queue, EXT:solr fails to remove the item on deletions or updates of the corresponding record in TYPO3.

By adjusting the checks in the record monitor, garbage remover and update handler this issue is resolved. The
ConfigurationAwareRecordService is no longer using BackendUtility::getRecord() as this would return hidden records.

Resolves: #4107


dkd-kaehm (approved):

dogawaf:
These PR reintroduce this issue: #3547

Issue #4095, opened 2024-06-24
[BUG] "Clear Index Queue" TypeError Site is null


sal-lochbaum:
**Describe the bug**
When I click the "Clear Index Queue" button in the Index Queue backend module I get an error response from Typo3:

(1/1) TypeError
ApacheSolrForTypo3\Solr\IndexQueue\Queue::deleteItemsBySite(): Argument #1 ($site) must be of type ApacheSolrForTypo3\Solr\Domain\Site\Site, null given, called in /var/www/html/vendor/apache-solr-for-typo3/solr/Classes/Controller/Backend/Search/AbstractModuleController.php on line 231

**To Reproduce**
1. Have multiple sites
1. Go to Index Queue backend module
2. Click on 'Clear Index Queue'
3. Confirm the popup

**Expected behavior**
A clean index queue

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.6
 - EXT:solr Version: latest dev-release-12.0.x
This probably also applies to master/13 as the relevant code is the same.

**Additional context**
There is a fallback in AbstractModuleController->initializeAction called "autoSelectFirstSiteAndRootPageWhenOnlyOneSiteIsAvailable" which probably "hides" this bug if you only have one site.

As far as I could debug it:

1. AbstractModuleController->clearIndexQueueAction() tries using $this->selectedSite which is null.
2. The AbstractModuleController->initializeAction is responsible for setting the site but the id set in the frontend is not received in the backend. i.e. this condition is always false
```
if ($this->request->hasArgument('id')) {
    $this->selectedPageUID = (int)$this->request->getArgument('id');
}
```
3. The id is set in the frontend, but the variable pageUID is null in IndexQueueModule/Index.html
` <f:form action="clearIndexQueue" additionalParams="{id: '{pageUID}'}">`
4. That fluid variable is beeing correctly set in AbstractModuleController->initializeView() but does not reach the Index.html

I narrowed it down to the initializeAction beeing called twice:

1. initializeAction is called automatically by Typo3 -> creates new View
2. initializeView is called automatically by Typo3
3. initializeAction is called manually in IndexQueueModuleController->indexAction -> creates new View w/o variables from step 2

IMO the manual call should be deleted
https://github.com/TYPO3-Solr/ext-solr/blob/4c13753ce11810e9bbf71745b3a0ca5872191d78/Classes/Controller/Backend/Search/IndexQueueModuleController.php#L68
```
[...]

PR #4120, opened 2024-07-29
[BUGFIX] Fix index queue clearance

dkd-friedrich:
Fixes index queue clearance in multi site
environment.

Fixes: https://github.com/TYPO3-Solr/ext-solr/issues/4095


dkd-kaehm (approved):
LGTM

PR #4145, opened 2024-09-10
Use PHP generator to prevent processing of all available site

sfroemkenjw:
# What this pr does

Use PHP generator with yield to process just the available sites needed

# How to test

With over 300 root pages it needs over 30 seconds to build up all available sites. With help of a PHP generator we can stop processing all available sites if just the first site is requested. It's also helpful to just check, if there are available sites.

This should speed up performance a lot.

PR #4154, opened 2024-09-16
Reduce call get available sites for TYPO3 12.4

sfroemkenjw:
What this pr does?

Use PHP generator with yield to process just the available sites needed
How to test

With over 300 root pages it needs over 30 seconds to build up all available sites. With help of a PHP generator we can stop processing all available sites if just the first site is requested. It's also helpful to just check, if there are available sites.

This should speed up performance a lot.

### DIFF
--- a/Tests/Unit/Domain/Site/SiteRepositoryTest.php
+++ b/Tests/Unit/Domain/Site/SiteRepositoryTest.php
@@ -89,7 +89,6 @@ class SiteRepositoryTest extends SetUpUnitTestCase
         $this->assertThatSitesAreCreatedWithPageIds([333], [
             0 => ['language' => 0],
         ]);
-        $this->assertCacheIsWritten();
 
         $site = $this->siteRepository->getFirstAvailableSite();
         self::assertInstanceOf(Site::class, $site);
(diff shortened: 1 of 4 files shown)


sfroemkenjw:
Related to PR: #4145

sfroemkenjw:
Hi @dkd-kaehm 

I don't understand. After your approval you will get 3 options to merge my PR: Merge, Rebase and Squash. In TYPO3 docs team we always work this way. I have no clue how to squash commits in my side.

Stefan

dkd-kaehm:
OK I'll do the squash.
I have a question on one change.

Tests/Unit/Domain/Site/SiteRepositoryTest.php
```diff
@@ -89,7 +89,6 @@ public function canGetFirstAvailableSite(): void
         $this->assertThatSitesAreCreatedWithPageIds([333], [
             0 => ['language' => 0],
         ]);
-        $this->assertCacheIsWritten();
```
dkd-kaehm, line 92: Why removing that assertion?

sfroemkenjw: Caching of available site only happens while calling `getAvailableSites`. But as `hasAvailableSites` works with the generator directly "typo3ManagedSolrSites", no caching happens. That's why I have removed that check.

dkd-kaehm:
Squashed and rebased to release-12.0.x on 2024.12.12 state.

dkd-kaehm:
Replaced by #4260

PR #4164, opened 2024-09-23
[TASK] Adapt simulated environment for TYPO3 13

dkd-friedrich:
To allow indexing various environments have to be simulated, in TYPO3 13
the ContentObjectRenderer of TSFE has to initialized with the simulated
request and the request attribute 'frontend.typoscript' hat to be set.


dkd-kaehm (approved):
:+1:

PR #4192, opened 2024-10-14
Fix formatting and update Slack link on FAQ

trolologuy:
# What this pr does

The list was not diplayed properly and the slack link wasn't redirecting. Tested the changes with: https://www.tutorialspoint.com/online_restructure_editor.php

# How to test

If you open this page:
https://docs.typo3.org/p/apache-solr-for-typo3/solr/11.0/en-us/FAQ/Index.html

You'll notice the broken list as first item. Also the slack link doesn't work.


dkd-kaehm:
release-11.0.x is not supported anymore, but I'll merge that PR

Issue #4160, opened 2024-09-19
[TASK:T13] fix mount-pages issues


dkd-kaehm:
- [ ] `ConnectionManagerTest::canFindSolrConnectionForMountedPageIfMountPointIsGiven`
- [ ] `PageIndexerTest::canIndexMountedPage`
- [ ] `PageIndexerTest::canIndexMultipleMountedPage`

```
1) ApacheSolrForTypo3\Solr\Tests\Integration\ConnectionManagerTest::canFindSolrConnectionForMountedPageIfMountPointIsGiven
ApacheSolrForTypo3\Solr\NoSolrConnectionFoundException: Could not find a Solr connection for page [24] and language [0].

/home/runner/work/ext-solr/ext-solr/.Build/vendor/typo3/cms-core/Classes/Utility/GeneralUtility.php:2808
/home/runner/work/ext-solr/ext-solr/Classes/ConnectionManager.php:231
/home/runner/work/ext-solr/ext-solr/Classes/ConnectionManager.php:206
/home/runner/work/ext-solr/ext-solr/Classes/ConnectionManager.php:118
/home/runner/work/ext-solr/ext-solr/Tests/Integration/ConnectionManagerTest.php:219

--

2) ApacheSolrForTypo3\Solr\Tests\Integration\IndexQueue\FrontendHelper\PageIndexerTest::canIndexMountedPage
Could not find content from mounted page in solr
Failed asserting that '{\n
  "responseHeader":{\n
    "status":0,\n
    "QTime":0,\n
    "params":{\n
      "q":"*:*"}},\n
  "response":{"numFound":0,"start":0,"numFoundExact":true,"docs":[]\n
  }}\n
' [ASCII](length: 161) contains ""title":"FirstShared (Not root)"" [ASCII](length: 32).

/home/runner/work/ext-solr/ext-solr/Tests/Integration/IndexQueue/FrontendHelper/PageIndexerTest.php:316

3) ApacheSolrForTypo3\Solr\Tests\Integration\IndexQueue\FrontendHelper\PageIndexerTest::canIndexMultipleMountedPage
Unexpected amount of documents in the core
Failed asserting that '{\n
  "responseHeader":{\n
    "status":0,\n
    "QTime":0,\n
    "params":{\n
      "q":"*:*"}},\n
  "response":{"numFound":0,"start":0,"numFoundExact":true,"docs":[]\n
  }}\n
[...]

PR #4195, opened 2024-10-16
[TASK] Adjust mount point indexing

dkd-friedrich:
Mount point indexing and corresponding tests have been adjusted for
TYPO3 13. Mount points are supported in general and the mounted pages
will be indexed like standard pages.

But there are some points to consider:

*   Cross-site mounts require that `config.index_enable = 1` is set
    in the pagetree of the mounted page, as TYPO3 loads the
    TypoScript of the mounted page.
    This is probably a bug in the TYPO3 core and may change, due to
    changes in RootlineUtility->generateRootlineCache() the mount page
    pid isn't used any more.
*   Pages in a pagetree without a site configuration cannot be
    indexed, in fact TYPO3 currently can't mount a page from a page
    tree without a site configuration and an exeception occurs.

Test procedure in PageIndexerTest has been improved and now uses
the PageUriBuilder to build the url to index, instead of building
the url in the test itself.

Resolves: https://github.com/TYPO3-Solr/ext-solr/issues/4160

---

*Caution*: Do not squash!

### DIFF
--- a/Documentation/Releases/solr-release-13-0.rst
+++ b/Documentation/Releases/solr-release-13-0.rst
@@ -5,15 +5,30 @@
 Releases 13.0
 =============
 
+
+Release 13.0.0-beta-1
+=====================
+
+This is a first beta release for TYPO3 13.4 LTS
+
+New in this release
+-------------------
+
+Adjust mount point indexing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Mount point indexing and corresponding tests have been adjusted for TYPO3 13. Mount points are supported in general and the mounted pages will be indexed like standard pages.
+
+But there is a point to consider: Mounted pages from a pagetree without a site configuration cannot be indexed, in fact TYPO3 currently can't mount a page from a page tree without a site configuration and an exeception occurs.
+The behavior is intentionally designed this way in TYPO3 core, the background is that it is not possible to specify the languages of the mounted page tree without Site Configuration.
+
+.. note::
+   We require at least TYPO3 13.4.2, as this version contains some bugfixes that address problems with the determination of TypoScript and the site configuration of mounted pages.
+
 Release 13.0.0-alpha-1
 ======================
 
 This is a first alpha release for upcoming TYPO3 13 LTS
-
-Known Bugs
-----------
-
-*   Mount pages cannot be indexed, see `#4160 <https://github.com/TYPO3-Solr/ext-solr/issues/4160>`__
 
 New in this release
 -------------------
--- a/Tests/Integration/IndexQueue/FrontendHelper/PageIndexerTest.php
+++ b/Tests/Integration/IndexQueue/FrontendHelper/PageIndexerTest.php
@@ -337,7 +337,7 @@ class PageIndexerTest extends IntegrationTestBase
 
         $this->cleanUpAllCoresOnSolrServerAndAssertEmpty();
         $this->importCSVDataSet(__DIR__ . '/Fixtures/can_index_mounted_page_from_another_site.csv');
-        $this->addTypoScriptToTemplateRecord(111, 'config.index_enable = 1');
+        $this->addTypoScriptToTemplateRecord(1, 'config.index_enable = 1');
         $this->indexQueuedPage();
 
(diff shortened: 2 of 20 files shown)


dkd-friedrich:
Test `ConnectionManagerTest::canFindSolrConnectionForMountedPageIfMountPointIsGiven` is still disabled and fails.

Reason is probably a bug in `RootlineUtility->generateRootlineCache()`, as the pid of the mounted page is used
for the root line instead of the mount page pid like in former issues. We should investigate and discus this
behavior.

Documentation/Releases/solr-release-13-0.rst
```diff
@@ -5,19 +5,44 @@
...
+Adjust mount point indexing
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Mount point indexing and corresponding tests have been adjusted for TYPO3 13. Mount points are supported in general and the mounted pages will be indexed like standard pages.
+
+But there are some points to consider:
+
+*   Cross-site mounts require that `config.index_enable = 1` is set in the pagetree of the mounted page, as TYPO3 loads the TypoScript of the mounted page
+*   Mounted pages from a pagetree without a site configuration cannot be indexed, in fact TYPO3 currently can't mount a page from a page tree without a site configuration and an exeception occurs
+
```
dkd-kaehm, line 45: Double pasted the copy?

Tests/Integration/IndexQueue/FrontendHelper/PageIndexerTest.php
```diff
@@ -285,67 +287,133 @@ public function canExecuteAdditionalPageIndexer(): void
...
+     *  |   |
+     *  |   |—[24] FirstShared (Not root)
      *  |
-     *  ——[14] Mount Point (to [24] to show contents from)
+     *  |—[ 1] Page (Root Testpage 1)
+     *      |
+     *      |—[14] Mount Point (to [24] to show contents from)
+     */
+    #[Test]
+    public function canIndexCrossSiteMountsWithoutSiteConfiguration(): void
```
dkd-kaehm, line 373: Is that scenario possible in TYPO3 Backend at all? 
As far as I know, TYPO3 creates a site-config.yaml file as soon the page is marked as root.
Can that test-case be removed?

dkd-kaehm (changes requested):
Is there a issue in forge for trouble in `RootlineUtility->generateRootlineCache()`?

dkd-friedrich:
> Is there a issue in forge for trouble in `RootlineUtility->generateRootlineCache()`?

Yes, I've created an issue and prepared a patch, see https://forge.typo3.org/issues/105376

Documentation/Releases/solr-release-13-0.rst
dkd-friedrich: Right, will remove the duplicate

Tests/Integration/IndexQueue/FrontendHelper/PageIndexerTest.php
dkd-friedrich: This is the original test using a sysfolder with pages that should be mounted (no site root), but currently TYPO3 cannot mount such pages and an exception occurs:

```
 (1/1) #1521716622 TYPO3\CMS\Core\Exception\SiteNotFoundException 
No site found in root line of page 86
```

This issue already exists in TYPO3 12, I am currently not sure why this test worked in TYPO3 12, test results in v12 might be misleading.

As it's not a valid and supported scenario in the TYPO3 core, we could IMHO delete the test.

dkd-friedrich:
I am postponing work on this problem until we receive feedback in https://forge.typo3.org/issues/105376.

Our further action depends on whether the patch is adopted.

Tests/Integration/IndexQueue/FrontendHelper/PageIndexerTest.php
dkd-friedrich: I kept this test for now, till we could check if this is a bug in TYPO3 or if mounts outside of a side aren't intended to work

dkd-kaehm (approved):

dkd-kaehm (approved):
:100:

Issue #4197, opened 2024-10-23
[FEATURE] Support multiple values in suggestField setting


saitho:
**Is your feature request related to a problem? Please describe.**
I want to use the suggest feature but need multiple fields to be considered.
The use case is searching for a product title or product id.

**Describe the solution you'd like**
The setting `plugin.tx_solr.search.suggest.suggestField` can be used to define which field is used for suggests.
This field currently only works with a single value. I'd like to be able to set multiple comma-separated values, just like queryFields.

Example:
```
plugin.tx_solr {
  suggest {
    showTopResults = 1
    additionalTopResultsFields = productNumber_stringS
    suggestField = spell,productNumber_stringS
  }
  search.query {
    queryFields = title^5.0, productNumber_stringS^10.0
    filter.productOnly = type:tx_myextension_products
  }
}
```

**Target versions**
12.0.5

PR #4198, opened 2024-10-23
[FEATURE] Support multiple values in suggestField

saitho:
It is now possible to set multiple fields in suggestField, similar to queryFields, using commas.

# What this pr does

Please add a description here

# How to test

Fixes: #4197


saitho:
Change not needed on my side anymore, as I'm now using the regular search instead of suggest endpoint.

PR #4200, opened 2024-10-24
[TASK] Do not ignore database exceptions

### LINKED ISSUE #4199: [TASK] Do not ignore database exceptions
**What should be done in the scope of this task?**

Remove a try/catch block around a database select query. Faulty database queries should mostly never get muted.

christophlehmann:
# What this pr does

This removes a try catch block around a select query. 

When a database query is not okay for any reason, then i like to get the reason directly.

Fixes: #4199


masi:
I agree that muted exceptions are a no-go, but IMHO a proper handling is much better than letting the application break with an unhandled exception. And there are already enough of completely unhelpful stack traces in logs created by Solr.

I suggest at least logging the SQL statement and the context in some readable manner before letting the calling code choke on the exception.

PR #4201, opened 2024-10-25
[TASK] Remove JSONP callback in suggest

bmack:
# What this pr does

This change removes the callback logic
in the Suggest AJAX Call via JSONP, as JSONP
is known to be used to call untrusted third-party
code, and can thus be removed, as custom suggest
code is done anyway via custom JS implementations.

See https://en.wikipedia.org/wiki/JSONP#Security_concerns

# How to test

See if suggest still works as expected.


dkd-kaehm (approved):
:100:

Issue #3837, opened 2023-10-17
[BUG] Operation not permitted (ix-file-permissions.sh)


varioous:
When starting solr service the start fails everytime with: 

`ddev logs -s solr
/opt/solr/docker/scripts/solr-foreground: running /docker-entrypoint-initdb.d/as-sudo/fix-file-permissions.sh
chown: changing ownership of '/var/solr/data/cores': Operation not permitted`

.ddev-docker-compose.solr.yaml
```yaml
services:
  solr:
    container_name: ddev-${DDEV_SITENAME}-solr
    image: typo3solr/ext-solr:12.0.x-dev
    ports:
      - 8983
    labels:
      com.ddev.site-name: ${DDEV_SITENAME}
      com.ddev.approot: $DDEV_APPROOT
    environment:
      - VIRTUAL_HOST=$DDEV_HOSTNAME
      - HTTP_EXPOSE=8983
    volumes:
      - ./solr/cores:/var/solr/data/cores
      - solr:/var/solr
  web:
    links:
      - solr

volumes:
  solr:
```

**Expected behavior**
Solr started successfully

**Used versions:**
 - TYPO3 Version:12.4
 - EXT:solr Version: 12.0.0

**Additional context**
Setup: Windows + Docker-Desktop + WSL2

dkd-kaehm:
@varioous
The path `.ddev/solr/cores` is mounted to the exported docker volume without synchronize the unix user ids.
In Apache Solr and EXT:solr docker images, the service runs with unix user id 8983, your host user has other UID.
The circular bind mounts  `.ddev/solr/cores` on sub path of `exported docker volume` is not good idea.

If you want to enable/disable cores, use the provided feature ENV-"TYPO3_SOLR_ENABLED_CORES":
https://docs.typo3.org/p/apache-solr-for-typo3/solr/12.0/en-us/Appendix/DockerTweaks.html
If not, then just remove the `- ./solr/cores:/var/solr/data/cores` line.

If you want for example deploying the stuff from your git repo to Apache Solr Service/Stages,
or test some changes on Schemas, etc,
then inspire you from [our solr-ddev-site](https://github.com/TYPO3-Solr/solr-ddev-site/blob/8a59bcb9b24e964415e5c2375fa7917bd17941f1/.ddev/docker-compose.solr.yaml).

We'll not fix the "specials" like this, to avoid even more complexity, but we want to document the Docker Build tweak of EXT:solr image on #3838.

[...]

PR #4219, opened 2024-11-12
!!![FIX] Docker execution order issue for as-sudo tweaks

itzonban:
This PR renames the file `as-sudo-tweaks.sh` to `0_as-sudo-tweaks.sh` 
and moves the folder 
`/docker-entrypoint-initdb.d/as-sudo/`
to
`/docker-entrypoint-initdb.d-as-sudo/` to fix the execution order issue 
when setting the correct file permissions when starting the docker container, 
leading to a `Operation not permitted` errors. 

More details see: 
* https://github.com/TYPO3-Solr/ext-solr/issues/3837#issuecomment-2461668377.
* https://github.com/TYPO3-Solr/ext-solr/pull/4219#issuecomment-2622600937

### Impact:

This change requires adjustments in your Docker setup, only if you modified:
files `/docker-entrypoint-initdb.d/as-sudo/` in folder 
or `/docker-entrypoint-initdb.d/as-sudo-tweaks.sh`.

Make sure to use:
* `/docker-entrypoint-initdb.d/0_as-sudo-tweaks.sh` instead of `/docker-entrypoint-initdb.d/as-sudo-tweaks.sh`
* `/docker-entrypoint-initdb.d-as-sudo` instead of `/docker-entrypoint-initdb.d/as-sudo/`

Fixes: #3837

---

### What this pr does

This PR renames the file `as-sudo-tweaks.sh` to `0_as-sudo-tweaks.sh` to fix the execution order issue when setting the correct file permissions when starting the docker container, leading to a `Operation not permitted` error. More details see https://github.com/TYPO3-Solr/ext-solr/issues/3837#issuecomment-2461668377.

Fixes: #3837

I tested the solution with v12.0.5, the issue probably exists since v11.6.0.


dkd-kaehm:
@itzonban 
Nice catch, thanks for fixing that issue.
I'll port the fix, and merge it asap.

dkd-kaehm:
This change is not enough, because the scripts will be executed anyway and in wrong manner.
**The scripts within `as-sudo/` folder MUST NOT be executed without root privileges.**

We should move the `as-sudo` (to root `(docker-entrypoint-as-sudo`) away from `/docker-entrypoint-initdb.d/` and adjust the Dockerfile and `as-sudo-tweaks.sh` file.

---

Output of `find /docker-entrypoint-initdb.d/ -mindepth 1 -type f | sort -n`

### Before renaming: 

The `fix-file-permissions.sh` executed twice:
1. without sudo
2. with sudo via `as-sudo-tweaks.sh`

```
docker-entrypoint-initdb.d/as-sudo/fix-file-permissions.sh
docker-entrypoint-initdb.d/as-sudo-tweaks.sh
docker-entrypoint-initdb.d/disable-cores.sh
```

### After renaming:

The `fix-file-permissions.sh` executed twice:
1. with sudo via `as-sudo-tweaks.sh`
2. without sudo

```
docker-entrypoint-initdb.d/0_as-sudo-tweaks.sh
docker-entrypoint-initdb.d/as-sudo/fix-file-permissions.sh
docker-entrypoint-initdb.d/disable-cores.sh
```

itzonban:
Thanks for the fix and the new version 👍

PR #4245, opened 2024-12-05
[BUGFIX] Prevent error when overlay is not available

magicsunday:
This fixes:

Core: Error handler (BE): PHP Warning: Trying to access array offset on value of type null in /var/www/vendor/apache-solr-for-typo3/solr/Classes/ContentObject/Relation.php line 311

# What this pr does

Please add a description here

# How to test


magicsunday:
Looks like I pushed the wrong branch for the 12 release. I've corrected it.

Issue #4249, opened 2024-12-09
[TASK] Remove Implicitly nullable parameter declarations deprecated (12.0.x)


thomashohn:
Fix Implicitly marking parameter as nullable is deprecated as a preparation for PHP 8.4

Issue #4247, opened 2024-12-09
[TASK] Fix Implicitly marking parameter as nullable is deprecated (main)


thomashohn:
Fix Implicitly marking parameter as nullable is deprecated as a preparation for PHP 8.4

PR #4250, opened 2024-12-09
[TASK] Remove Implicitly nullable parameter declarations deprecated

thomashohn:
# What this pr does

Remove Implicitly nullable parameter declarations deprecated in PHP 8.4

# How to test

Normal tests should be green

Fixes: #4249, #4247
Ports: #4248


dkd-kaehm (approved):
:100:

Issue #4255, opened 2024-12-12
[BUG] Relation with foreignLabel not working for mm tables


tillhoerner:
**Describe the bug**
The new `foreignLabel` property in `SOLR_RELATION` does not seem to work when used with mm tables.

**Expected behavior**
The `foreignLabel` property should correctly retrieve the label values from the related records when using mm tables.

**Used versions:**
 - TYPO3 Version: 12.4.23
 - EXT:solr Version: 12.0.5

### Target versions:

- [x] main aka release-13.0.x
- [x] release-12.0.x

PR #4256, opened 2024-12-12
[BUGFIX] Respect foreignLabel in related items from mm table

tillhoerner:
Fixes: #4255


dkd-kaehm (approved):
@tillhoerner 
Thanks for fixing that issue.
Would you like to port that into rlease-12.0.x?

Issue #4204, opened 2024-10-30
[TASK] Migrate Documentation to the new rendering


linawolf:
The old container will go out of Buisiness in November

---

Target Versions:

- [x] main
- [ ] release-12.0.x

dkd-kaehm:
Not yet fully done.

dkd-kaehm:
Version-Matrix table is not final, therefore the port into 12.0.x must be done after table get final state.

PR #4257, opened 2024-12-12
[DOCS] workaround for version matrix

dkd-kaehm:
Currently it is not possible to use spans in tables, like it was possible on Sphix-Variant. 
See: https://github.com/phpDocumentor/guides/issues/1187

On the test `test_documentation` CI job, we want to generate the docs twice to be able to recognize the errors and warnings in the docs. 
See: https://github.com/phpDocumentor/guides/issues/1188

Fixes: #4204


dkd-kaehm:
Will merge to prevent CI fails.

PR #4272, opened 2024-12-18
[BUGFIX] site has no attribte language

achimfritz:
get SiteLanguage from request instead of site


dkd-kaehm (approved):
Thanks for fixing that issue.

PR #4273, opened 2024-12-18
[BUGFIX] set PageInformation contentFromPid

achimfritz:
initialize TSFE should set contentFromPid in PageInformation because ContentObjectRenderer calls getContentFromPid method


dkd-kaehm (approved):
Thanks for the fix.

PR #4283, opened 2025-01-16
[BUGFIX] Add check if generator is valid before traversing it

jacobsenj:
If there's no solr configuration for any site, saving pages or content results in a PHP exception:

Uncaught TYPO3 Exception: Cannot traverse an already closed generator | Exception thrown in file .../Classes/Domain/Site/SiteRepository.php in line 135

The added check prevents this error.

Fixes: #4284

### DIFF
--- a/Classes/Domain/Site/SiteRepository.php
+++ b/Classes/Domain/Site/SiteRepository.php
@@ -106,7 +106,9 @@ class SiteRepository
     {
         $siteGenerator = $this->getAvailableTYPO3ManagedSites($stopOnInvalidSite);
         $siteGenerator->rewind();
-
+        if (!$siteGenerator->valid()) {
+            return null;
+        }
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
:100:

sfroemkenjw:
All methods in SiteRepository using the result from

```php
$siteGenerator = $this->getAvailableTYPO3ManagedSites($stopOnInvalidSite);
```

have to test against `->valid()`

If this is done, the patch is OK for me.

Stefan

jacobsenj:
I've added the checks for the remaining three calls.

sfroemkenjw (approved):
Look good to me. Thank you for providing that patch.

dkd-kaehm:
@andreaswolf 
Must be ported to release-12.0.x before merging.

PR #4338, opened 2025-03-27
[BUGFIX] Close #4337: fix exception in MountPagesUpdater when mount point has indexing disabled

### LINKED ISSUE #4337: [BUG] MountPagesUpdater throws an exception if indexing is disabled
**Describe the bug**

If indexing is disabled for the mount page, the attempt to fetch solr site in `ApacheSolrForTypo3\Solr\Domain\Index\Queue\RecordMonitor\Helper::addPageToMountingSiteIndexQueue` will return null and there will be an exception. A check is missing there.

Code is at https://github.com/TYPO3-Solr/ext-solr/blob/main/Classes/Domain/Index/Queue/RecordMonitor/Helper/MountPagesUpdater.php#L87

```
    protected function addPageToMountingSiteIndexQueue(int $mountedPageId, array $mountProperties): void
    {
        $siteRepository = GeneralUtility::makeInstance(SiteRepository::class);
        $mountingSite = $siteRepository->getSiteByPageId($mountProperties['mountPageDestination']);

        /** @var Page $pageInitializer */
        $pageInitializer = GeneralUtility::makeInstance(Page::class);
        $pageInitializer->setSite($mountingSite);
```

Here `$mountingSite` will be `null and it is correct if indexing is disabled for the mount point, but `setSite` expects a valis site.

**To Reproduce**
1. Create a new root page, which is a mount point to the existing root page.
2. Disable indexing via `solr_enabled_read: false` in the site config for that new site
3. Edit any page in the site -> exception

 - TYPO3 Version: 12.4.28
 - EXT:solr Version: 12.0.1

dmitryd:
# What this pr does

This PR adds a check that solr site exists for the given mount point.

# How to test

See the ticket

Fixes: #4337


dkd-kaehm:
@dmitryd 
is it still required in main?
We want a backports up to 12.1

dmitryd:
@dkd-kaehm  We did this change for version 12, but the same code exists in 14. So I think yes, it should go to the main branch too. I had an exception without this patch in 14 too.

Issue #4341, opened 2025-04-01
[FEATURE] Autosuggest requiest count limitation


dkd-kaehm:
**Is your feature request related to a problem? Please describe.**

If the standard EXT:solr templates are used and the auto-suggest function is enabled, the server may be overloaded with requests and become overwhelmed.

If the search input field has the focus and something is accidentally placed on the keyboard, causing a key with auto-repeat to be pressed, the input field will fill with an endless string, for which a request is sent to the server after each change (=each additional character). These requests are generated faster than they can be processed => the server becomes overloaded.

**Describe the solution you'd like**
Limit the input field with maxlength="50".
After 50 characters, no further changes are made to the search text, and no further requests for search suggestions are made.

Please implement this limit in the extension's default templates
or restrict/allow the AJAX script to be restricted (e.g., minimum/maximum number of characters for autosuggest, minimum time interval for requests, etc.).

**Describe alternatives you've considered**
Do the limitation on server side...

**Additional context**
Internal Redmine: `#215471`

**Target versions**
All supported

PR #4342, opened 2025-04-01
#4341 Form.html: maxlength for search text input field

pi-phi:
# What this pr does

inserting `maxlength="50"` to input field of search form 

# How to test

focus a search field with active auto suggest function. hold down the space key and fill the input field with spaces.
before:
watch the server log as requests with longer and longer search string appear until the server is locked up
after: 
just a few request appear as after 50 characters no further changes to the input text appear and trigger further AJAX requests

Fixes: #4341


dkd-kaehm (approved):
Thanks for this Pull-request

Issue #4353, opened 2025-05-07
[BUG] Indexing Queue is filled with wrong items on nested root pages


calien666:
**Describe the bug**
When setting up TYPO3 with nested Root Pages (pages is_siteroot=1), the Initializer does not stop on the next root page and queues the items from another root page with different Site Configuration. This leads to an item queue with items couldn't get indexed, as the Site Configuration does not match the given item.

**To Reproduce**

Test CSV:
```csv
"pages"
,"uid","pid","is_siteroot","doktype","slug","title"
,1,0,1,1,"/","Page 1"
,2,1,0,1,"/subpage","Subpage"
,3,1,1,1,"/","Page 2 (nested)"
,4,3,0,1,"/subpage","Subpage"
```

Result after index queue both sites:

```csv
"tx_solr_indexqueue_item"
,"uid","root","item_type","item_uid","indexing_configuration","indexed"
,1,1,"pages",1,"pages",0
,2,1,"pages",2,"pages",0
,3,1,"pages",3,"pages",0
,4,1,"pages",4,"pages",0
,5,3,"pages",3,"pages",0
,6,3,"pages",4,"pages",0
```

Expected result:
```csv
"tx_solr_indexqueue_item"
,"uid","root","item_type","item_uid","indexing_configuration","indexed"
,1,1,"pages",1,"pages",0
,2,1,"pages",2,"pages",0
,4,3,"pages",3,"pages",0
,5,3,"pages",4,"pages",0
```

The uids could differ, but in the given result, the IDs 3 and 4 should not appear.

**Expected behavior**
Respectation of nested root pages and stopping index queue writing then.

**Screenshots**
None

**Used versions (please complete the following information):**
 - TYPO3 Version: 11.5.41
 - Browser: CLI and Testing Framework
 - EXT:solr Version: 11.6.0
 - Used Apache Solr Version: 8.11.4
 - PHP Version: 8.2.28
 - MySQL Version: MariaDB 10.3.39

**Additional context**
Even if I reported this issue regarding to 11 (which is out of support), I will give fast information, if this behaviour is still a problem inside TYPO3 12 and 13.

dkd-kaehm:
@calien666 
thanks for reporting that issue.

[...]

PR #4355, opened 2025-05-08
Bugfix/4012 Correct closest page with template in rootline logic and refactor repository

### LINKED ISSUE #4012: [BUG] ConfigurationPageResolver::getClosestPageIdWithActiveTemplate relies on IN() to force a sort order (which it does not)
**Describe the bug**
The method returns whatever the query returns as first row. This is not guaranteed to be the first value in the list within ÌN()`.

**Expected behavior**
The page id that really is the first one with a sys_template record within the rootline.

**To Reproduce**
I'm not sure how an arbitrary result can be reproduced.

I noticed the behaviour when I initialized the queue for a site that is inside another one (microsite). Also the outer site is inside a page that holds a sys_template record (not a site itself, just a page to hold TypoScript defaults for a number of sites).

Some suggest to use `FIELD()` but  I don't know if this will work with DBAL/Doctrine:
```
SELECT * FROM t WHERE x IN (3,2,1) ORDER BY FIELD(x,3,2,1)
```

wazum:
# What this PR does

This PR addresses issue #4012 by:

**Correcting the logic for finding the closest page with an active system template within a given rootline.**
    *   Ensured that the rootline data provided to the `SystemTemplateRepository` is interpreted correctly (root page first, current page last).
    *   Verified that the sorting logic correctly identifies the template on the current page or its nearest ancestor as the "closest".
    *   Updated unit tests to reflect the correct rootline structure and expected outcomes.

# How to test

1.  Verify that the integration tests in `SystemTemplateRepositoryTest.php` pass.
2.  Manually test scenarios where a TYPO3 page tree has system templates on various levels of the rootline. Ensure that the system correctly identifies the template from the closest parent page in the rootline that has a template.
    *   Example:
        *   Root Page (no template)
            *   Level 1 Page (no template)
                *   Level 2 Page (has template A)
                    *   Current Page (has template B) -> Template B should be found.
        *   Root Page (no template)
            *   Level 1 Page (has template C)
                *   Level 2 Page (no template)
                    *   Current Page (no template) -> Template C should be found.

Fixes: #4012

### DIFF
--- a/Classes/System/Records/SystemTemplate/SystemTemplateRepository.php
+++ b/Classes/System/Records/SystemTemplate/SystemTemplateRepository.php
@@ -18,6 +18,7 @@ declare(strict_types=1);
 namespace ApacheSolrForTypo3\Solr\System\Records\SystemTemplate;
 
 use ApacheSolrForTypo3\Solr\System\Records\AbstractRepository;
+use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Exception as DBALException;
 
 /**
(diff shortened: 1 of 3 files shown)


dkd-kaehm:
The example in that pull-requiest is not the same as in issue it is reversed(roots without templates in top and no-roots with templates on bottom).

To your issue, which should work currently without changes:
EXT:solr relies and uses the `is_siteroot` PIDs everywhere(beginning by `tx_solr_indexqueue_item`) and tries to initialize the TSFE to use the TypoScript, so the last chain link with TypoScript must always be the `is_siteroot`-Page with template. 

EXT:solr requires a root page with template to work at all.
So you can assign a "empty or small" template to each of your micro-sites(whose root pages are marked as `is_siteroot`) and apply a whole TypoScript on some shared parent-page(try as non-root first, and if it works don't switch to `is_siteroot` because of #4353)

@wazum Please try the suggestion without your change and with.

wazum:
Hm, that's exactly my problem as described in the issue:

> I noticed the behaviour when I initialized the queue for a site that is inside another one (microsite). Also the outer site is inside a page that holds a sys_template record (not a site itself, just a page to hold TypoScript defaults for a number of sites).

- the container on top has a template (with general settings for all microsites/tenants) and no "Use as Root Page" set
- all microsites/tenants have a template and have "Use as Root Page" set (they are associated with a site/domain)

And this is solved with this pull request.
If you take a look at the original code, you see the problem instantly … it's just a query for all pages with a template and then returns any randomly (and not the one that is closest).
Please explain what is unclear here.

PR #4357, opened 2025-05-09
Update hardcoded legacy css resource filepath

chrrynobaka:
The existing file path wouldn't work with composer installation and the directory is not correct either.
This pull request updates the path to use an extension-relative EXT: path instead.


dkd-kaehm:
@chrrynobaka 
Nice catch. Thanks for fixing that issue.
It must be ported to release-12.0.x as well, because of same folder structures on TYPO3 12.

PR #4360, opened 2025-05-15
[TASK] Extend RecordUpdatedEvent to indicate creations

dkd-friedrich:
# What this pr does

Adds flag `$isNewRecord` to `RecordUpdatedEvent`, allowing to differ between record creations or updates.

Note: For compatibility reasons event RecordUpdatedEvent is extended, in v14 a separate event will be introduced and has to be handled.


dkd-kaehm (approved):
:100:

Issue #4371, opened 2025-06-04
[FEATURE] allow non-standard SOLR_HOME for containers with empty volumes


morph027:
**Is your feature request related to a problem? Please describe.**

Containers where the attached volume is empty and is not a bnd mount won't start. This came up while running the container image on AWS ECS Fargate which does not copy the contents of a Dockefile `VOLUME` to the mountpoint (EFS in this case) - https://github.com/aws/containers-roadmap/issues/863

**Describe the solution you'd like**

Copy the contents of `/var/solr/data` to another location where a volume is attached and use this location as `SOLR_HOME`

**Describe alternatives you've considered**

Building a custom image works.

**Additional context**

When starting with an empty volume:

```
docker run --rm -it --tmpfs /var/solr:uid=8983,gid=8983 typo3solr/ext-solr:13.0
/opt/solr/docker/scripts/solr-foreground: running /docker-entrypoint-initdb.d/0_as-sudo-tweaks.sh
/opt/solr/docker/scripts/solr-foreground: running 'sudo /docker-entrypoint-initdb.d-as-sudo/fix-CVE-2025-24814.sh'
grep: /var/solr/data/configsets/ext_solr_13_0_0/conf/solrconfig.xml: No such file or directory

/opt/solr/docker/scripts/solr-foreground: running 'sudo /docker-entrypoint-initdb.d-as-sudo/fix-file-permissions.sh'


/opt/solr/docker/scripts/solr-foreground: running /docker-entrypoint-initdb.d/disable-cores.sh

Starting Solr
Solr will start in SolrCloud mode by default in version 10, and you will have to provide --user-managed if you want to stay on the user-managed (aka. standalone) mode.

Java 17 detected. Enabled workaround for SOLR-16463
[0.001s][warning][pagesize] UseLargePages disabled, no large pages configured and available on the system.
CompileCommand: exclude com/github/benmanes/caffeine/cache/BoundedLocalCache.put bool exclude = true
WARNING: A command line option has enabled the Security Manager
WARNING: The Security Manager is deprecated and will be removed in a future release
2025-06-04 18:43:30.047 INFO  (main) [c: s: r: x: t:] o.e.j.s.Server jetty-10.0.22; built: 2024-06-27T16:03:51.502Z; git: 5c8471e852d377fd726ad9b1692c35ffc5febb09; jvm 17.0.15+6
[...]

PR #4372, opened 2025-06-04
[FEATURE]: allow non-standard SOLR_HOME for containers with empty volumes

morph027:
# What this pr does

Populates non-standard `SOLR_HOME` for containers where the attached volume is empty. This came up while running the container image on AWS ECS Fargate which does not copy the contents of a Dockefile `VOLUME` to the mountpoint (EFS in this case) - https://github.com/aws/containers-roadmap/issues/863

Workaround is to set a different `SOLR_HOME` (even if not advised by upstream Apache Solr) and copy the contents of the image to this location.

This only affects container runtimes other than Docker (e.g. containerd) and mounts which are not bind mounts.

# How to test

* docker image build locally `docker buildx build --load -t typo3solr/ext-solr:dev -f Docker/SolrServer/Dockerfile .`
* run with different `SOLR_HOME`:  `docker run --rm -it --tmpfs /var/lib/solr:uid=8983,gid=8983 -e SOLR_HOME=/var/lib/solr typo3solr/ext-solr:dev`

Fixes: #4371

### DIFF
--- a/Docker/SolrServer/docker-entrypoint-initdb.d/0_SOLR_HOME.sh
+++ b/Docker/SolrServer/docker-entrypoint-initdb.d/0_SOLR_HOME.sh
@@ -8,7 +8,7 @@ copy_to_volume() {
   exit 0
 }
 
-if [ "${SOLR_HOME}" != "/var/solr/data" ] && [ ! -f "${SOLR_HOME}/solr.xml" ]; then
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
See: https://github.com/TYPO3-Solr/ext-solr/issues/4371#issuecomment-3032137937

Issue #4374, opened 2025-06-26
[TASK] Drop workaround for cObj IMAGE


dkd-friedrich:
**What should be done in the scope of this task?**

Since the first versions of EXT:solr the `AbstractIndexer` contains a workaround allowing to use IMAGE cObj in indexing configuration. To resolve an issue determining the image file, the current work directory is changed to the public web folder.

This workaround is no longer required and may cause warnings about missing permissions depending on the setup, thus it should be removed.

---

Target versions: 13, 12

PR #4377, opened 2025-07-07
[TASK] Drop workaround for cObj IMAGE

dkd-friedrich:
# What this pr does

Since the first versions of EXT:solr the AbstractIndexer contains a workaround allowing to use IMAGE cObj in indexing configuration. To resolve an issue determining the image file, the current work directory is changed to the public web folder.

This workaround is no longer required and thus removed.

Resolves: #4374


dkd-kaehm (approved):
FIN

Issue #4381, opened 2025-07-10
[BUG] SOLR_RELATION breaks if no relations exist


koehnlein:
**Describe the bug**

If no references are found, indexing stops with an error:

```
1476107295: TYPO3\CMS\Core\Error\Exception: PHP Warning: Trying to access array offset on null in /var/www/html/vendor/apache-solr-for-typo3/solr/Classes/ContentObject/Relation.php line 333 in /var/www/html/vendor/typo3/cms-core/Classes/Error/ErrorHandler.php:141
```

**Expected behavior**
Indexing should not stop and the document should be created with the remaining properties

**Screenshots**

<img width="1600" height="156" alt="Image" src="https://github.com/user-attachments/assets/45668e4b-4f59-493a-87a8-3eab5f790161" />

**Used versions (please complete the following information):**
 - TYPO3 Version: 13.4.15
 - EXT:solr Version: 13.0.2
 - Used Apache Solr Version: 9.8.1
 - PHP Version: 8.4.7
 - MySQL Version: 10.11.11-MariaDB-ubu2204-log

PR #4382, opened 2025-07-10
Avoid PHP warning if related record was not found

koehnlein:
# What this pr does

Set `''` as third parameter `$defaultValue` for `$contentObject->stdWrapValue()` as fallback

Fixes: #4381


dkd-friedrich (approved):
Thanks for fixing

PR #4391, opened 2025-08-01
[BUGFIX] Ensure rootline order is respected detecting closest template

### LINKED ISSUE #4012: [BUG] ConfigurationPageResolver::getClosestPageIdWithActiveTemplate relies on IN() to force a sort order (which it does not)

calien666:
# What this pr does

When detecting the closest template inside a rootline, it is necessary, respecting the order of the rootline itself.

The current database call has a random sorting, which does not respect the ordering, therefore, could return a wrong dataset if multiple templates are given in the root line.

To ensure the order is respected, every page inside rootline is called single against the database and returns the pid if a template record exists.

# How to test

Given a rootline with multiple siteroots, the database query does not respect the order of the elements and only returns the first hit form the database, which could be the right looked up, but not has to be.

Example build:

```
0
|- 165 (doktype 254)
 |- 271 (doktype 1, is siteroot, has root template)
  |- 348 (doktype 254)
   |- 396 (doktype 254)
    |- 491 (doktype 1, is siteroot, has root template)
     |- 29455 (doktype 254)
      |- 29456 (doktype 254, Records located here)
```

The system calls with the rootline:
```
7 => 29456
6 => 29455
5 => 491
4 => 396
3 => 348
2 => 271
1 => 165
```

And the database always returns `271` as the found pid, which is wrong.

Fixes: #4012

### DIFF
--- a/Classes/System/Records/SystemTemplate/SystemTemplateRepository.php
+++ b/Classes/System/Records/SystemTemplate/SystemTemplateRepository.php
@@ -35,18 +35,29 @@ class SystemTemplateRepository extends AbstractRepository
      */
     public function findOneClosestPageIdWithActiveTemplateByRootLine(array $rootLine): ?int
     {
-        $rootLinePageIds = [0];
+        // rootline always is given in reverse order, so iterate over, as it will go to top of rootline
+        // from the current called pid
         foreach ($rootLine as $rootLineItem) {
-            $rootLinePageIds[] = (int)$rootLineItem['uid'];
+            $foundPage = $this->getTemplateIdForGivenPageId($rootLineItem['uid']);
+            if ($foundPage !== null) {
+                return $foundPage;
+            }
         }
+        return null;
+    }
 
+    /**
+     * @throws DBALException
+     */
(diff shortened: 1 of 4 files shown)


calien666:
Additional information: I detected this issue already int EXT:solr 11, as we are working on a TYPO3 v11, but the logic is the same for later versions.

Issue #4409, opened 2025-09-06
[BUG] Indexing of hidden translated records for solr 13


ulrike-cosmoblonde:
On a TYPO3 v13 project using solr 13.0 (13.0.3 currently) all hidden records are indexed.
The websites languages are configured in strict mode.
If the original record is active, also the translated record gets indexed.
This happens on full indexing of a record type (I tested for several different types, like news ) as well as after any changes on the original record.
If the hidden record is re-saved, it is removed from the index.

Can you please check and fix?

Thank you!

ulrike-cosmoblonde:
The problem seems to be in file apache-solr-for-typo3/solr/Classes/IndexQueue/Indexer in the function getItemRecordOverlayed, which return any existing record overlay without checking for enabled settings.
I have created a patch file, which works for my project - maybe someone can confirm this error and check if the patch works

[fix-indexing-of-hidden-overlay-records.patch](https://github.com/user-attachments/files/22405985/fix-indexing-of-hidden-overlay-records.patch)

PR #4427, opened 2025-10-03
[BUGFIX] Prevent indexing of hidden records

honsa:
Add visibility check for item records based on hidden property.

# What this pr does
Fix issue https://github.com/TYPO3-Solr/ext-solr/issues/4409
Check if record is hidden and not add to document if the record is hidden

# How to test
Index record which has a hidden translation


Classes/IndexQueue/Indexer.php
```diff
@@ -377,7 +377,10 @@ protected function itemToDocument(Item $item, int $language = 0): ?Document
         $document = null;
 
         $itemRecord = $this->getFullItemRecord($item, $language);
-        if (!is_null($itemRecord)) {
+
+        $isVisible = ((is_array($itemRecord) && empty($itemRecord['hidden'])));
```
dkd-kaehm, line 381: Please see the patch file of https://github.com/TYPO3-Solr/ext-solr/issues/4409#issuecomment-3307127824
The TCA defintion for "hidden" must be used instead of hard-coded `['hidden']` array access key.

Issue #4451, opened 2025-10-31
[BUG] Incorrect view.available templates in multi site TYPO3 installation


WebsiteDeveloper:
**Describe the bug**
In a multi page TYPO3 installation, if both pages have different setups for available templates the current code would always use the first pages config and does not allow for different template configurations for each site.

**To Reproduce**
1. Create a TYPO3 installation with 2 sites configured
2. Create a template for each site with different Configurations for available templates
3. Create a page on each site with a the search plugin
4. Only the first sites configuration is used

**Expected behavior**
Allow different configurations based on included template in the site.

**Used versions (please complete the following information):**
 - TYPO3 Version: 12.4.38
 - Browser: firefox
 - EXT:solr Version: 12.0.7

PR #4452, opened 2025-10-31
[BUGFIX] pass a request with page id to Configuration manager

WebsiteDeveloper:
# What this pr does

This fixes an issue where, in a multi page TYPO3 installation, if both pages have different setups for available templates the previous code would always use the first pages config

# How to test

1.  Create a TYPO3 installation with 2 sites configured
2.  Create a template for each site with different Configurations for available templates
3.  Create a page on each site with a the search plugin
4. The available templates for each site are different

Fixes: #4451


WebsiteDeveloper:
Yes we are currently using it in a production v12 instance where it gets patched in, so should be backportable without issues.

dkd-kaehm (approved):

Issue #4419, opened 2025-09-26
[BUG] Can't remove facet via condition


goldi42:
I discovered a problem with **the** TypoScript Configuration. 

If you have **facets** globally defined, for example the type **facet**, and try to remove the facet inside a TypoScript Condition, for example **for** a **news record** search, the TypoScript inside the Plugin shows the "removed" facet.  
It **seems** like the [ConfigurationManager](https://github.com/TYPO3-Solr/ext-solr/blob/main/Classes/System/Configuration/ConfigurationManager.php#L54) doesn't use the evaluated TypoScript from **the** TYPO3 Frontend and so you can't unset the facet for the plugin.  
I tried to load the TypoScript inside another plugin and there the TypoScript was evaluated correctly, so **it seems** that this is not a TYPO3 Bug.

**To Reproduce**

1. Make sure you have a facet globally defined and *it is shown on the global search
2. **Create** a new Page and add the result plugin. Add TypoScript with a condition for this page  
    **and** remove the facet like `plugin.tx_solr.search.faceting.facets.type >`
3. Open the page and start the search.
4. The facet will still be rendered.

**Expected behavior**

The facet should not be rendered on the page with the condition.

**Used versions (please complete the following information):**

- TYPO3 Version: 13.4.17

- Browser: Chorome

- EXT:solr Version: 13.0.3

- Used Apache Solr Version: 9.8.1

- PHP Version: 8.3

- MySQL Version: MariaDB 10.3

dkd-kaehm:
@goldi42 Please check #4340

dkd-kaehm:
Fixed in #4376.

Issue #2333, opened 2019-07-10
[FEATURE] Make facet configurations disengageable


Mabahe:
**Is your feature request related to a problem? Please describe.**
Once a facet is configured in `plugin.tx_solr.search.faceting.facets.` it could not be removed later for different configurations, based e.g. on conditions. Tried:
* `facetName >`
* `facetName\. >`
* `facetName = __UNSET`
* `facetName\. = __UNSET`
in different combinations.
It only helps creating own content elements or to include the facet configurations from extra files with a condition, e.g.: 
```
<INCLUDE_TYPOSCRIPT: source="FILE:EXT:[...]/optional/setup_solr_facet_type.ts" condition="[globalVar = TSFE:id = 23|42]">

**Describe the solution you'd like**
Having an option like: `plugin.tx_solr.search.faceting.facets.facetName.disabled = 1`

**Describe alternatives you've considered**
Checking TS parsing to make an unset work.

PR #4376, opened 2025-07-04
[BUGFIX] Don't re-instantiate TypoScript in FE/Search context

dkd-kaehm:
Fixes: #4340, #4419

PR #4473, opened 2025-11-11
[BUGFIX] Add setting to disabled facets

BastiLu:
There is an issue resolving the typoscript configuration properly. Currently there is no way to override the default tx_solr configuration in the context of another content element/plugin. The problem occurs in the AbstractBaseController initializeAction. If e.g. __UNSET is used to unset a facet this is properly done in line 169. As intended, the facet is removed, but also the __UNSET rule. In line 179 mergeSolrConfiguration() the facet is added again from the "default" plugin configuration as the __UNSET rule is now removed. This might affect other settings than facets too (e.g. sorting.options).

Resolves: #4419 #2333

# What this pr does

Please add a description here

# How to test

### DIFF
--- a/Classes/Controller/AbstractBaseController.php
+++ b/Classes/Controller/AbstractBaseController.php
@@ -179,7 +179,7 @@ abstract class AbstractBaseController extends ActionController
             $this->typoScriptConfiguration->mergeSolrConfiguration(
                 $typoScriptService->convertPlainArrayToTypoScriptArray($pluginSettings),
                 true,
-                false
+                true
             );
         }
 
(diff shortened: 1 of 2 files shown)


dkd-kaehm:
@BastiLu
Thanks for fixing that issue, but this branch isn't supported anymore and is locked in this repository. 
The branch is moved on non-public ELTS repository for security issues only.

---

Is the change on `Classes/Controller/AbstractBaseController.php` for TYPO3 12 and 13 relevant?
If so, please create a new issue with description and steps to reproduce, so we can add the integration tests.
Please note: Unsetting of TypoScript configurations should work now in TYPO3 12 since EXT:solr 12.0.0 and in 13 dev state since #4376.

PR #4501, opened 2025-12-10
[BUGFIX] Use the return value of the UPDATE statement correctly in the `LastSearchesRepository::update` method

RZivaralam:
To Reproduce

This confirms that the exception is triggered in valid and expected situations, for example when:

multiple users perform the same search query at nearly the same time, or

the same search term is executed repeatedly within the same second.

In these cases:

the UPDATE statement does find the row (correct sequence_id)

but writes identical values (keywords and second-based tstamp)

MySQL / MariaDB correctly returns affectedRows = 0

the current logic incorrectly interprets this as “row not found”

### DIFF
--- a/Classes/Domain/Search/LastSearches/LastSearchesRepository.php
+++ b/Classes/Domain/Search/LastSearches/LastSearchesRepository.php
@@ -139,7 +139,7 @@ class LastSearchesRepository extends AbstractRepository
             ->set('keywords', $lastSearchesRow['keywords'])
             ->executeStatement();
 
-        if ($affectedRows < 1) {
+        if ($affectedRows === false) {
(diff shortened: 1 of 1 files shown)


dkd-kaehm:
@RZivaralam 
Is there a way to add an integration test case for that?
I still do not get why we need that.

RZivaralam:
To reproduce the issue:

1. Set plugin.tx_solr.search.lastSearches.mode = global, so the last searches are stored in the database.
2. Search for the same keyword several times in the frontend.
3. After a few searches, the update statement does not update any row anymore and returns 0 affectedRows < 1
4. The current check interprets this as if the column `sequence_id` does not exist.
5. But in this case, the column exists. The real meaning is only that there was no new update to perform.

So the problem is not that `sequence_id` is missing. The problem is that the current `if` condition interprets a “no update” result as “column does not exist”.

In my opinion, this additional exception is not necessary. If the column `sequence_id` really does not exist, the database layer would already throw an SQL error like `Unknown column 'sequence_id'`.

So the check should either be removed or changed, because affectedRows < 1 here does not necessarily mean that the column is missing.

PR #4514, opened 2026-01-20
[BUGFIX] Filter Solr connections based on language fallback settings

### LINKED ISSUE #4513: [BUG] PageIndexer filterSolrConnectionByPage does not respect site language fallbacks
**Describe the bug**
When flag 'Hide default language of page' is set and a language's only fallback language is 0, the page should not be indexed because it does not exist.

**To Reproduce**
Create a Site having 3 languages. Language with id 1 should have fallbackType: strict and fallbacks: '0'. Create a page and set the flag 'Hide default language of page' then just create one translation for language with id 2.

**Expected behavior**
Page with language id 1 should not be indexed.

 - TYPO3 Version: 13.4.23
 - EXT:solr Version: 3.1.0

siwa-ddiesenreither:
# What this pr does

Added a check if then solrConnections language has only 0 as fallback language - if so and the default language is hidden also hide this language.

# How to test

See description of issue.

Fixes: #4513

### DIFF
--- a/Classes/IndexQueue/PageIndexer.php
+++ b/Classes/IndexQueue/PageIndexer.php
@@ -28,6 +28,7 @@ use Exception;
 use Psr\Log\LogLevel;
 use RuntimeException;
 use TYPO3\CMS\Backend\Utility\BackendUtility;
+use TYPO3\CMS\Core\Site\SiteFinder;
 use TYPO3\CMS\Core\Type\Bitmask\PageTranslationVisibility;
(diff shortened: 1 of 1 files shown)


dkd-kaehm (changes requested):
Thanks for fixing that issue.
Please fix the static-analysis issues:

https://github.com/TYPO3-Solr/ext-solr/actions/runs/21178675347/job/60916087419?pr=4514#step:10:69

PR #4589, opened 2026-03-10
[TASK] Refactor FrontendAwareEnvironment and adjust tests

froemken:
- Refactor `FrontendAwareEnvironment` to use constructor property promotion.
- Remove unused `SingletonInterface` implementation and related usages.
- Replace `GeneralUtility::makeInstance` calls with dependency injection.
- Add `frontend_simulation` service configuration in `Services.yaml`.
- Simplify `TsfeTest` by using an explicitly instantiated `FrontendAwareEnvironment`.

### DIFF
--- a/Classes/FrontendSimulation/FrontendAwareEnvironment.php
+++ b/Classes/FrontendSimulation/FrontendAwareEnvironment.php
@@ -32,7 +32,6 @@ use TYPO3\CMS\Core\Exception\SiteNotFoundException;
 use TYPO3\CMS\Core\Http\ServerRequest;
 use TYPO3\CMS\Core\Localization\Locales;
 use TYPO3\CMS\Core\Routing\PageArguments;
-use TYPO3\CMS\Core\SingletonInterface;
 use TYPO3\CMS\Core\Site\Entity\Site;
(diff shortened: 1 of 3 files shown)


Classes/FrontendSimulation/FrontendAwareEnvironment.php
```diff
@@ -213,26 +207,16 @@ public function getServerRequestByPageIdIgnoringLanguage(int $pageId): ?ServerRe
...
         }
-        return $this->getServerRequestByPageIdAndLanguageFallbackChain($pageId, ...$availableLanguageIds);
-    }
 
-    /**
-     * Returns the page ID from a ServerRequest.
-     *
-     * Convenience method to extract the page ID from the PageInformation attribute.
-     */
-    public function getPageIdFromRequest(ServerRequest $request): int
```
froemken: Sure this is not needed anymore? It's not part of the commit message and in this patch this method was only found ones. I don't see from where it is/was called.

dkd-kaehm: Please decide. 
If you ask me: I did not change anything on that branch.

PR #4618, opened 2026-04-27
[BUGFIX] Correct field name casing for subTitle and navTitle in TypoScript queryFields

### LINKED ISSUE #4617: [BUG] subTitle and navTitle fields not searched due to case mismatch in TypoScript queryFields
**Describe the bug**
Searching does not return results matching content stored in the subTitle and navTitle fields, even though both fields are correctly indexed in Solr. The root cause is a case mismatch between the Solr schema and the TypoScript configuration. The schema (general_schema_fields.xml) defines the fields as subTitle and navTitle (camelCase), but the TypoScript references them as subtitle and navtitle (all lowercase). Since Solr field names are case-sensitive, the fields are indexed correctly but never included in the actual search query.

**To Reproduce**
1. Index a page that has a subtitle set.
2. Search for a keyword that only appears in the subtitle.
3. No results are returned.  

**Expected behavior**
 Search results include documents where the search term matches content in subTitle or navTitle.

**Additional context**
Configuration/TypoScript/Solr/setup.typoscript references subtitle and navtitle in four places — queryFields, phrase.fields, bigramPhrase.fields, and trigramPhrase.fields — while the actual field names defined in Resources/Private/Solr/configsets/ext_solr_14_0_0/conf/general_schema_fields.xml are subTitle (line 101) and navTitle (line 102).

amirarends:
The Solr schema defines subTitle and navTitle as camelCase field names, but the TypoScript configuration referenced them as subtitle and navtitle (all lowercase). Since Solr field names are case-sensitive, both fields were indexed correctly but silently excluded from all search queries.

# What this pr does

Corrects the field name casing in `Configuration/TypoScript/Solr/setup.typoscript` for all four field lists: `queryFields`, `phrase.fields`, `bigramPhrase.fields`, and `trigramPhrase.fields`.

# How to test

1. Index a page that has a subtitle set.
2. Search for a keyword that only appears in the subtitle.
3. Verify that the page appears in the search results.

Fixes: #4617

### DIFF
--- a/Configuration/TypoScript/Solr/setup.typoscript
+++ b/Configuration/TypoScript/Solr/setup.typoscript
@@ -78,7 +78,7 @@ plugin.tx_solr {
       allowedSites = __solr_current_site
 
       // qf parameter http://wiki.apache.org/solr/DisMaxQParserPlugin#qf_.28Query_Fields.29
-      queryFields = content^40.0, title^5.0, keywords^2.0, tagsH1^5.0, tagsH2H3^3.0, tagsH4H5H6^2.0, tagsInline^1.0, description^4.0, abstract^1.0, subtitle^1.0, navtitle^1.0, author^1.0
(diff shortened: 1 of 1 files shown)


dkd-kaehm (approved):
Merci.
I created the backports, all PRs will be merged at once.

Issue #4031, opened 2024-05-08
solrcloud is  compatible with version 12???


RZivaralam:
hey, I have a question : 
is  Solrcloud with typo3 version 12 compatible or not?
I used Solrcloud but I have the following errors :( in loggin and I don't have any connection to Solr in typo3 info :(

8.5.2024, 13:05:31 | ERROR false |   | CollectionHandlingUtils | Collection core_en,&#8203; operation create failed
8.5.2024, 13:05:31 | ERROR true |   | RequestHandlerBase | Client exception
8.5.2024, 13:07:19 | ERROR false |   | RequestHandlerBase | Server exception
8.5.2024, 13:07:19 | ERROR false |   | HttpSolrCall | 500 Exception
8.5.2024, 13:07:19 | ERROR false |   | HttpSolrCall | 500 Exception
8.5.2024, 13:07:19 | ERROR false |   | RequestHandlerBase | Server exception
8.5.2024, 13:07:19 | ERROR false |   | HttpSolrCall | 500 Exception

dkd-dobberkau:
Currently, it only works with Apache Solr Server in Standalone Mode.

antonio-masotti:
Are there any plans to make TYPO3 12 compatible with Solr Cloud?

dkd-dobberkau:
We have no plans at the moment. 
What is your use case?

PR #4621, opened 2026-04-29
[FEATURE] Support solrcloud mode

schliesser:
# What this pr does

Add support for solrcloud.

When solr is running in `solrcloud` mode, it uses collections as the main entrypoint instead of cores. A collection exists of multiple cores, that can be spread across different machines. `solrcloud` is typically used in cloud hostes environments for scalability.

From the extension perspective, I only had to change the `reloadCoreByName()` method to select the correct API endpoint depending of the mode.

In addition the status page now shows the solr mode.

Functions that work without changes:

- Fetching informations about solr server
- Indexing
- Searching (including Autosuggest)
- Clearing the index
- Index fields info
- Adding/Removing Synonyms
- Adding/Removing Stop Words

# How to test

Will update here as soon as the ddev add-on ddev-typo3-solr is updated. I'm currently working on supporting both modes  (standalone / solrcloud) in this add-on for easy testing.

Fixes: #4031

### DIFF
--- a/Classes/Report/SolrStatus.php
+++ b/Classes/Report/SolrStatus.php
@@ -119,6 +119,7 @@ class SolrStatus extends AbstractSolrStatus
             ->getAdminService();
 
         $solrVersion = $this->checkSolrVersion($solrAdmin);
+        $solrMode = $this->checkSolrMode($solrAdmin);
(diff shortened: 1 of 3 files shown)


dkd-kaehm:
@schliesser 
Thanks for that PR
Why not against main branch with Apache Solr 10?

Issue #3115, opened 2021-12-08
[BUG] Plugin "PostBigRequest" has no effect


Smove:
**Describe the bug**
Since TYPO3 has integrated its own EventDispatcher and EventListenerProvider, the plugin "PostBigRequest" no longer has any effect. Problem is, that in the plugin itself in the "initPluginType" function a check of the EventDispatcher for the class from the Symfony components is done and the function for the "PRE_EXECUTE_REQUEST" event is not registered. The plugin "PostBigExtractRequest" throws directly error, because there is no query but simply assumed that the EventDispatcher is compatible. But in TYPO3 Dispatcher the method "addListener" does not exist.

If you register the event of the PostBigRequest itself via a Services.yaml file it works as desired.

**To Reproduce**
Add as many facets as necessary to get over the "GET" parameter limit. 

**Expected behavior**
Since MR#2476 it should automatically switch a search query to POST. However, this does not happen.

**Used versions (please complete the following information):**
 - TYPO3 Version: 10.4.32
 - EXT:solr Version: 11.1.2
 - Used Apache Solr Version: 8.9.0
 - PHP Version: 7.4.23

**Additional context**
The following content in your own Services.yaml fixes the problem:
```  Solarium\Plugin\PostBigRequest:
    tags:
      - name: event.listener
        method: 'preExecuteRequest'
        identifier: 'myext.solr.postbigrequestfix'
        event: Solarium\Core\Event\PreExecuteRequest
```

dkd-kaehm:
@Smove 
thanks for reporting.
Please transfer the suggestion to a pull-request.

Affected versions: 
* release-11.1.x
* release-11.5.x
* main

Smove:
Even though above is an example of a solution, I am not sure if this is the best solution. Better would probably be to build some kind of layer in which the functionality from TYPO3 dispatcher and the dispatcher expected in Solarium are combined.  This way other plugins are also compatible and not only this one.

dkd-kaehm:
OK,
it seems the POST requests per default are not recommended for Apache solr see https://github.com/solariumphp/solarium/issues/794#issuecomment-638697548

> In general switching to POST should be avoided wherever possible because you (might) bypass all caches.

That feature and other plug-ins MUST be configurable by EXT:solr. 
So we want to register all the solarium plug-ins(currently only once) via EXT:solr configuration flags only, preferably in site-yaml near to and for each solr connection.

---

[...]

PR #4631, opened 2026-05-04
[BUGFIX] Convert long Solr GET requests to POST on PSR-14 dispatcher

### LINKED ISSUE #4630: PostBigRequest plugin silently does nothing on TYPO3's PSR-14 dispatcher
## Summary

EXT:solr enables Solarium's `PostBigRequest` plugin so long Solr requests are converted from GET to POST. On TYPO3 this plugin never runs, because Solarium's plugin only registers its listener when the dispatcher is a Symfony `EventDispatcherInterface`. TYPO3's PSR-14 dispatcher does not implement that Symfony interface, so the listener is never attached and the plugin object becomes a silent no-op.

The result: any Solr request with a long URL (for example a request with many `facet.field` parameters on a large schema) hits HTTP 414 even though EXT:solr believes PostBigRequest will handle it.

## Where the problem comes from

EXT:solr injects the PSR-14 dispatcher into the Solarium client:

- `Configuration/Services.yaml`, line 111: https://github.com/TYPO3-Solr/ext-solr/blob/f0ec842133604265845d0f9f6d877dcd999d0781/Configuration/Services.yaml#L111
  ```yaml
  $eventDispatcher: '@Psr\EventDispatcher\EventDispatcherInterface'
  ```
- `Classes/System/Solr/SolrConnection.php`, lines 242-243: https://github.com/TYPO3-Solr/ext-solr/blob/f0ec842133604265845d0f9f6d877dcd999d0781/Classes/System/Solr/SolrConnection.php#L242-L243
  ```php
  $client = new Client($adapter, $this->eventDispatcher);
  $client->getPlugin('postbigrequest');
  ```

In Solarium 6.4.1 the plugin guards its registration on Symfony's interface (`vendor/solarium/solarium/src/Plugin/PostBigRequest.php`):

```php
protected function initPluginType()
{
    $dispatcher = $this->client->getEventDispatcher();
    if (is_subclass_of($dispatcher, '\Symfony\Component\EventDispatcher\EventDispatcherInterface')) {
        $dispatcher->addListener(Events::PRE_EXECUTE_REQUEST, [$this, 'preExecuteRequest'], 10);
    }
}
```

`TYPO3\CMS\Core\EventDispatcher\EventDispatcher` implements PSR-14 only and is not a subclass of the Symfony interface, so this `if` branch is skipped.

## Reproduction

PHP script using EXT:solr's installed Solarium (6.4.1) and the same TYPO3 PSR-14 dispatcher EXT:solr injects:

```php
<?php
require __DIR__ . '/.Build/vendor/autoload.php';

use Psr\Container\ContainerInterface;
use Psr\Http\Client\ClientInterface as Psr18Client;
use Psr\Http\Message\RequestFactoryInterface;
use Psr\Http\Message\RequestInterface as PsrRequest;
use Psr\Http\Message\ResponseInterface as PsrResponse;
use Psr\Http\Message\StreamFactoryInterface;
use Psr\Http\Message\StreamInterface;
use Solarium\Core\Client\Adapter\Psr18Adapter;
use Solarium\Core\Client\Client;
use Solarium\Core\Client\Endpoint;
use Solarium\Core\Client\Request;
use Solarium\Core\Event\PreExecuteRequest;
use TYPO3\CMS\Core\EventDispatcher\EventDispatcher;
use TYPO3\CMS\Core\EventDispatcher\ListenerProvider;

$container = new class implements ContainerInterface {
    public function get(string $id): mixed { return null; }
    public function has(string $id): bool { return false; }
};
$dispatcher = new EventDispatcher(new ListenerProvider($container));

$psr18 = new class implements Psr18Client {
    public function sendRequest(PsrRequest $r): PsrResponse { throw new RuntimeException('not used'); }
};
$requestFactory = new class implements RequestFactoryInterface {
    public function createRequest(string $m, $u): PsrRequest { throw new RuntimeException('not used'); }
};
$streamFactory = new class implements StreamFactoryInterface {
    public function createStream(string $c = ''): StreamInterface { throw new RuntimeException('not used'); }
    public function createStreamFromFile(string $f, string $m = 'r'): StreamInterface { throw new RuntimeException('not used'); }
    public function createStreamFromResource($r): StreamInterface { throw new RuntimeException('not used'); }
};
$client = new Client(new Psr18Adapter($psr18, $requestFactory, $streamFactory), $dispatcher);
$client->getPlugin('postbigrequest');

$request = new Request();
$request->setMethod(Request::METHOD_GET);
$params = ['q' => '*:*', 'rows' => '0', 'facet' => 'true'];
for ($i = 0; $i < 200; $i++) {
    $params['facet.field'][] = 'someFacetableField_' . $i;
}
$request->setParams($params);

echo 'before: method=' . $request->getMethod() . ' length=' . strlen($request->getQueryString()) . PHP_EOL;
$dispatcher->dispatch(new PreExecuteRequest($request, new Endpoint()));
echo 'after:  method=' . $request->getMethod() . ' length=' . strlen($request->getQueryString()) . PHP_EOL;
```

Output:

```
before: method=GET length=6936
after:  method=GET length=6936
```

The request stays GET. With a Symfony dispatcher the same code rewrites to POST.

## Proposed fix

Register a PSR-14 event listener for `Solarium\Core\Event\PreExecuteRequest` in EXT:solr that does the same conversion as the Solarium plugin (rewrite GET to POST when the query string is longer than a configurable threshold). The listener works directly with TYPO3's dispatcher.

I will open a pull request from my fork.

## Environment

- TYPO3 13.4
- EXT:solr main (f0ec8421)
- Solarium 6.4.1

wazum:
## What was changed

Adds a TYPO3 PSR-14 event listener for `Solarium\Core\Event\PreExecuteRequest` that converts long GET requests to POST. The listener mirrors what Solarium's `PostBigRequest` plugin does, but is registered through TYPO3's `#[AsEventListener]` attribute so it actually runs on TYPO3's PSR-14 dispatcher.

Solarium's plugin guards its listener registration on Symfony's `EventDispatcherInterface` ([`vendor/solarium/solarium/src/Plugin/PostBigRequest.php`](https://github.com/solariumphp/solarium/blob/6.4.1/src/Plugin/PostBigRequest.php#L99) line 99). TYPO3's PSR-14 dispatcher does not extend that Symfony interface, so the existing `$client->getPlugin('postbigrequest')` call in [`SolrConnection::getClient()`](https://github.com/TYPO3-Solr/ext-solr/blob/f0ec842133604265845d0f9f6d877dcd999d0781/Classes/System/Solr/SolrConnection.php#L243) line 243 is a silent no-op. Long Solr URLs then hit HTTP 414.

The new listener checks the request method and the query-string length and rewrites GET to POST when the threshold (1024 bytes by default, same as Solarium) is exceeded. The existing `$client->getPlugin('postbigrequest')` call in `SolrConnection` is left in place so the plugin still works in environments that use a Symfony dispatcher; the new listener fires once per request regardless and is idempotent (post-conversion the request method is already POST, so a second pass is a no-op).

## Files

- `Classes/EventListener/SolariumRequest/PostBigRequestListener.php` — the listener
- `Tests/Unit/EventListener/SolariumRequest/PostBigRequestListenerTest.php` — three unit tests (short GET stays GET, long GET becomes POST with body and cleared params, POST is left alone)

Fixes #4630

### DIFF
--- a/Classes/EventListener/SolariumRequest/PostBigRequestListener.php
+++ b/Classes/EventListener/SolariumRequest/PostBigRequestListener.php
@@ -1,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the TYPO3 CMS project.
+ *
+ * It is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License, either version 2
+ * of the License, or any later version.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.txt file that was distributed with this source code.
+ *
+ * The TYPO3 project - inspiring people to share!
+ */
+
+namespace ApacheSolrForTypo3\Solr\EventListener\SolariumRequest;
+
+use Solarium\Core\Client\Request;
+use Solarium\Core\Event\PreExecuteRequest;
+use TYPO3\CMS\Core\Attribute\AsEventListener;
+
+use function strlen;
+
+/**
+ * Bridges Solarium's PostBigRequest plugin onto TYPO3's PSR-14 event dispatcher.
+ *
+ * Solarium's PostBigRequest plugin only registers its listener when the dispatcher
+ * is a Symfony EventDispatcherInterface. TYPO3 ships a PSR-14 dispatcher that does
+ * not extend that Symfony interface, so the plugin call inside SolrConnection is a
+ * silent no-op. This listener performs the same conversion (GET to POST when the
+ * query string exceeds the threshold) and works directly with TYPO3's dispatcher.
+ *
+ * @noinspection PhpUnused Listener for {@link PreExecuteRequest}
+ */
+final readonly class PostBigRequestListener
+{
+    public const DEFAULT_MAX_QUERY_STRING_LENGTH = 1024;
+
+    public function __construct(
+        private int $maxQueryStringLength = self::DEFAULT_MAX_QUERY_STRING_LENGTH,
+    ) {}
+
+    #[AsEventListener(
+        identifier: 'solr.solarium.post-big-request',
+    )]
+    public function __invoke(PreExecuteRequest $event): void
+    {
+        $request = $event->getRequest();
+
+        if ($request->getMethod() !== Request::METHOD_GET) {
+            return;
+        }
+
+        $queryString = $request->getQueryString();
+        if (strlen($queryString) <= $this->maxQueryStringLength) {
+            return;
+        }
+
+        $charset = $request->getParam('ie') ?? 'utf-8';
+
+        $request->setMethod(Request::METHOD_POST);
+        $request->setContentType(Request::CONTENT_TYPE_APPLICATION_X_WWW_FORM_URLENCODED, ['charset' => $charset]);
+        $request->setRawData($queryString);
+        $request->clearParams();
+    }
+}
--- a/Classes/EventListener/SolariumRequest/PostBigRequestListener.php
+++ b/Classes/EventListener/SolariumRequest/PostBigRequestListener.php
@@ -17,6 +17,7 @@ declare(strict_types=1);
 
 namespace ApacheSolrForTypo3\Solr\EventListener\SolariumRequest;
 
+use ApacheSolrForTypo3\Solr\System\Configuration\ExtensionConfiguration;
 use Solarium\Core\Client\Request;
 use Solarium\Core\Event\PreExecuteRequest;
 use TYPO3\CMS\Core\Attribute\AsEventListener;
@@ -32,6 +33,9 @@ use function strlen;
  * silent no-op. This listener performs the same conversion (GET to POST when the
  * query string exceeds the threshold) and works directly with TYPO3's dispatcher.
  *
+ * Disabled by default; opt in via the "enablePostBigRequest" extension setting.
+ * Apache Solr does not cache POST requests, so enabling this can reduce cache hits.
+ *
  * @noinspection PhpUnused Listener for {@link PreExecuteRequest}
  */
 final readonly class PostBigRequestListener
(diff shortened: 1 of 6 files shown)


dkd-kaehm:
@wazum thanks for the PR!

Have you seen the earlier discussions about postBigRequest?
For example: https://github.com/TYPO3-Solr/ext-solr/issues/3115#issuecomment-1783262158

Back then the Solarium maintainers explicitly advised against enabling postBigRequest by default.
Do you know if their recommendation has changed since then?

wazum:
@dkd-kaehm 

  I checked the current state on the Solarium side and I think the situation is a bit different than it might look:

  - Solarium does not disable `PostBigRequest` by default in any policy sense. Its own Client class even auto-loads it for query types that typically produce long URLs (`createStream()`, `createGraph()`):

```
  // Streaming expressions tend to be very long. Therefore we use the 'postbigrequest' plugin.
  $this->getPlugin('postbigrequest');
```  

- The official Solarium documentation example for select queries also starts with `$client->getPlugin('postbigrequest');` and the comment "autoload the postbigrequest plugin". So the upstream recommendation is basically: turn it on if you might hit the limit.

The other thing I want to point out: [`SolrConnection::getClient()`](https://github.com/TYPO3-Solr/ext-solr/blob/f0ec842133604265845d0f9f6d877dcd999d0781/Classes/System/Solr/SolrConnection.php#L243) already calls `$client->getPlugin('postbigrequest')` (line 243). So the extension's code already says "we want this enabled". The problem is just that on TYPO3 **this call doesn't do anything**, because the dispatcher is not a _Symfony_ one. So this PR is not really changing a policy — it's making the existing code actually work as intended(?).

That said, I'm happy to make the listener configurable (e.g. a Site setting to disable it, or an option for the threshold or whatever). Just let me know which direction you prefer and I'll adjust the PR.

Thanks again for taking a look!

dkd-kaehm:
IMHO ext_conf_template.txt is excellent place for config.

dkd-kaehm:
Please make it configurable as suggested.
We are quite close to the stable release.
Please also add a note to the documentation stating that enabling it is still not recommended, as Apache Solr does not cache POST requests.

dkd-kaehm (approved):

PR #4644, opened 2026-05-08
[BUGFIX] fix suggestion query if routeEnhancer is set

### LINKED ISSUE #4635: [BUG] v14: suggest_controller.js appends query parameter with hard-coded "&" and breaks with PageType route enhancers
**Describe the bug**

The current `Resources/Public/JavaScript/suggest_controller.js` builds the suggest request URL like this:

```js
const url = `${this.suggestURL}&tx_solr%5BqueryString%5D=${encodeURIComponent(query)}`;
```

This assumes that `data-suggest` already contains a query string, for example:

```html
data-suggest="/?type=7384"
```

However, when TYPO3 uses a `PageType` route enhancer for the suggest `typeNum`, the generated suggest URL can be a clean path without any query string, for example:

```html
data-suggest="/de/search/suggest/"
```

Example site config:

```yaml
routeEnhancers:
  Suffix:
    type: PageType
    map:
      suggest/: 7384
```

In that case the current JS creates an invalid URL:

```text
/de/search/suggest/&tx_solr%5BqueryString%5D=test
```

The correct URL should be:

```text
/de/search/suggest/?tx_solr%5BqueryString%5D=test
```

**Expected behavior**

`suggest_controller.js` should append `tx_solr[queryString]` using `?` when `data-suggest` has no query string, and `&` when it already has one.

**Possible fix**

```js
const separator = this.suggestURL.includes('?') ? '&' : '?';
const url = `${this.suggestURL}${separator}tx_solr%5BqueryString%5D=${encodeURIComponent(query)}`;
```

Alternatively, the URL could be built with `URL` / `URLSearchParams`.

**Why this matters**

The current implementation works for URLs like `/?type=7384`, but breaks when TYPO3 routing turns the suggest `typeNum` into a clean URL through a `PageType` route enhancer.

dkd-lehnebach:
Fixes: #4635

### DIFF
--- a/Resources/Public/JavaScript/suggest_controller.js
+++ b/Resources/Public/JavaScript/suggest_controller.js
@@ -86,7 +86,8 @@ class SuggestController {
     this.abortController = new AbortController();
 
     try {
-      const url = `${this.suggestURL}&tx_solr%5BqueryString%5D=${encodeURIComponent(query)}`;
+      const separator = this.suggestURL.includes('?') ? '&' : '?';
+      const url = `${this.suggestURL}${separator}tx_solr%5BqueryString%5D=${encodeURIComponent(query)}`;
       const response = await fetch(
         url,
         { signal: this.abortController.signal }


dkd-kaehm:
@dmitryd

Is OK so?

dmitryd:
@dkd-kaehm yes, looks good to me!

dkd-kaehm (approved):

PR #4658, opened 2026-05-18
[BUGFIX] Make suggest widget form submission preventable

### LINKED ISSUE #4657: [BUG] Cannot prevent form submission when using autocomplete
### Description
In [suggest_controller.js](https://github.com/TYPO3-Solr/ext-solr/blob/main/Resources/Public/JavaScript/suggest_controller.js), the form is submitted programmatically in two scenarios:

- Pressing **Enter** when the suggestion box is closed (`enterHandler()`)
- Selecting an **autocomplete suggestion** (non-document suggestion) from the list (`selection` event callback)

Both currently call `this.form.submit()`.
Calling `form.submit()` directly does not fire the `submit` event. As a result, any external code that relies on `event.preventDefault()` to block or modify the submission (e.g. for AJAX `preventDefault()` or custom logic) cannot prevent the form from being submitted.

### Steps to Reproduce
1. Add a `submit` event listener to the search form that calls `event.preventDefault()`.
2. Type a search query into the Solr suggest input.
3. Press **Enter** while the suggestion box is closed (or select an autocomplete suggestion from the list).
4. The form submits directly - the `submit` is not prevented.

### Expected behavior
When the suggest controller submits the form, it should behave like a normal submit button click – firing a `cancelable` `submit` event that allows `preventDefault()` to work.

**Used versions:**
 - TYPO3 Version: 14.3.1
 - Browser: Chrome 148.0.7778.97
 - EXT:solr Version: 14.0.0-beta1
 - Used Apache Solr Version: 9.10.1
 - PHP Version: 8.5.5
 - MySQL Version: 10.11.16-MariaDB

### Suggested solution
Replace `this.form.submit()` with `this.form.requestSubmit()` (modern, fires submit event and respects validation) or, for broader compatibility, a small helper that dispatches a `submit` event and only calls `form.submit()` if the event was not cancelled.

MDN reference: [`HTMLFormElement.requestSubmit()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/requestSubmit)

danilovq:
# What this pr does

Replace `this.form.submit()` with `this.form.requestSubmit()` (modern, fires submit event and respects validation).

# How to test

1. Add a `submit` event listener to the search form that calls `event.preventDefault()`.
2. Type a search query into the Solr suggest input.
3. Press **Enter** while the suggestion box is closed (or select an autocomplete suggestion from the list).
4. `submit` event is fired and the page does not reload/submit.

Fixes: #4657

### DIFF
--- a/Resources/Public/JavaScript/suggest_controller.js
+++ b/Resources/Public/JavaScript/suggest_controller.js
@@ -182,7 +182,7 @@ class SuggestController {
     // If nothing is selected or suggestion box is closed then submit form
     if ((cursor === undefined || cursor === -1) || !this.autoCompleteInstance.isOpen) {
       if (this.autoCompleteInstance.input.value.length) {
-        this.form.submit()
+        this.form.requestSubmit()
       }
     } else {
       this.autoCompleteInstance.select(cursor)
@@ -333,7 +333,7 @@ class SuggestController {
                 // Get href from selected item and open it
                 window.location.href = resultItems[selectionIndex].children[0].href
               } else {
-                this.form.submit()
+                this.form.requestSubmit()
               }
             },
             navigate: (event) => {


dkd-kaehm (approved):
Thanks for fixing that issue.

PR #4672, opened 2026-05-26
[DOCS] Update Indexing.rst: clearing documents only for current site

pi-phi:
fixing example code:
when clearing documents only delete documents for matching type and site

# What this pr does

don't delete all documents for type, but only for type and matching site

# How to test

use one solr instance for multiple TYPO3 instances and index documents with same type for both instances.

all documents would be deleted but rebuild only for current site 


dkd-kaehm (approved):
Merci

PR #4677, opened 2026-06-03
[BUGFIX] Skip languages absent on a resolved site for queued page item

### LINKED ISSUE #4675: [BUG] IndexQueue: "Language X does not exist on site Y"
**Describe the bug**
Hit this after upgrading from `13.x` to `14.0.0-beta3` on a multi-site TYPO3 v14 install. 

We have a parent site (call it corporate) with languages `[0=en, 1=de, 2=fr, 3=it]` and Solr connections for all four. Underneath it there is a child site (asia) with `[0=en, 1=zh-CN, 2=zh-HK]`. Pages under the asia tree end up in the queue under corporates site config, so `Item::getSite()` returns corporate.

The queue worker loops over the connections coming from corporate (so it tries languages 1, 2, 3), but `buildServerRequest` resolves the page itself via `SiteFinder::getSiteByPageId(...)`, which returns `asia`.

The first call to `$asia->getLanguageById(1)` throws, because those IDs map to different languages on asia or do not exist there at all. The whole queue item fails.

In v13.x the page indexer issued a real HTTP request with `&L=<int>` and the frontend on the other end sorted it out; the queue side did not need a SiteLanguage object. In v14 the in-process sub-request `frontendApplication->handle($request)` needs the language object up front, so the mismatch crashes instead of being absorbed downstream.

**To Reproduce**
No custom code needed, only site config and one page.
 
1. Configure two sites:    
- Site A, root page 1, languages [0=en, 1=de, 2=fr], Solr connections for all three.
- Site B, root page 100 placed inside Site A pagetree, languages [0=en].
2. In Site A Solr configuration, make sure the queue picks up pages from Site B tree (typical default: root-line based).
3. Create or edit a page under root 100 so it lands in the index queue.
4. Run the queue worker:  vendor/bin/typo3 solr:index:queue or trigger the scheduler task.

The queue item fails with `Language 1 does not exist on site "B"` the moment the worker tries Site A `de` connection.

`IndexingService::getPageSolrConnections()` returns the connection list from `$item->getSite()->getAllSolrConnectionConfigurations()` (Site A). `buildServerRequest` later calls   `SiteFinder::getSiteByPageId($pageUid)` (Site B). 

Nothing reconciles the two language sets between those two calls. `executeRecordIndexingSubRequest` lands in the same `buildServerRequest,` so the record indexing path looks affected the same way, though I have not reproduced it explicitly with a record item.

Workaround we're running locally: A patch on IndexingService that intersects the connection list with the page-site languages before the loop runs:

`return array_intersect_key($solrConnections, $pageSite->getLanguages());`

**Used versions**
  - TYPO3 14.x
  - EXT:solr 14.0.0-beta3

amirarends:
# What this pr does

In `IndexingService::getPageSolrConnections()`, intersect the connection list with the languages defined on the page's own site fetched via `SiteFinder::getSiteByPageId()` before returning it. Languages that the site of queue item advertises but the site of the page does not provide are filtered out. When at least one language gets dropped, the new code logs an info entry with the queue item UID, the queue item's root page UID, the page's site identifier, and the dropped language IDs. If the page does not resolve to any site `SiteNotFoundException`, the filter falls back to returning the unfiltered list, preserving the existing downstream failure mode rather than masking it here. 

# How to test

Provided functional tests sets up two nested sites that share the same Solr backend:

- outer site, root page 1, languages EN/DE/DA, DE+DA with EN fallback, all Solr-enabled
- inner site, root page 2 nested at pid=1, language EN only

The fixture pre-seeds a queue item against the outer root for a page that physically lives in the inner tree. 
After running `IndexService->indexItems()`, the test asserts:

- the queue item's errors column stays empty and indexed is greater than 0
- core_en received the page
- core_de and core_da did not receive the page

Without the filter, the queue item's indexed stays 0, because `indexPageItem()` returns false after the caught DE/DA exceptions, which the test catches.

Fixes: #4675

### DIFF
--- a/Classes/IndexQueue/IndexingService.php
+++ b/Classes/IndexQueue/IndexingService.php
@@ -124,6 +124,17 @@ readonly class IndexingService
         $success = true;
         $pageUserGroup = (int)($this->buildPageParameters($item)['pageUserGroup'] ?? 0);
         foreach ($solrConnections as $languageUid => $solrConnection) {
+            if (!$this->isLanguageApplicableForPage($item, $languageUid)) {
+                $this->logger->info(
+                    'Skipping language not configured on the resolved site',
+                    [
+                        'indexQueueItem' => $item->getIndexQueueUid(),
+                        'queueItemRootPageUid' => $item->getRootPageUid(),
+                        'language' => $languageUid,
+                    ],
+                );
+                continue;
+            }
             $accessGroups = $this->findUserGroupsForPage($item, $languageUid);
             if ($pageUserGroup > 0) {
                 // A page with fe_group restriction has no publicly-accessible content variants:
@@ -144,6 +155,17 @@ readonly class IndexingService
         }
 
         return $success;
+    }
+
+    protected function isLanguageApplicableForPage(Item $item, int $language): bool
+    {
+        try {
+            $pageSite = $this->siteFinder->getSiteByPageId($this->resolvePageUid($item));
+            $pageSite->getLanguageById($language);
+            return true;
+        } catch (\InvalidArgumentException | SiteNotFoundException) {
+            return false;
+        }
     }
 
     /**


amirarends:
@dkd-kaehm I have reworked this PR. After thinking through your feedback and the alternatives, i have landed on the smallest possible change that resolves the reported failure mode: `IndexingService::indexPageItem()` now pre-checks each language against the resolved site using `Site::getLanguageById()`. When the site does not configure that UID, the iteration is skipped cleanly with an info log. No failed sub-request, no error trail, and the queue item completes across the languages that actually apply and gets its indexed timestamp.

On reflection, how nested-site pages should be routed through the queue is a EXT:solr architecture discussion, not a bugfix scope. There are legitimate setups (shared corporate pagetrees with country/market subsites being one of them) where a single conceptual answer does not fit everyone. 👍

dkd-kaehm:
@amirarends
Thanks for reworking this, and sorry for the long silence.

I still can't merge it.

The guard only helps as long as the language IDs of the two sites don't overlap.
In your own example (A `1=de`, B `1=zh-CN`) `getLanguageById(1)` succeeds on B, so the check passes,
the sub-request renders zh-CN and `SolrIndexingMiddleware` writes the result into A's German core, with the siteHash of B.
The exception is gone, but we get a wrong document in a core where neither of the two sites can find it(Data-Junk).
IMHO that is worse than the error we have now.

v13 was not better by the way. `PageUriBuilder` resolved the site of the page as well,
`PageRouter::generateUri()` calls `$site->getLanguageById()` there too,
and the FE part took the connection from `$item->getRootPageUid()`.
So there is no old behaviour to restore here.

I would rather go back to the initializer and record-monitor, as I proposed in #4675 and as you agreed there yourself:
> ... Once is_siteroot=1 on a descendant page is respected as a site boundary, the cross-site contamination disappears at the source. ...

Then every sub-site initializes its own pages with its own languages and connections.
`additionalPageIds` stays the opt-in if an outer site really wants foreign pages.
That needs a changelog entry and a note that the index queue must be re-initialized.

---

To be clear though, this is a real EPIC and a breaking change, not a quick fix.
Proper support for nested root sites pulls in much more than the initializer:
the TYPO3 language modes of both sites, `plugin.tx_solr.index.queue.[indexConfig].additionalPageIds`,
the siteHash and allowedSites resolution and so on.
There is a whole string of consequences attached to it, so it needs fine planning first.

Nothing wrong with your setup by the way, shared page trees with market sub-sites have to keep working.
If you want to kick the initializer part, go ahead and community and I will jump on.

Issue #4659, opened 2026-05-19
[BUG] Spellchecking: Correct terms are removed from query


seirerman:
**Describe the bug**
When solr detects a misspelled search term and either shows a message ("did you mean ...") or directly links to the search results based on the corrected search term, all other search terms are ignored. They should still be part of the new query.

**To Reproduce**

1. `plugin.tx_solr.search.spellchecking = 1`
2. Use a search query containing more than one word, with at least one typo
3. Solr suggests to use the correct term and offers a link (or redirects, if plugin.tx_solr.search.spellchecking.searchUsingSpellCheckerSuggestion=1)
4. Only the corrected search term is used, the other correct terms are removed from the query

**Expected behavior**
The new search results should be based on all the search terms, not just the corrected one. The user has to add the missing search terms again.

**Screenshots**

<img width="333" height="198" alt="Image" src="https://github.com/user-attachments/assets/80453151-68a2-4ca0-8ccc-7cb5be99eb12" />

The link then points to `...?q=formular`

**Used versions (please complete the following information):**
 - TYPO3 Version: 13.4.29
 - EXT:solr Version: 13.1.1
 - Used Apache Solr Version: 9.10.1

dkd-kaehm:
@seirerman 
Please review and test #4671

seirerman:
This is for the main branch (TYPO3 14/solr 14.x), right?
I'm still on TYPO3 13/solr 13.x...

PR #4682, opened 2026-06-17
[BUGFIX:BP:13.1] Spellchecking: keep correct terms when offering corrections

dkd-kaehm:
The "did you mean" link and the spellchecker auto-correction used to
drop every correctly-spelled term from a multi-word query, leaving
only the suggested replacement word — so a search for "hello formuller"
would offer "formular" instead of "hello formular".

The primary path now uses Solr's `spellcheck.collations` field, which
already contains the full corrected query. When Solr discards the
collation (e.g. because `spellcheck.maxCollationTries=1` filters out
collations that match no documents), a per-suggestion `fullQuery`
falls back to reconstructing the query from the suggestion's
`startOffset`/`endOffset` against `responseHeader.params.q`, with a
word-boundary `preg_replace` as a last-resort fallback.

Also fixes a latent bug where the misspelled term was captured from
the array index (integer 0) instead of the actual string in the
alternating flat NamedList.

`ResponseAdapter::$spellcheck` is now typed as `SpellcheckSection|null`
— a shape-only stdClass subclass — so IDEs and static analyzers can
autocomplete the well-known `suggestions` and `collations` fields.

Fixes: #4659
Ports: #4671

### DIFF
--- a/Classes/Domain/Search/ResultSet/ResultSetReconstitutionProcessor.php
+++ b/Classes/Domain/Search/ResultSet/ResultSetReconstitutionProcessor.php
@@ -127,10 +127,14 @@ class ResultSetReconstitutionProcessor implements SearchResultSetProcessor
             return $resultSet;
         }
 
+        $originalQuery = $this->resolveOriginalQuery($resultSet);
+
         $misspelledTerm = '';
-        foreach ($response->spellcheck->suggestions as $key => $suggestionData) {
+        foreach ($response->spellcheck->suggestions as $suggestionData) {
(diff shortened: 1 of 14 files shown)


dkd-kaehm:
@seirerman 
Please review and test this one, that is the 13 version, which is cherry-picked from #4671.

seirerman:
Looks good to me.

One tiny thing: The URL goes from `?q=adsf+yxcv` to `?q=asdf%20yxcv`. Still works fine, though.

seirerman (approved):
Looks good to me.

PR #4697, opened 2026-07-02
[BUGFIX] Ensure correct overlay handling for related records

### LINKED ISSUE #4696: [BUG] Indexing fails with TypeError when a SOLR_RELATION foreign record has no translation for the target language
**Describe the bug**
When indexing a translated record via `SOLR_RELATION` (`ApacheSolrForTypo3\Solr\ContentObject\Relation`) whose related foreign-table record has no translation for the current language, indexing crashes with a fatal `TypeError`.

In `Relation::resolveRelatedValue()` (`Classes/ContentObject/Relation.php`), when `$languageUid > 0`, the related record is passed through `FrontendOverlayService::getOverlay()`:

```php
if ($this->getLanguageUid($parentContentObject) > 0 && !empty($foreignTableName)) {
    $relatedRecord = $this->getFrontendOverlayService($parentContentObject)->getOverlay($foreignTableName, $relatedRecord);
}

$contentObject = clone $parentContentObject;
$contentObject->start($relatedRecord, $foreignTableName);
```

If the foreign-table record only exists in the default language (`sys_language_uid = 0`) and has no localization for the requested language, `PageRepository::getLanguageOverlay()` legitimately returns `null`. This `null` is not checked and is passed directly into `ContentObjectRenderer::start()`, whose first argument is typed as `array`, causing:

```
TypeError: TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer::start(): Argument #1 ($data) must be of type array, null given
```

**To Reproduce**
1. Have a record (e.g. `tx_news_domain_model_news`) with a relation field mapped via `SOLR_RELATION` to a foreign table, configured to resolve a label field.
2. Create a translation of the source record while the related foreign-table record exists only in the default language (`sys_language_uid = 0`, no `l10n_parent` translation available).
3. Trigger indexing of the translated record.
4. The indexing sub-request throws a `TypeError` and the record is not indexed.

**Expected behavior**
When the related foreign-table record has no translation for the current language, `resolveRelatedValue()` fall back to the default-language record instead of throwing an uncaught `TypeError`, so the record can be indexed.

**Stack trace**
```
TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer::start(): Argument #1 ($data) must be of type array, null given, called in /app/vendor/apache-solr-for-typo3/solr/Classes/ContentObject/Relation.php on line 332 and defined in /app/vendor/typo3/cms-frontend/Classes/ContentObject/ContentObjectRenderer.php:506

Stack trace (excerpt):
#0 Relation.php(332): ContentObjectRenderer->start(NULL, 'tt_address')
#1 Relation.php(289): Relation->resolveRelatedValue(NULL, Array, 'first_name', Object(ContentObjectRenderer), 'tt_address')
#2 Relation.php(133): Relation->getRelatedItemsFromForeignTable('tx_news_domain_...', 5785, Array, Object(ContentObjectRenderer))
#3 Relation.php(83): Relation->getRelatedItems(Object(ContentObjectRenderer))
#4 ContentObjectRenderer.php(693): Relation->render(Array)
...
#88 IndexQueueWorkerTask.php(48): IndexService->indexItems(50)
```

 - TYPO3 Version: 14.3.3
 - EXT:solr Version: 14.0.0-beta3
 - PHP Version: 8.3.0

BastiLu:
Prevents a fatal TypeError,
when indexing a translated record via
SOLR_RELATION (ApacheSolrForTypo3\Solr\ContentObject\Relation) whose related foreign-table record has no translation for the current language, while indexing.

# What this pr does
Fall back to original language record instead of throwing error.

# How to test
See bugreport.

Fixes: #4696

### DIFF
--- a/Classes/System/UserFunctions/FlexFormUserFunctions.php
+++ b/Classes/System/UserFunctions/FlexFormUserFunctions.php
@@ -42,7 +42,7 @@ class FlexFormUserFunctions
      */
     public function getFacetFieldsFromSchema(array &$parentInformation): void
     {
-        $pageRecord = $parentInformation['flexParentDatabaseRow'];
+        $pageRecord = $parentInformation['flexParentDatabaseRow'] ?? '';
         // @todo: Fix type hinting issue properly on whole call chain.
         $configuredFacets = $this->getConfiguredFacetsForPage($pageRecord['pid'] ?? null);
 
@@ -139,7 +139,7 @@ class FlexFormUserFunctions
      */
     public function getAvailableTemplates(array &$parentInformation): void
     {
-        $pageRecord = $parentInformation['flexParentDatabaseRow'];
+        $pageRecord = $parentInformation['flexParentDatabaseRow'] ?? '';
         if (!is_array($pageRecord) || !isset($pageRecord['pid'])) {
             $parentInformation['items'] = [];
             return;


dkd-kaehm:
@BastiLu Thanks for fixing that issue, but it changes the behavior to the pure fallback, which is not fine.
Some language modies like strict are not possible then.
I would propose to fetch that error earlier, so if the overlay not applicable for those records, the method must return null as initially implemented.

We have multiple issues on language modies and want to be able to support all the modies properly, we must avoid a new side effects like that.

BastiLu:
Reopened here: https://github.com/TYPO3-Solr/ext-solr/pull/4718

PR #4704, opened 2026-07-09
[BUGFIX] Fix leaked language Context aspect between indexing sub-requests

### LINKED ISSUE #4703: [BUG] Context 'language' aspect leaks between IndexingService sub-requests
**Describe the bug**

The language aspect of the `Context` Singleton leaks between sub-requests.
This leads to solr documents being indexed with a translated title if a custom
`PageTitleProvider` is implemented using the request attribute `frontend.page.information`.

The `PrepareTypoScriptFrontendRendering` Middleware sets the request attribute `frontend.page.information` using
`PageInformationFactory::create()` (TYPO3 Core fetches the page record *before* it sets
the language aspect that is correct for the current request):

1. `setPageAndRootline()` calls `PageRepository::getPage($id)`, which decides via
   `PageRepository::getLanguageOverlay()` whether/which language overlay to apply, based
   on the **current** `language` aspect on `Context`.
2. Only afterwards, in `settingLanguage()`, does it call
   `$this->context->setAspect('language', $languageAspect)` with the aspect that is
   correct for *this* request.

Because the `Context` singleton is not reset between sub-requests, step 1 of sub-request
still sees the `language` aspect left behind by the previous sub-request. The resulting
`PageInformation::getPageRecord()` is therefore overlaid with the *wrong* language for
that brief window, before step 2 corrects the aspect for the rest of that sub-request.

**We only noticed this because of a project-specific implementation**: our project has a custom `PageTitleProviderInterface`
implementation that reads
`$this->request->getAttribute('frontend.page.information')->getPageRecord()['title']`
directly to build a custom title. That attribute holds exactly the
`PageInformation` object populated in step 1 above, so it carries the leaked-language
title.
As EXT:solr itself does not read
`getPageRecord()` the same way at the same point in the lifecycle, installations without a similar custom title provider likely won't observe wrong titles
in Solr documents.

**To Reproduce**

1. Set up a multi-language site indexed with EXT:solr, so that the index queue contains at least two
   translated pages.
2. Add a page title provider that reads
   `$request->getAttribute('frontend.page.information')->getPageRecord()['title']`
   directly.
3. Run the index queue worker so both items are processed sequentially within the same
   PHP process.
4. Inspect the resulting Solr documents/page title field: the second page in the queue is indexed with the translated title.

**Expected behavior**

Each indexed page document (and anything reading `frontend.page.information` during an
indexing sub-request) reflects the page title in the language it was indexed for,
regardless of which other items were processed earlier in the same worker run.

 - TYPO3 Version: 14.3.3
 - EXT:solr Version: 14.0.0-beta3
 - PHP Version: 8.3.31

**Additional Context**
Example of custom PageTitleProvider implementation:

```
class CustomPageTitleProvider extends AbstractPageTitleProvider
{
    public function getTitle(): string
    {
        /** @var PageInformation $pageInformation */
        $pageInformation = $this->request->getAttribute('frontend.page.information');
        return $pageInformation->getPageRecord()['title'];
    }
}
```

BastiLu:
# What this pr does

Snapshot `Context` singleton (language Aspect) state that the frontend sub-request clobbers.
See Issue for details

# How to test

See Issue

Fixes: #4703

### DIFF
--- a/Classes/System/UserFunctions/FlexFormUserFunctions.php
+++ b/Classes/System/UserFunctions/FlexFormUserFunctions.php
@@ -42,7 +42,7 @@ class FlexFormUserFunctions
      */
     public function getFacetFieldsFromSchema(array &$parentInformation): void
     {
-        $pageRecord = $parentInformation['flexParentDatabaseRow'];
+        $pageRecord = $parentInformation['flexParentDatabaseRow'] ?? '';
         // @todo: Fix type hinting issue properly on whole call chain.
         $configuredFacets = $this->getConfiguredFacetsForPage($pageRecord['pid'] ?? null);
 
@@ -139,7 +139,7 @@ class FlexFormUserFunctions
      */
     public function getAvailableTemplates(array &$parentInformation): void
     {
-        $pageRecord = $parentInformation['flexParentDatabaseRow'];
+        $pageRecord = $parentInformation['flexParentDatabaseRow'] ?? '';
         if (!is_array($pageRecord) || !isset($pageRecord['pid'])) {
             $parentInformation['items'] = [];
             return;


dkd-kaehm:
@BastiLu please check the introduced fails in tests.

dkd-kaehm:
@BastiLu Seems the commit from #4697 is included in this pr.

BastiLu:
@dkd-kaehm Sorry for this mess.. reopened here: https://github.com/TYPO3-Solr/ext-solr/pull/4717

PR #4722, opened 2026-07-20
[BUGFIX] Disable page cache before page indexing sub-requests

### LINKED ISSUE #4721: [BUG] Disable page cache before page indexing sub-request
**Describe the bug**
Page indexing sub-requests fails with HTTP 500 for **every page that has a valid page cache entry** for the requested language. 

The queue item errors out with only `Sub-request returned error status` / `statusCode: 500`; the underlying exception is: ``` RuntimeException: Setup array has not been initialized. This happens in cached Frontend scope where full TypoScript is not needed by the system. in typo3/cms-core/Classes/TypoScript/FrontendTypoScript.php:182 ``` thrown from `Util::getSolrConfiguration()` via `Typo3PageContentExtractor::excludeContentByClass()` while building the page document in `SolrIndexingMiddleware::handlePageIndexing()`. 

Because it depends on the page cache, the failure pattern is confusing in practice: pages recently visited by frontend users (warm cache) fail to index, uncached pages index fine. On production sites this means the **most popular pages are the most likely to be missing or stale in the index**, while DEV/staging environments (cold caches) work flawlessly.

`SolrIndexingMiddleware` disables the page cache via `CacheInstruction`, but the middleware is registered `after: typo3/cms-frontend/prepare-tsfe-rendering`. TYPO3 core reads the `frontend.cache.instruction` attribute at the start of `PrepareTypoScriptFrontendRendering`. 

So on a page cache hit:

1. Core serves the cached page content and sets `$needsFullSetup = false` → `FrontendTypoScript` contains only settings + config, no setup array ("cached Frontend scope").
2. `SolrIndexingMiddleware` disables the cache, but it is too late; the decision is already made.
3. `Typo3PageContentExtractor` needs `plugin.tx_solr.*` from the full setup → `getSetupArray()` throws.

**To Reproduce**
1. TYPO3 v13/14 + EXT:solr 14, page indexing configured.
2. Open any indexable page in the frontend as an anonymous visitor, so a page cache entry is written.
3. Run the index queue worker for that page.
4. The item fails with `Sub-request returned error status`, statusCode 500; the exception above is logged by the middleware. Flush the pages cache and re-run: the same item indexes successfully.

amirarends:
Resolves #4721

# What this pr does

Fixes `indexPage` sub-requests failing with a 500 (`RuntimeException: Setup array has not been initialized`) for every page that has a valid page cache entry.

`IndexingService::executeSubRequest()` now sets a disabled `CacheInstruction` as request attribute before dispatching the sub-request into the frontend application, so `PrepareTypoScriptFrontendRendering` bypasses the page cache and always computes the full TypoScript setup.

# How to test

1. Open an indexable page in the frontend as anonymous visitor (writes a page cache entry).
2. Run the index queue worker for that page → fails with `Sub-request returned error status` before this patch, succeeds after.
3. Setting the instruction early also keeps the existing behaviour of never *storing* indexing renders in the page cache (same `CacheInstruction` semantics as before, just applied in time).

Fixes: #4721

### DIFF
--- a/Classes/IndexQueue/IndexingService.php
+++ b/Classes/IndexQueue/IndexingService.php
@@ -46,6 +46,7 @@ use TYPO3\CMS\Core\PageTitle\PageTitleProviderManager;
 use TYPO3\CMS\Core\Routing\InvalidRouteArgumentsException;
 use TYPO3\CMS\Core\Site\SiteFinder;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Frontend\Cache\CacheInstruction;
 use TYPO3\CMS\Frontend\Http\Application as FrontendApplication;
 
 /**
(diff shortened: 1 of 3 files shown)


dkd-kaehm (approved):
@amirarends 
Thanks for fixing that issue.
I added a migration test to cover that.

That is really wanted for EXT:solrfal 14 ;)

PR #4733, opened 2026-08-14
[BUGFIX] Restore Context aspects after indexing sub-request

hdj-typoconsult:
Frontend\Http\Application::initializeContext() overwrites aspects on the shared Context singleton during each index queue sub-request. Without restoring them, a backend-triggered indexing run leaves a logged-out "backend.user" aspect, so BackendUserAuthenticator expires the be_typo_user cookie and logs the editor out of the backend.

Save the clobbered aspects (date, visibility, workspace, backend.user, frontend.user) before the sub-request and restore them in the finally block, preserving the original "unset when previously absent" semantics.

# What this pr does

Please add a description here

# How to test


dkd-kaehm (approved):
@hdj-typoconsult 
Thanks for reporting and fixing that issue.
I squashed both commits, because both must be in single commit.

PR #4749, opened 2026-08-26
[BUGFIX] Restore teaser for results without a match in the highlighted field

### LINKED ISSUE #4748: [BUG] Enforced unified highlighter (13.1.4) silently breaks result teasers: `hl.alternateField` is ignored and the ViewHelper falls back to the full, untruncated content
## Describe the bug

Since 13.1.4 `Highlighting::build()` unconditionally enforces the unified highlighter (`Classes/Domain/Search/Query/ParameterBuilder/Highlighting.php`, lines 142–145):

```php
$highlighting->setMethod(SolariumHighlightingInterface::METHOD_UNIFIED);
$highlighting->setOffsetSource(SolariumHighlightingInterface::OFFSETSOURCE_ANALYSIS);
$highlighting->setBoundaryScannerType('WORD');
$highlighting->setFragsizeIsMinimum(false);
```

Up to 13.1.3 the FastVector highlighter was used whenever `fragmentSize >= 18` (13.1.3, lines 132–141). I understand the enforcement is part of the SST-235567 security fix.

However, the unified highlighter **does not support `hl.alternateField` / `hl.maxAlternateFieldLength`** — those parameters only exist for the original and FastVector highlighters. The configset shipped by this very release still relies on them (`Resources/Private/Solr/configsets/ext_solr_13_1_0/conf/solrconfig.xml`, `/select` handler):

```xml
<str name="hl.method">unified</str>                              <!-- line 149 -->
...
<str name="f.content.hl.alternateField">content</str>            <!-- line 154, dead -->
<str name="f.content.hl.maxAlternateFieldLength">200</str>       <!-- line 155, dead -->
```

Consequence: for every result where the search term matches only `title`, `keywords` or heading fields (i.e. not the stored `content`), Solr now returns **no highlighting entry at all** (previously: the first 200 characters via `alternateField`). `HighlightResultViewHelper::getHighlightedContent()` then falls back to the **entire untruncated `content` field** (`Classes/ViewHelpers/Document/HighlightResultViewHelper.php`, line 45) — result lists render teasers of 10,000+ characters.

## To Reproduce

Against a core built from the `ext_solr_13_1_0` configset (Solr 9.10.1), pick a document whose title contains a word that does not occur in its `content`, then run the same query with all three highlighter methods:

```
q=title:Impressum & defType=lucene & hl=true & hl.fl=content
   & hl.requireFieldMatch=true & hl.fragsize=155 & hl.method=...
```

| `hl.method`             | highlighting response for `content`        |
|-------------------------|--------------------------------------------|
| `original`              | 200 chars (alternateField applies)         |
| `fastVector` (≤ 13.1.3) | 200 chars (alternateField applies)         |
| `unified` (13.1.4)      | **empty** → ViewHelper renders full field  |

In the frontend: search for a term that only matches page titles → the results list shows the complete page content instead of a ~200 character teaser. Rolling back to 13.1.3 (no reindex needed) restores the cropped teasers immediately.

## Expected behavior

Results that match outside the `content` field keep showing a short leading-text teaser, as they did up to 13.1.3.

## Suggested fix

The unified highlighter has its own equivalent: `hl.defaultSummary`. Solarium 6.4 already supports it (`HighlightingTrait::setDefaultSummary()`, request builder maps it to `hl.defaultSummary`).

1. `Highlighting::build()`: since the method is now enforced centrally in PHP, the replacement behavior should be enforced there too:

   ```php
   $highlighting->setDefaultSummary(true);
   ```

2. Configset `ext_solr_13_1_0` (`solrconfig.xml`, `/select` defaults): replace the dead lines 154–155 with

   ```xml
   <str name="f.content.hl.defaultSummary">true</str>
   ```

3. Extend `Docker/SolrServer/docker-entrypoint-initdb.d-as-sudo/fix-SST-235567-*-highlighter-defaults.sh`
   so existing cores migrated to the unified defaults get the same replacement.

Note: `defaultSummary` returns leading passages in `fragsize` chunks (with the default `hl.snippets=3` roughly 3 × 155 chars), not a hard 200-character cap — close to, but not identical with, the old `maxAlternateFieldLength` behavior. As a belt-and-suspenders measure the full-content fallback in `HighlightResultViewHelper` (line 45) could additionally be truncated to `fragmentSize` instead of returning the whole field.

## Used versions

 - TYPO3 Version: 13.4.34
 - EXT:solr Version: 13.1.4 (regression introduced here; 13.1.2/13.1.3 not affected)
 - Used Apache Solr Version: 9.10.1
 - solarium/solarium 6.4.1
 - PHP Version: 8.3.33

peratoner-louis:
## Problem

Since 13.1.4 `Highlighting::build()` enforces the Unified Highlighter unconditionally (part of the SST-235567 security fix). The Unified Highlighter does not support `hl.alternateField` / `hl.maxAlternateFieldLength` — those options only exist for the original and FastVector highlighters — yet the shipped `ext_solr_13_1_0` configset still relies on them in the `/select` handler defaults.

As a result, documents matching only in `title`, `keywords` or heading fields get no highlighting entry from Solr anymore (previously: the first 200 characters via `alternateField`), and `HighlightResultViewHelper` falls back to rendering the entire untruncated `content` field — result lists show teasers of 10,000+ characters.

Verified on Solr 9.10.1 (same core, same title-only-match query): `hl.method=original` → 200 chars, `fastVector` → 200 chars, `unified` → empty.

## What was changed

* `Highlighting::build()` now requests `hl.defaultSummary=true`, the Unified Highlighter's equivalent: it returns the leading portion of the field when no passages match. Solarium 6.4 supports this natively (`setDefaultSummary()`). Set globally in PHP, mirroring how the highlighter method itself is enforced; per-field server overrides (`f.<field>.hl.defaultSummary=false`) remain possible.
* The `ext_solr_13_1_0` configset replaces the dead `f.content.hl.alternateField` / `f.content.hl.maxAlternateFieldLength` defaults with `f.content.hl.defaultSummary=true`.
* The `fix-SST-235567-*-highlighter-defaults.sh` migration script gets a second, independently guarded, idempotent block that applies the same replacement to existing cores — including cores already migrated by an earlier run of the script. All checks and edits are scoped to the `/select` request handler, so customized handlers (e.g. a FastVector handler with its own `alternateField`) are left untouched. It writes its own backup file: `solrconfig.xml.Backup-unified-defaultSummary`.
* `TxSolrSearch.rst` is updated in a separate `[DOCS]` commit: the "FastVector if `fragmentSize` >= 18" rule and the term-vector requirement no longer apply since 13.1.4; the reference now describes the unconditional Unified Highlighter, the default-summary fallback, and links the versioned Solr 9.10 highlighting guide.

## Behavioral note

`hl.defaultSummary` returns leading passages of roughly `hl.snippets × fragmentSize` characters — close to, but not identical with, the former hard 200-character `maxAlternateFieldLength` cap. The Unified Highlighter offers no separate length parameter for the default summary; the only knob (`hl.snippets`) would also reduce the number of fragments for real matches, so the approximate length is accepted.

## Verification

* New unit test `defaultSummaryIsRequestedWhenHighlightingIsEnabled` proves the parameter is emitted; full unit suite green (1019 tests), php-cs-fixer and PHPStan clean.
* Migration script dry-run matrix (all states re-run twice; second run is a byte-identical no-op, XML stays valid):

| Initial state | Result |
|---|---|
| Pre-13.1.4 core (`hl.method=original` + `alternateField`) | `/select` migrated to unified + `defaultSummary` |
| Core already migrated by the previous script version (unified + dead `alternateField`) | `/select` gets `defaultSummary`, dead options removed |
| Custom handler already containing `f.content.hl.defaultSummary` | `/select` still migrated, custom handler untouched |
| Custom FastVector handler with its own `alternateField` + `maxAlternateFieldLength` | Custom handler fully preserved, `/select` migrated |

Fixes #4748

### DIFF
--- a/Classes/Domain/Search/Query/ParameterBuilder/Highlighting.php
+++ b/Classes/Domain/Search/Query/ParameterBuilder/Highlighting.php
@@ -143,6 +143,7 @@ class Highlighting extends AbstractDeactivatable implements ParameterBuilderInte
         $highlighting->setOffsetSource(SolariumHighlightingInterface::OFFSETSOURCE_ANALYSIS);
         $highlighting->setBoundaryScannerType('WORD');
         $highlighting->setFragsizeIsMinimum(false);
+        $highlighting->setDefaultSummary(true);
 
         if ($this->getPrefix() !== '' && $this->getPostfix() !== '') {
             $highlighting->setSimplePrefix($this->getPrefix());
--- a/Docker/SolrServer/docker-entrypoint-initdb.d-as-sudo/fix-SST-235567-2026050810000025-highlighter-defaults.sh
+++ b/Docker/SolrServer/docker-entrypoint-initdb.d-as-sudo/fix-SST-235567-2026050810000025-highlighter-defaults.sh
@@ -44,3 +44,21 @@ if ! grep -q "${PATCH_MARKER}" "${PATH_AND_FILENAME_SOLRCONFIG}" \
 			<str name=\"hl.bs.type\">WORD</str>\\
 			<str name=\"hl.fragsizeIsMinimum\">false</str>" "${PATH_AND_FILENAME_SOLRCONFIG}"
 fi
+
+# The Unified Highlighter does not support hl.alternateField: replace the
+# leading-text fallback with its hl.defaultSummary equivalent.
+# Runs independently of the block above so that cores migrated by an earlier
+# version of this script are aligned as well. All checks and edits are scoped
+# to the /select request handler to leave customized handlers untouched.
+PATH_AND_FILENAME_BACKUP_SUMMARY="${PATH_SOLRCONFIG}/solrconfig.xml.Backup-unified-defaultSummary"
+SELECT_HANDLER_RANGE='/<requestHandler name="\/select"/,/<\/requestHandler>/'
+
+if sed -n "${SELECT_HANDLER_RANGE}p" "${PATH_AND_FILENAME_SOLRCONFIG}" | grep -q '<str name="f.content.hl.alternateField">content</str>' \
+	&& ! sed -n "${SELECT_HANDLER_RANGE}p" "${PATH_AND_FILENAME_SOLRCONFIG}" | grep -q 'f.content.hl.defaultSummary'; then
+	echo "  replacing f.content.hl.alternateField by f.content.hl.defaultSummary in the /select handler (alternateField is ignored by the Unified Highlighter)"
+
+	cp "${PATH_AND_FILENAME_SOLRCONFIG}" "${PATH_AND_FILENAME_BACKUP_SUMMARY}"
+
+	sed -i "${SELECT_HANDLER_RANGE} s|<str name=\"f.content.hl.alternateField\">content</str>|<str name=\"f.content.hl.defaultSummary\">true</str>|" "${PATH_AND_FILENAME_SOLRCONFIG}"
+	sed -i "${SELECT_HANDLER_RANGE} {/<str name=\"f.content.hl.maxAlternateFieldLength\">/d}" "${PATH_AND_FILENAME_SOLRCONFIG}"
+fi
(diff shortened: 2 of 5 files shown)


dkd-kaehm (approved):
Thanks.

PR #4753, opened 2026-08-28
[DOCS] Expand multi-value cObjs section with an example

thomasrawiel:
Adding an example for the multi-value cObjs section as discussed in slack.

Let me know if you need the full code-example or migration path


dkd-kaehm (approved):
Thanks @thomasrawiel

PR #4759, opened 2026-09-01
[!!!][BUGFIX] Report why an Index Queue item failed, and stop queueing foreign sites' pages

### LINKED ISSUE #4707: [BUG] Index queue items are never marked as failed when indexing returns false, queue gets permanently stuck
**Describe the bug**
In `IndexService::indexItems()`, when `IndexingService::indexItems()` returns `false` (for example because the indexing sub-request returned an error status, or no Solr connection could be resolved), the queue item is neither marked as indexed (`updateIndexTimeByItem()` is only called on success) nor marked as failed (`markItemAsFailed()` is only reached in the `catch` block, but no exception is thrown on the `false` path).

The item therefore keeps `changed > indexed` and `errors = ''`, stays at the top of the queue ordering, and is picked again on every worker run. Once the number of such silently failing items reaches the worker's `documentsToIndexLimit`, every run fetches the same batch, all of them fail again, and the whole queue makes **zero progress**. The Index Queue module shows no errors on any item, so there is nothing pointing at the cause.

There are at least three code paths in `IndexingService` that return `false` without throwing:
- sub-request returns status >= 400 (`executeSubRequest()` logs "Sub-request returned error status" and returns `null`)
- sub-request returns a non-JSON body, e.g. a 3xx redirect response (nothing is logged at all)
- `getPageSolrConnections()` / `getRecordSolrConnections()` return an empty array (nothing is logged at all)

We hit this in production: a single editor-created `sys_redirect` on a language base URL made every record indexing sub-request fail silently.

**To Reproduce**

1. Set up any site where an indexing sub-request fails, e.g. a page that throws an exception during uncached rendering (returns 500), or a `sys_redirect` on the sub-request URL (returns 3xx)
2. Fill the index queue and run the Index Queue Worker task repeatedly
3. Observe: the affected items keep `errors = ''` and `indexed = 0`, are re-fetched every run, and once enough items are affected the queue progress stays at 0% forever

**Expected behavior**
Items whose indexing returns `false` should be marked as failed so they are skipped on subsequent runs and become visible as errors in the Index Queue module, e.g.:

```php
if ($itemIndexed) {
    $this->indexQueue->updateIndexTimeByItem($itemToIndex);
    // ...
} else {
    $errors++;
    $this->indexQueue->markItemAsFailed($itemToIndex, 'Indexing returned false — see TYPO3 log for details');
}
```

dkd-kaehm:
Next step of the dead-code EPIC, and this one is release-relevant rather than cleanup: it fixes three reported issues found while migrating `IndexerTest`. The branch is based on #4758, but targets `main` so the workflows actually run, so until #4758 merges this diff also contains the earlier slices. **This slice is the last four commits**, `bfa8e217e` .. `d39cc72ad`.

Fixes: #4707
Relates: #4675
Replaces: #4677

## The three defects, and how they compound

**#4707 — the queue stops silently.** `IndexingService::indexItems()` can return `false` without raising an exception. `IndexService` stamped the index time only on success and marked an item as failed only from its `catch`, so such an item kept `changed > indexed` and `errors = ''`. It stayed at the top of the queue ordering and was refetched every run. Once as many items failed that way as `documentsToIndexLimit` allows, every run picked the same batch, all failed again, and the queue made **no progress at all** — while the Index Queue module showed no error on any item.

**The reason was thrown away.** `executeSubRequest()` caught every `Throwable`, wrote one log entry and returned `null`, which the caller read as "not indexed". So even once #4707 makes the failure visible, the message was "Indexing returned false without raising an exception". The throwable is handed on now, wrapped with the action, the item, its type and record uid and the language. `buildServerRequest()` runs inside the same `try`, so the two exceptions it raises — a site without a fully qualified base URL, and a language the page's site does not define — were swallowed identically and now arrive too.

A response with status >= 400 keeps returning `null`: that is an answer, not a failure, and `findUserGroups` legitimately gets one for a page that denies access.

**#4675 — an outer site queues a nested site's pages.** Initialization collected the whole subtree below a site root, so an outer site's queue also held the pages, and the records stored on them, of a site nested in its page tree, carrying the outer site's root in `tx_solr_indexqueue_item.root`. Indexing then took the Solr connections from the outer site while the page resolved to the inner one. TYPO3 assigns language ids per site and will not reconcile them — forge [#95688](https://forge.typo3.org/issues/95688) was closed with "the only real solution is to define languages per page tree root". The item either failed with "Language X does not exist on site Y" or produced a document carrying one site's `siteHash` in another site's core, where neither site can find it.

Together these explain why #4675 was reported as an exception by one user and as a stuck queue by another: the same contamination, surfacing through two different failure paths, neither of which recorded anything on the item.

## Why the language filter of #4677 is not the fix

#4677 intersected the connection list with the languages of the page's own site. That helps only while the two sites do not use the same id for different languages. With A `1=de` and B `1=zh-CN`, `getLanguageById(1)` succeeds on B, nothing is filtered, the sub-request renders zh-CN and the result is written into A's German core with B's `siteHash` — a wrong document in a core where neither site can find it, which is worse than the error it removes.

A locale comparison would avoid that trap, but with the boundary in place there is nothing left for it to guard: newly created items cannot carry a foreign root, stale ones are removed by the re-initialization the release note asks for, and the remaining paths are explicit opt-ins whose precondition core itself documents. So the cause is removed at the source instead: a page carrying `is_siteroot` starts a site of its own and its subtree belongs to that site's queue alone.

`Site::getPagesWithinSite()` and `PagesRepository::findAllSubPageIdsByRootPageWithinSite()` are new. `Site::getPages()` and `PagesRepository::getTreeList()` are untouched, so mount point resolution, garbage collection and EXT:solrconsole's verification keep their behaviour. `additionalPageIds` still adds foreign pages explicitly, and searching across a shared page tree stays the job of `allowedSites`.

`Site::getLanguageById()` throws a bare `\InvalidArgumentException` reading "Language 1 does not exist on site B", which says nothing about why EXT:solr asked B for a language it never configured. That message now names the item's root page, the site owning the page and the language id, and points at the setting whose documentation covers the constraint.

## Breaking, and what integrators have to do

Two reasons, both in the release note:

* Queue initialization no longer collects a nested site's pages. A site that relied on that loses those documents from its cores; the nested site indexes them into its own. `allowedSites` is the supported way to search across a shared page tree.
* `tx_solr_indexqueue_item.errors` becomes a `mediumtext`, which **needs a database update before the fix takes effect**. `Throwable::__toString()` collects the traces of the whole chain and starts with the innermost exception, so a nested sub-request produced a string that exceeded the previous column — and marking an item as failed then died with "Data too long", aborting the whole request. The messages of the chain are recorded now, the traces stay in the log.

So: *Analyze Database Structure*, empty `tx_solr_indexqueue_item`, re-initialize. One instruction covers both.

## Verified against a real installation, not only the harness

The delegation was written after tracing a production failure with the Xdebug MCP: a queue item reported only "Indexing returned false", and the swallowed cause turned out to be `TYPO3Fluid\Fluid\Core\ViewHelper\InvalidArgumentValueException: No Content Object definition found at TypoScript object path "lib.dynamicContent"`, thrown three levels deep in `CObjectViewHelper`. That is a TYPO3 issue, fixed in core by [`6ceccf85`](https://github.com/TYPO3/typo3/commit/6ceccf85b31389c193dfd7feb97b3ff8b39ba1ce) and part of 14.3.7, and the release note records it — this PR is what made it identifiable at all.

Both regression tests were checked in the only way that counts, by reverting the fix:

| test | without the fix | with it |
|---|---|---|
| `itemIsMarkedAsFailedWhenIndexingReturnsFalse` | item keeps `errors = ''` | marked failed |
| `theCauseOfAFailedSubRequestIsRecordedOnTheItem` | `errors` reads "Indexing returned false without raising an exception" | `errors` names the cause |

| check | result |
|---|---|
| PHPStan level 6 | no errors |
| coding standards | 0 of 647 files |
| `IndexServiceTest` | 9 tests, 49 assertions |
| initializer / site / queue | 39 tests, 236 assertions |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Classes/Domain/Index/IndexService.php
+++ b/Classes/Domain/Index/IndexService.php
@@ -124,6 +124,14 @@ class IndexService
                         if ($itemChangedDateAfterIndex > $itemToIndex->getChanged() && $itemChangedDateAfterIndex > time()) {
                             $this->indexQueue->setForcedChangeTimeByItem($itemToIndex, $itemChangedDateAfterIndex);
                         }
+                    } else {
+                        // Without this the item keeps changed > indexed and errors = '', so it is
+                        // fetched again on every run and the queue stops making progress silently.
+                        $errors++;
+                        $this->indexQueue->markItemAsFailed(
+                            $itemToIndex,
+                            'Indexing returned false without raising an exception, see the TYPO3 log for the reason.',
+                        );
                     }
 
                     $afterIndexItemEvent = new AfterItemHasBeenIndexedEvent($itemToIndex, $this->getContextTask(), $indexRunId);
--- a/Classes/Domain/Index/IndexService.php
+++ b/Classes/Domain/Index/IndexService.php
@@ -138,7 +138,7 @@ class IndexService
                     $this->eventDispatcher->dispatch($afterIndexItemEvent);
                 } catch (Throwable $e) {
                     $errors++;
-                    $this->indexQueue->markItemAsFailed($itemToIndex, $e->getCode() . ': ' . $e->__toString());
+                    $this->indexQueue->markItemAsFailed($itemToIndex, $this->buildItemErrorMessage($e));
                     $this->generateIndexingErrorLog($itemToIndex, $e);
                 }
             }
@@ -192,6 +192,28 @@ class IndexService
     /**
      * Generates a message in the error log when an error occurred.
      */
+    /**
+     * The message the Index Queue module shows on a failed item: the messages of the whole
+     * exception chain, without the stack traces.
+     *
+     * `Throwable::__toString()` puts the traces of every exception in the chain into one string
+     * and starts with the innermost one, so the sentence naming the item and the action ends up
+     * last. An indexing sub-request nests the same throwable several times, which made that
+     * string exceed what the column can hold. The traces stay in the log, see
+     * generateIndexingErrorLog().
+     */
+    protected function buildItemErrorMessage(Throwable $e): string
+    {
+        $messages = [];
+        $cause = $e;
+        while ($cause !== null) {
+            $messages[] = $cause::class . ' (' . $cause->getCode() . '): ' . $cause->getMessage();
+            $cause = $cause->getPrevious();
+        }
+
+        return implode("\n\nCaused by: ", $messages);
+    }
+
     protected function generateIndexingErrorLog(Item $itemToIndex, Throwable $e): void
     {
         $message = 'Failed indexing Index Queue item ' . $itemToIndex->getIndexQueueUid();
--- a/Classes/Domain/Site/Site.php
+++ b/Classes/Domain/Site/Site.php
@@ -284,6 +284,37 @@ class Site
         return $this->pagesRepository->findAllSubPageIdsByRootPage($pageId, $initialPagesAdditionalWhereClause);
     }
 
+    /**
+     * Same as getPages(), except that the pages of sites nested into this one are left out.
+     *
+     * Used when initializing the Index Queue: a nested site's pages belong to that site's queue.
+     *
+     * @return int[]
+     *
+     * @throws DBALException
+     */
+    public function getPagesWithinSite(
+        ?string $indexQueueConfigurationName = null,
+        ?string $additionalWhereClause = null,
+    ): array {
+        $initialPagesAdditionalWhereClause = '';
+        if ($indexQueueConfigurationName !== null) {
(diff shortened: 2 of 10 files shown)


amirarends:
@dkd-kaehm very appreciated, thank you for taking care of our bug reports!

PR #4766, opened 2026-09-02
[FEATURE] Allow modifying page indexing subrequests through a PSR-14 event

### LINKED ISSUE #4765: [FEATURE] Add a PSR-14 event to modify the indexing subrequest before it is dispatched
As EXT:solr v14 builds page indexing requests as internal subrequests (`ServerRequestInterface`) instead of real HTTP calls, there is no way for third-party extensions to inspect or modify that request before it is sent through the frontend middleware stack. This especially breaks frontend requests for a site that lives behind authentication.

Add a PSR-14 event, e.g. `AfterFrontendSubrequestHasBeenGeneratedEvent`, dispatched in `IndexingService::executeSubRequest()` right after the subrequest is built via `buildServerRequest()`. The event should expose the current `IndexQueue\Item` and the built `ServerRequestInterface`, and allow listeners to return a modified request that is then passed to TYPO3.

**Describe alternatives you've considered**
Patching `IndexingService` via a composer patch, which is fragile across EXT:solr releases and not upgrade-safe.

**Additional context**
Relevant since the switch to subrequest-based indexing in EXT:solr 14 (TYPO3 14 LTS compatibility epic). This feature request might be related to https://github.com/TYPO3-Solr/ext-solr/issues/2696

**Target versions**
\>=14

tehplague:
# What this pr does

Adds a new PSR-14 event, `ApacheSolrForTypo3\Solr\Event\Indexing\AfterFrontendSubrequestHasBeenGeneratedEvent`, dispatched in `IndexingService::executeSubRequest()` immediately after the internal indexing subrequest is built via `buildServerRequest()`.

The event exposes:

`getItem()`: Item — the index queue item currently being processed
`getRequest()`: ServerRequestInterface — the generated subrequest
`setRequest(ServerRequestInterface $request)` — lets a listener replace the request with a modified one

The (possibly modified) request returned by the event is then used for the rest of the indexing subrequest.

# Why is this needed?

Since EXT:solr builds page indexing requests as internal `ServerRequestInterface` subrequests instead of real HTTP calls, there was no extension point to inspect or modify that request before it runs through the frontend middleware chain — e.g. to add custom headers, query parameters, or request attributes that other middlewares/extensions rely on during indexing. This event restores that customization point in a way that's compatible with the subrequest-based indexing pipeline.

# How to test

```php
use ApacheSolrForTypo3\Solr\Event\Indexing\AfterFrontendSubrequestHasBeenGeneratedEvent;
use TYPO3\CMS\Core\Attribute\AsEventListener;

#[AsEventListener]
final class ModifyIndexingSubrequest
{
    public function __invoke(AfterFrontendSubrequestHasBeenGeneratedEvent $event): void
    {
        $request = $event->getRequest()->withHeader('Authorization', 'Basic ' . base64_encode('foouser:barpassword'));
        $event->setRequest($request);
    }
}

```

Fixes: #4765

### DIFF
--- a/Classes/Event/Indexing/AfterFrontendSubrequestHasBeenGeneratedEvent.php
+++ b/Classes/Event/Indexing/AfterFrontendSubrequestHasBeenGeneratedEvent.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the TYPO3 CMS project.
+ *
+ * It is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License, either version 2
+ * of the License, or any later version.
+ *
+ * For the full copyright and license information, please read the
+ * LICENSE.txt file that was distributed with this source code.
+ *
+ * The TYPO3 project - inspiring people to share!
+ */
+
+namespace ApacheSolrForTypo3\Solr\Event\Indexing;
+
+use ApacheSolrForTypo3\Solr\IndexQueue\Item;
+use Psr\Http\Message\ServerRequestInterface;
+
+class AfterFrontendSubrequestHasBeenGeneratedEvent
+{
+    public function __construct(
+        protected readonly Item $item,
+        protected ServerRequestInterface $request
+    ) {
+    }
+
+    public function getItem(): Item
+    {
+        return $this->item;
+    }
+
+    public function getRequest(): ServerRequestInterface
+    {
+        return $this->request;
+    }
+
+    public function setRequest(ServerRequestInterface $request): self
+    {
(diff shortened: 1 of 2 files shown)


dkd-kaehm:
Please do not use the merge functionality of git within EXT:solr pull-request related branches, but rebase instead.
EXT:solr GitHub repo requires linear history...

----

Did you tried to solve that problem with own Middleware, as described in Slack-Thread?

dkd-kaehm:
@tehplague 

Did you tried to solve that problem with own Middleware, as described in Slack-Thread?
So most probably is that event not requiered at all?

PR #4779, opened 2026-09-15
[TASK] Derive the release label from the extension key

dkd-kaehm:
## What

`Build/Helpers/PREPARE_RELEASE.sh` had the `EXT:solr Version` line of the bug report template hardcoded:

```bash
sed -i "s/\(EXT:solr Version: \[e\.g\. \)[^]]*\]/\1${version}]/" "${BUG_REPORT}"
```

The label is now derived from `extra.typo3/cms.extension-key`, which the sibling `GET_TER_BUILD_INPUTS.sh` already reports as `key=`.

## Why

`da7a00e0a` designed these helpers for reuse — `BUILD_TER_ZIP.sh` resolves its helper via `BASH_SOURCE` and `cd`s to the calling repository precisely so that *"add-ons run this from their own repository"*. `PREPARE_RELEASE.sh` is the one helper that could not be reused as-is, because an add-on's bug report template lists **several** extension versions:

```
 - EXT:solr Version: [e.g. 14.0.0]
 - EXT:solrfal Version: [e.g. 14.0.0]
 - EXT:solrconsole Version: [e.g. 14.0.0]
```

Run verbatim from EXT:solrconsole, the old `sed` stamped the **`EXT:solr`** line with solrconsole's version and left **`EXT:solrconsole`** stale — wrong in both directions. Without this, every add-on needs its own copy of the script differing in one anchor.

## Verification

Run against the real templates with `99.9.9`:

**EXT:solr** (`key=solr`) — unchanged from current behaviour, only its own line and `release=` move:

```
 - TYPO3 Version: [e.g. 14.3.0]
 - EXT:solr Version: [e.g. 99.9.9]      <- moved
 - Used Apache Solr Version: [e.g. 10.0.1]
release="99.9.9"   version="14.0"       <- version= untouched
```

**EXT:solrconsole** (`key=solrconsole`) — its own line moves, the sibling lines stay put:

```
 - EXT:solr Version: [e.g. 14.0.0]
 - EXT:solrfal Version: [e.g. 14.0.0]
 - EXT:solrconsole Version: [e.g. 99.9.9]   <- moved
```

`bash -n` clean.

## Note

Only `14.0.1` carries `da7a00e0a`; `14.0.0` is behind it. Add-ons reusing these helpers therefore need `apache-solr-for-typo3/solr: ^14.0.1`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dkd-kaehm:
## Pushed a hardening follow-up — please review before merging

While checking whether this could break the other add-ons, I found two defects, one of them introduced by the first version of this PR. Both are fixed in the amended commit.

**1. The helper's failure was swallowed (introduced here).** `readonly VAR="$(...)"` takes the exit status of the *builtin*, not of the command substitution, so `set -e` never saw `GET_TER_BUILD_INPUTS.sh` exit 1. An extension without `extra.typo3/cms.version` — which is EXT:tika and EXT:solrdebugtools on `main` today — got an empty label, skipped the template, and **exited 0**:

```
$ PREPARE_RELEASE.sh 14.0.1
Documentation/guides.xml: 14.0.0 -> 14.0.1
extra.typo3/cms.version is not defined
.github/ISSUE_TEMPLATE/bug_report.md:  -> 14.0.1      <- claimed success, changed nothing
exit=0
```

The key is now read in its own statement, so `set -e` sees it.

**2. A half-edited tree (pre-existing).** `guides.xml` was written before the template was touched, so any template failure left the release version bumped in one file only. Both inputs are now validated before anything is written.

**3. A template without the extension's own line was silent (pre-existing).** EXT:solrdebugtools' template has only an `EXT:solr Version` line and no `EXT:solrdebugtools` one. It now aborts instead of reporting a bump that did not happen.

### Verification

| Case | Before | After |
|---|---|---|
| EXT:solr, real files | only `EXT:solr` line + `release=` move | identical |
| EXT:solrconsole, real files | only its own line moves, siblings untouched | identical |
| No `extra.typo3/cms.version` (tika, solrdebugtools) | exit 0, `guides.xml` bumped, template silently skipped | exit 1, **nothing written** |
| Template lacks `EXT:<key>` line (solrdebugtools) | exit 0, `guides.xml` bumped, template silently skipped | exit 1, **nothing written** |

### Blast radius today

No repository other than EXT:solr references `PREPARE_RELEASE.sh` or `ext:release:prepare` on `main` — checked `ext-tika`, `ext-solrfal`, `ext-solrconsole`, `ext-solrdebugtools`. So this only changes EXT:solr's own behaviour, which is unchanged, plus EXT:solrconsole once its side lands.

Two add-ons will need a small prerequisite before they can adopt `ext:release:prepare`, and now get a clear error instead of a silent no-op:

* **EXT:tika** — add `extra.typo3/cms.version`
* **EXT:solrdebugtools** — add `extra.typo3/cms.version` and an `EXT:solrdebugtools Version: [e.g. ...]` line to its bug report template

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4785, opened 2026-09-16
[BUGFIX] Ignore an emptied filter section in the plugin FlexForm

### LINKED ISSUE #4784: [BUG] Emptying the plugin's filter section crashes the search page (foreach over string in ConfigurationService)
**Describe the bug**

Emptying the plugin's *Query → Filter* section (`search.query.filter`) takes the whole search page down with a 500.

When the last filter row is removed in the backend, FormEngine writes the section back as an element that still contains its indentation whitespace:

```xml
<field index="search.query.filter">
    <el index="el">
    </el>
</field>
```

`FlexFormService::convertFlexFormContentToArray()` parses that as the **string** `"\n    "`, not as an array. `ConfigurationService::getFilterFromFlexForm()` guards the value only with `empty()` — and `empty("\n    ")` is `false` — so it falls through to `foreach ($filters as $filter)` over a string:

```
(1/1) #1476107295 TYPO3\CMS\Core\Error\Exception
PHP Warning: foreach() argument must be of type array|object, string given in
.../solr/Classes/System/Service/ConfigurationService.php line 107
```

Stack: `ConfigurationService::getFilterFromFlexForm()` ← `overrideFilter()` ← `overrideConfigurationWithFlexFormSettings()` ← `AbstractBaseController::initializeAction()` ← `SearchController::initializeAction()`.

Only one of the three possible shapes is fatal. Verified against TYPO3's own `FlexFormService`:

| FlexForm | parsed value | `empty()` | result |
|---|---|---|---|
| `<el index="el">` + newline/indent | `string("\n    ")` | `false` | **crash** |
| `<el index="el"/>` (truly empty) | `string("")` | `true` | ok |
| whole `<field>` element absent | `NULL` | `true` | ok |

The first shape is what an editor produces through the normal backend UI, so this is reachable without touching the database.

**To Reproduce**

1. Put a *Search: Form, Result, Additional Components* (`solr_pi_results`) plugin on a page.
2. Open it in the backend, tab **Query**, add a filter row under *Filter* (any field/value), save.
3. Delete that row again, save. The stored `pi_flexform` now has an empty `<el index="el">` containing whitespace.
4. Open the search page in the frontend → exception instead of the search form.

**Expected behavior**

An empty filter section is treated as "no filter configured", exactly like an absent one. It should not be possible to break the frontend by adding and removing a filter row in the backend.

**Actual behavior**

Uncaught `TYPO3\CMS\Core\Error\Exception` (the PHP warning is promoted to an exception by TYPO3's error handler), the page returns a 500.

**Suggested fix**

Guard the type instead of relying on `empty()` in `getFilterFromFlexForm()`:

```php
if (!is_array($filters) || $filters === []) {
    return [];
}
```

PR follows.

**Used versions:**
 - TYPO3 Version: 14.3.7
 - Browser: n/a (server side)
 - EXT:solr Version: 14.0.1 (code is identical on `main`)
 - Used Apache Solr Version: 10.0.0
 - PHP Version: 8.4.20
 - MySQL Version: MariaDB 10.11

**Additional context**

`main` carries the same code, so this is not fixed there. Unrelated to #4688, which is a backend `TcaFlexProcess` problem with `TCEFORM ... disabled`.

audef1:
**What was changed**

`ConfigurationService::getFilterFromFlexForm()` now checks that the FlexForm filter section actually is an array, instead of relying on `empty()`.

Removing the last row of the plugin's *Query → Filter* section in the backend leaves the element behind with its indentation:

```xml
<field index="search.query.filter">
    <el index="el">
    </el>
</field>
```

`FlexFormService::convertFlexFormContentToArray()` parses that as the **string** `"\n    "`. `empty("\n    ")` is `false`, so the guard does not catch it and the method runs `foreach()` over a string — the search page then returns a 500:

```
(1/1) #1476107295 TYPO3\CMS\Core\Error\Exception
PHP Warning: foreach() argument must be of type array|object, string given in
.../Classes/System/Service/ConfigurationService.php line 107
```

Only that one shape is fatal — a truly empty `<el index="el"/>` parses as `""` and an absent field as `NULL`, both caught by `empty()`. Since the whitespace variant is exactly what FormEngine writes, an editor can take the frontend down through the normal backend UI, without touching the database.

**Regression test**

`ConfigurationServiceTest::emptyFilterSectionIsIgnored()` covers all three empty shapes. The unit test config has `failOnWarning="true"`, so the test is a real guard:

- without the fix: `exit=1`, warning raised at `ConfigurationService.php:107` for the `section emptied in the backend` case
- with the fix: `exit=0`, full `ConfigurationServiceTest` suite green (36 tests, 42 assertions)

`php-cs-fixer` reports no changes for either file.

Fixes #4784


dkd-kaehm (approved):
@audef1 
Thanks for fixing that issue.
Which EXT:solr versions are affected?
Which ports are required?

audef1:
Thanks for the quick review!

**Which versions are affected**

All of them — `getFilterFromFlexForm()` guards the value with `empty()` only, and that is identical on every branch I could fetch:

| branch | EXT:solr | guard |
|---|---|---|
| `main` | 14.0 | `empty($filters)`, no type check |
| `release-13.1.x` | 13.1 | same |
| `release-13.0.x` | 13.0 | same |
| `release-12.1.x` | 12.1 | same |
| `release-12.0.x` | 12.0 | same |
| `release-11.6.x` / `release-11.5.x` / `release-11.2.x` / `release-11.1.x` / `release-11.0.x` / `release-10.0.x` | 11.x / 10.x | same |

Nothing here is version specific: the trigger is the FlexForm shape FormEngine writes when the last filter row is removed (`<el index="el">` containing only the indentation), which `FlexFormService` parses as the string `"\n    "`, and `empty("\n    ")` is `false` in every PHP version.

**Which ports are required**

For us only `main` / 14.0 matters — we are on TYPO3 14.3.7 with EXT:solr 14.0.1, and that is the only line we run.

Going by the version matrix, the other currently maintained lines carrying the same code are `release-13.1.x` (TYPO3 13.4) and `release-12.1.x` (TYPO3 12.4), plus the ELTS branches `release-11.6.x` and `release-11.2.x`. The method is byte identical there, so the patch and the test apply unchanged. I am happy to open the backport PRs if you want them — just say which branches.

**Environment where it was found**

* TYPO3: 14.3.7
* EXT:solr: 14.0.1
* PHP: 8.4.20
* Apache Solr: 10.0.0
* DB: MariaDB 10.11
* Application context: `Production/Staging`

**One note on severity, which may matter for the backport decision**

How bad this is depends on `$GLOBALS['TYPO3_CONF_VARS']['SYS']['exceptionalErrors']`:

* with `E_WARNING` included, TYPO3's `ErrorHandler` turns the warning into `#1476107295 TYPO3\CMS\Core\Error\Exception` and the search page returns a 500 — that is the case we hit and the stack trace in #4784;
* with TYPO3's default, where `E_WARNING` is *not* exceptional, the request survives: PHP raises the warning, the `foreach` body never runs, `getFilterFromFlexForm()` returns `[]` and the filters configured in the plugin are silently dropped.

So the failure mode is either a white screen or silently ignored filters, depending on the installation.

The PHP message wording differs by version but the warning itself does not: PHP >= 8.0 says "foreach() argument must be of type array|object, string given", PHP 7.x said "Invalid argument supplied for foreach()".

PR #4797, opened 2026-09-23
Preserve standard tabs for Solr plugins

infabo:
## Summary

All three Solr content element plugins replace the type configuration created during plugin registration. This removes TYPO3's standard Appearance and Categories tabs and any fields added by other extensions.

Keep that default configuration for all three plugins. Insert `pi_flexform` after the header palette only for the search and results plugins, which have FlexForms. The frequently searched plugin needs no additional fields.

## Verification

- Applied the patch to Solr 13.1.5 and rebuilt TYPO3 13.4 TCA in a project using the extension.
- Confirmed all three Solr CTypes retain the Appearance and Categories tabs. Search and results retain `pi_flexform`; frequently searched does not gain one.
- `php -l Configuration/TCA/Overrides/tt_content.php` and `git diff --check` pass.

Backend UI verification was not performed in the upstream checkout.


dkd-kaehm:
@infabo FYI: not your fault currently. TYPO3 testing framwork has a new guard for fixtures.
I'll fix that soon and rebase your PRs branch.

